PR #146, opened 2023-05-05
Setup and use React Query

annavik:
React Query is a data-fetching library for web applications. It includes hooks for fetching, caching and updating asynchronous data in React. This will make it possible for us to reduce amount of manual work needed here, especially once we start doing more mutations.

Some links for more info:
- https://www.npmjs.com/package/@tanstack/react-query) 
- https://tanstack.com/query/latest/docs/react/overview)

Updates in in PR:
- Replace our custom fetching with React Query hooks.
- Make React Query dev tools available when app is running in dev mode.
- Fetch status summary every 10 second.
- Data is now cached during the app session, meaning if the user change page and go back, we will show the previous data while new data is being fetched in the background.
- Data is now being updated more often, for example when the browser window with the app running gets in focus. This only goes for data that is relevant for the current app state, which is nice.
- Show both "is loading" (when we fetch data for the first time) and "is fetching" (when we update data in background) in UI (will send a clip to you on Slack @mihow, it was too big to add here).
- Some component cleanup.


ui/src/utils/language.ts
```diff
@@ -128,6 +129,7 @@ const ENGLISH_STRINGS: { [key in STRING]: string } = {
 
   /* OTHER */
   [STRING.CLOSE]: 'Close',
+  [STRING.LOADING_DATA]: 'Loading data',
```
mihow, line 132: Can we display "Updating data..." when the page is loading new data in the background? I believe you mention two different loading messages in the PR description but I am only seeing one in the deploy preview.

mihow (approved):
This is badass! I love that the previously fetched data remains visible and it shows a loading status as an update is fetched. The browser focus and better polling are all great too. Thanks for integrating this library.

ui/src/utils/language.ts
annavik: Yes good point, I tweaked the message to say "Updating data" when we are just updating. On initial load, we only showed the large spinner before, I added the message here as well now, but with the text "Loading data".

Screenshot 2023-05-08 at 10 33 11

Screenshot 2023-05-08 at 10 33 31

Also did a small tweak to the error state since you last watched. If we have an error but are loading, we show the spinner, not the error message.

PR #161, opened 2023-05-30
New combo box components + more advanced map features

annavik:
### Summary of changes:
- Add new combo box component to design system (using [Radix dropdown menu](https://www.radix-ui.com/docs/primitives/components/dropdown-menu) and new lib [⌘K](https://cmdk.paco.me/)).
- Add feature "Search on the map"
  - For the UI we use the new combo box here.
  - For location data, we use the [Nominatim API](https://nominatim.org/). This API uses OpenStreetMap data, which makes it a got fit for us, since OSM is our current map tile provider.
- Add feature "Use current location" to deployment map (accessing browser location if we have permission).
- Add feature "Reset" to deployment map (making it possible to revert to initial position).
- Some refactoring:
  - Splitting up the map component into one editable and one static one for simplicity.
  - Restructuring files (for example each deployment section get its own folder) for bring some order.

### Screenshots:
New combo box component added to Storybook:
Screenshot 2023-05-30 at 11 07 28

Deployment map with some new features:
Screenshot 2023-05-30 at 11 07 17


ui/src/design-system/components/form-stepper/form-stepper.tsx
```diff
@@ -58,9 +58,9 @@ const FormStepperItem = ({
       [styles.active]: active,
     })}
     onClick={onClick}
-    onKeyDown={(event) => {
```
annavik, line 61: Unrelated cleanup

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
```diff
@@ -0,0 +1,57 @@
...
+    position: new MarkerPosition(latitude, longitude),
+  }
+}
+
+const getFetchUrl = (searchString: string) =>
+  `${API_URL}?format=json&q=${searchString}`
+
+export const useGeoSearch = (searchString: string) => {
+  const [fetchUrl, setFetchUrl] = useState<string>()
+  const debouncedSetFetchUrl = useCallback(_.debounce(setFetchUrl, 200), [
```
annavik, line 26: If typing fast we don't want to call the API on each change

ui/src/design-system/components/button/button.tsx
```diff
@@ -39,15 +41,15 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
         className={classNames(styles.button, {
           [styles.success]: theme === ButtonTheme.Success,
           [styles.plain]: theme === ButtonTheme.Plain,
-          [styles.disabled]: disabled,
+          [styles.disabled]: disabled ?? loading,
         })}
-        disabled={disabled}
+        disabled={disabled ?? loading}
```
kimf, line 46: Not really part of this PR, but recently read a twitter thread, that I can't find again, where they discussed a lot about the accessibility of disabled buttons..
Found this one when doing a quick googling now https://axesslab.com/disabled-buttons-suck/

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/geo-search.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  const items = useMemo(() => {
+    if (!data?.length) {
+      return []
+    }
+    return data.map((result) => ({
+      id: result.osmId,
+      label: result.displayName,
+    }))
+  }, [data])
```
kimf, line 23: What is the reasoning for using useMemo here instead of deriving the data in the render function?

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
kimf: I wonder if it would be cleaner to keep this method "debounce free", and then debounce the usage of useGeoSearch?

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
```diff
@@ -0,0 +1,57 @@
...
+    setFetchUrl,
+  ])
+  const [data, setData] = useState<SearchResult[]>()
+  const [isLoading, setIsLoading] = useState(true)
+  const [error, setError] = useState<Error>()
+
+  useEffect(() => {
+    debouncedSetFetchUrl(getFetchUrl(searchString))
+  }, [searchString])
```
kimf, line 35: is the useEffect needed here as the fetchUrl is in state? Will it not get called on each render when searchString (prop) or url (state) changes anyway?

kimf (approved):

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/geo-search.tsx
annavik: Since `items` is an object, if we would not use `useMemo`, the object would get a new ref each render, which might cause unnecessary rendering, especially since we pass it as a prop to child components. Probably not a huge impact in practice though. I know React is pretty optimised to handle re rendering. Might be better to optimise for readability?

kimf: That is sound reasoning. I was just curious :D

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
annavik: Yes good point, that got cleaner! I could not debounce the usage of useGeoSearch (rules of hooks), but I extracted the debouncing from it. What do you think?

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
annavik: Fixed together with cleanup above.

ui/src/pages/deployment-details/deployment-details-form/section-location/geo-search/useGeoSearch.ts
kimf: Looks good!

ui/src/design-system/components/button/button.tsx
annavik: Yes this is an interesting topic, had a nerdy chat with my UX colleagues in Umeå. I agree most of the time it's better to avoid disabled buttons. In our case, we use disabled state on the deployment save button (if we have no changes to save yet or if data is missing or malformed). One challenge for us is the form has multiple sections, but the save button is shared. For example, how can we communicate there is an issue in a section that is currently not in view?

Something to think about for sure, thanks for bringing it up!

PR #176, opened 2023-06-27
Capture picker

annavik:
Here comes a draft for the new capture picker component, so we can start test things out! Keeping it as a draft, because design is still WIP.

Some details:
- Changing active item can be done by click, by using arrow buttons or by using keyboard arrow (up and down).
- Active item will do a smooth scroll into view if not visible (for example when using keyboard navigation).
- Bar width goes from 0 to 100% where the max amount of detections decides what is 100%.
- Still just using dummy data (but added a few more captures so we can test out scroll features).

Screenshot 2023-06-27 at 15 44 31

### DIFF
--- a/ui/src/data-services/hooks/useSessions.ts
+++ b/ui/src/data-services/hooks/useSessions.ts
@@ -2,7 +2,6 @@ import { useQuery } from '@tanstack/react-query'
 import axios from 'axios'
 import { FetchParams } from 'data-services/types'
 import { getFetchUrl } from 'data-services/utils'
-import { ServerOccurrence } from '../models/occurrence'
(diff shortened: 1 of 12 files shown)


mihow:
I'm not sure why the backend test just started failing, but the Netlify preview is still being deployed. I think this looks great for MVP! We can try 10k images soon.

annavik:
> I'm not sure why the backend test just started failing, but the Netlify preview is still being deployed. I think this looks great for MVP! We can try 10k images soon.

Yes, weird! Is it ok to merge this despite the failing BE stuff you think? :)

mihow (approved):

PR #181, opened 2023-07-03
Consume API v.2

annavik:
This includes some new queries for fetching captures and capture detections. Also, as part of this PR we are also adding mutations to create and update deployments. Handling of server errors will happen in an upcoming PR, for now we are doing a silent fail.


ui/src/data-services/hooks/useCaptures.ts
```diff
@@ -0,0 +1,49 @@
...
+import { Capture, ServerCapture } from 'data-services/models/capture'
+import { FetchParams } from 'data-services/types'
+import { getFetchUrl } from 'data-services/utils'
+
+const COLLECTION = 'captures'
+
+const convertServerRecord = (record: ServerCapture) => new Capture(record)
+
+const FETCH_PARAMS: FetchParams = {
+  pagination: { page: 0, perPage: 10000 },
```
annavik, line 12: I'm changing this later @mihow 😅 Need to do some infinite scroll thing.

PR #184, opened 2023-07-04
Update UI for occurrence details

annavik:
**In this PR:**
- Fetch occurrence details (before we just used data from collection view)
- Update occurrence details view
- Update occurrence gallery view
- Selections based on occurrence instead of detection in session details

After updates:
Screenshot 2023-07-04 at 18 42 36

Screenshot 2023-07-04 at 18 47 05


mihow (approved):
Love the gallery view updates, and the image list. This view is becoming really usable! I will add some notes about extra fields on the occurrence detail later.

mihow:
I added `duration` and `duration_label` to the Occurrence detail response. We can add `detections_count` as well. Classifications info yet to come.

PR #185, opened 2023-07-05
Filtering and adding more data

annavik:
### In this PR:
- Add filtering feature
- Add filtering links in various places
- Add more info to views after API updates + tweak FE models
- Extract gallery from occurrences and add it to sessions and species views as well

### Screenshots:
One of the new gallery views:
Screenshot 2023-07-05 at 20 31 11

More info in some places, for example in occurrence detail view:
Screenshot 2023-07-05 at 18 59 35

Filtered collection (very basic UI for now):
Screenshot 2023-07-05 at 18 59 24


ui/src/utils/useFilters.ts
mihow, line 1: @annavik This is how the filtering params look for database models related with a foreignkey. They are autogenerated filters using a library, I can probably make custom aliases later

mihow (approved):

mihow (approved):

ui/src/utils/useFilters.ts
annavik: Great!!

Maybe labels below (those ones will be visible in the filter popup) can still be just "Deployment" and "Session", I guess that is the only thing users have to care about, or what do you think?

PR #186, opened 2023-07-05
Mixed UI tweaks

annavik:
### In this PR:
- Add species detail view
- Link from occurrence details to species details and the other way around
- Update copy "Images"->"Captures"
- Skip conditional render of session playback (thanks to access to size of first frame)

### Screenshots:
To be added.


mihow (approved):
The linking between species detail and occurrence detail is so great! I like how you can go back and forth and the list view behind the modal changes very seamlessly as well. I should be able to speed up the API loading so it's very fast for detail views. I did fix on major bug! It was showing examples with the _lowest_ score rather than the highest! (on the backend)

PR #187, opened 2023-07-06
Infinte scroll captures

annavik:
Adding infinite scroll pagination to captures. We do this with:
- [Browser Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)
- [useInfiniteQuery](https://tanstack.com/query/v4/docs/react/reference/useInfiniteQuery)

Next step is to add infinite scroll in the other direction as well (if starting with an offset). Also we could probably optimize the rendering of lists with 10 000 items, but this is a start.

Screenshot 2023-07-06 at 18 42 57

Some other stuff included in this PR:
- Extrating the capture list to a pure UI component, to make testing easier.
- Updating breadcrumb value for session.
- Make it possible to go to first frame with an occurrence (if on first page).
- Hide classification tab for now in occurrence details.


mihow (approved):
This is so helpful already. I am catching bugs in the capture data.

PR #191, opened 2023-07-07
Hook up jobs with BE

annavik:
### Some notes:
- We render stages dynamically. It means stage list can have any size.
- We render stage params dynamically. It means param list can have any size.
- Some params are still missing value (see comments).
- The job status and job progress is made up, but easy to replace with BE values (see comments).
- We refetch both jobs and job details every 10 second. We show a loading text in UI when this happens.

### Screenshots:
Screenshot 2023-07-07 at 12 38 08

Screenshot 2023-07-07 at 12 37 41


ui/src/data-services/hooks/jobs/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const COLLECTION = 'jobs'
+export const REFETCH_INTERVAL = 10000 // Refetch every 10 second
```
annavik, line 2: @mihow here is the polling configured for jobs, you can tweak this value or set to undefined :)

ui/src/data-services/hooks/useStatus.ts
```diff
@@ -14,7 +14,7 @@ interface Status {
 }
 
 const COLLECTION = 'status/summary'
-const REFETCH_INTERVAL = undefined // TODO: Refetch every 10 second
+const REFETCH_INTERVAL = 10000 // Refetch every 10 second
```
annavik, line 17: Also added polling for status (numbers in header).

ui/src/data-services/hooks/useStatus.ts
```diff
@@ -40,7 +40,6 @@ export const useStatus = (): {
         .get<ServerStatus>(fetchUrl)
         .then((res) => convertServerRecord(res.data)),
     refetchInterval: REFETCH_INTERVAL,
-    retry: 0,
```
annavik, line 43: I skipped fetch retry while status endpoint was gone. Adding back!

ui/src/data-services/models/job.ts
```diff
@@ -4,56 +4,102 @@ import { STRING, translate } from 'utils/language'
...
 
-  get status(): number {
-    return this._job.status
+  get status(): JobStatus {
+    return this.getStatus(this._job.status)
   }
 
   get statusDetails(): string {
-    return 'More details about the job status.'
+    // TODO: Return BE value here when available.
```
annavik, line 62: BE could perhaps return something here? This label will show up here:

Screenshot 2023-07-07 at 12 48 39
Screenshot 2023-07-07 at 12 48 47

ui/src/data-services/models/job.ts
```diff
@@ -4,56 +4,102 @@ import { STRING, translate } from 'utils/language'
...
+    return `${this.statusValue * 100}% completed.`
+  }
+
+  get statusValue(): number {
+    // TODO: Return BE value here when available (number between 0 and 1).
+    switch (this.status) {
```
annavik, line 68: Mocked a value based on status for now. This value will control status bar size:

Screenshot 2023-07-07 at 12 50 26

annavik:
Project name and some params (more than shown here) are missing value in details view:

Screenshot 2023-07-07 at 12 51 48

ui/src/data-services/models/job-details.ts
```diff
@@ -0,0 +1,52 @@
...
+      (stage: any) => stage.key === key
+    )
+
+    if (!stage || !progress) {
+      return undefined
+    }
+
+    const name = stage.name
+    const status = this.getStatus(progress.status)
+    const statusLabel = this.getStatusLabel(status)
```
annavik, line 37: This status label is basic, should probably include info about time remaining, but I was not fully sure about the logic here, so skipped for now. This is the place to tweak if needed :)

Screenshot 2023-07-07 at 13 02 41

ui/src/data-services/models/job.ts
mihow: I added a `summary` section to the `progress` JSON with the overall progress & status

image

ui/src/data-services/models/job-details.ts
mihow: I added a new "status label" to the stages progress. So `status` is the ENUM key with the status value. `progress` is the float value of the progress (from 0-1) and `status_label` is a formatted text string.

image

mihow:
> Project name and some params (more than shown here) are missing value in details view:
> 
> Screenshot 2023-07-07 at 12 51 48

I added the missing project and some values for those params!

Also you can change the config & status JSON in the interactive HTML form now, if you want to test different values

image

mihow:
> > Project name and some params (more than shown here) are missing value in details view:



Oh!! Actually there isn't meant to be a value here. The config is just the static info needed to render the form fields. The `progress` object contains the actual value that is updated. Use the `key` from the field config to lookup the value from the `progress`.

image

mihow:
I guess we have two kinds of "fields" that look the same in the UI but get their values from different palces

1) Fields that are configuration for the job before it starts (Batch size, Model choice)
2) Fields that show a detail about the current status of the running job

Maybe I should just _remove_ the status fields from the config object?? Then the field name and the value with both only be in the progress object.

annavik:
> I guess we have two kinds of "fields" that look the same in the UI but get their values from different places
> 
> 1. Fields that are configuration for the job before it starts (Batch size, Model choice)
> 2. Fields that show a detail about the current status of the running job
> 
> Maybe I should just _remove_ the status fields from the config object?? Then the field name and the value with both only be in the progress object.

Oh I see! We can have as is, I just pushed a small update that will check if present in config, if not look in progress.

mihow (approved):

PR #192, opened 2023-07-08
Mixed fixes for FE

annavik:
Here comes some mixed fixes for FE:
- Add pagination to jobs and bump deployments page size to 100
- Link to occurrence details from detection tooltips in session details 
- Show server errors in deployment form
- Cleanup translations (tweak naming, remove unused, etc.)
- Add infinite scrolling for captures in both directions
- Pass selected occurrence with session details request, use capture offset response to make sure selected occurrence is visible

The infinite scrolling to show previous values could need some tweaking later on. We should probably store and retrieve scroll position. This article on the topic seems pretty nice, leaving it here: https://medium.com/swlh/bi-directional-infinite-scroll-in-vue-js-7ac5ba171933

For server errors, we are parsing the BE response to generate a message string. If we can't parse any response (for example if server is down), we will show a generic message. Here is how it could look, if we get an error message about the field "name". Later we want to also show this error in context (next to the input).

Screenshot 2023-07-08 at 03 43 44


mihow (approved):
The loading state of the Session detail page is awesome! Looks really slick how the layout is already in position and then the captures list does a little scroll up. 

Thanks for the error messages from the server in the deployments form. That's great for now!

Clicking from the session detail view to the occurrence detail and back again is really fun. And it's revealing more quirks in the data that I didn't notice before!

mihow:
Also I fixed the sorting & offset issues on my side. So captures are back in order and the `capture_page_offset` appears to be correct.

mihow:
Also I like the "Session start" and "Session end" strings instead of "no more items"!

mihow:
Also! The event detail endpoint supports passing an occurrence id OR detection id (https://app.preview.insectai.org/sessions/1?detection=2)

So we can consider linking images from the occurrence detail modal directly to the frame each one appears on, rather than the first frame of the occurrence :)

If the `capture_page_offset` property in the event detail response is not `null`, you could just always use it when fetching the list of captures -- regardless of what the URL parameter is. `occurrence=1`, `detection=1`, `capture=1`, `timestamp=2022-06-21T04:39:00` are all supported by the backend

PR #193, opened 2023-07-08
Suggestion: Skip connection status

annavik:
Until we have hooked up connection status with BE, maybe it's better to hide this to avoid any confusions during testing? Merge or close this, depending on what you think @mihow :) I'm leaving hook and component in code, just removing the usage.


mihow:
Thanks for this. I will merge if I don't get the backend check actually connecting next week.

annavik:
Closing this, seems we are keeping connection status!

PR #194, opened 2023-07-08
Small plot tweaks

annavik:
Small plot tweaks:
- Add session detail plots
- Add support for orientation
- Auto adjust margins based on label size
- Skip categorical x-axis

Screenshot 2023-07-08 at 14 47 17


mihow (approved):
This is awesome. Thank you for adding support for the horizontal bar chart and updating the session detail page. I will fix those species counts in the next import so they are not all just 1.

PR #195, opened 2023-07-08
Link to a specific capture in session detail view

annavik:
Adding support for links to a specific capture in session detail view. I thought selections based on capture here was a nice approach (instead of using detections) because it also fits with how we present selections in UI.

If both occurrence and capture are part of the URL, the capture will be prioritised for the first selected capture (I think BE is not doing the same for the offset, but I think maybe it should?).

Example links:
- https://deploy-preview-195--ami-web.netlify.app/sessions/1?occurrence=12&capture=30
- https://deploy-preview-195--ami-web.netlify.app/sessions/1?capture=6
- https://deploy-preview-195--ami-web.netlify.app/sessions/1?capture=6&occurrence=12 (here I think maybe an offset matching the capture would be better in the response?)

Also preparing detections in occurrence details to link to a specific capture in session details with that occurrence selected. We don't have this info atm, but I hope for it to work if you just add the capture here @mihow:

Screenshot 2023-07-08 at 15 00 41

### DIFF
--- a/ui/src/data-services/models/occurrence-details.ts
+++ b/ui/src/data-services/models/occurrence-details.ts
@@ -29,6 +29,8 @@ export class OccurrenceDetails extends Occurrence {
 
     return {
       id,
+      captureId:
+        detection.capture !== undefined ? `${detection.capture_id}` : undefined,
       image: {
         src: detection.url,
         width: detection.width,
--- a/ui/src/data-services/models/occurrence-details.ts
+++ b/ui/src/data-services/models/occurrence-details.ts
@@ -30,7 +30,7 @@ export class OccurrenceDetails extends Occurrence {
     return {
       id,
       captureId:
-        detection.capture !== undefined ? `${detection.capture_id}` : undefined,
+        detection.capture !== undefined ? `${detection.capture}` : undefined,
       image: {
         src: detection.url,
         width: detection.width,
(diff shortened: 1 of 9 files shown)


ui/src/data-services/models/occurrence-details.ts
```diff
@@ -29,6 +29,8 @@ export class OccurrenceDetails extends Occurrence {
 
     return {
       id,
+      captureId:
+        detection.capture !== undefined ? `${detection.capture}` : undefined,
```
annavik, line 33: If you want to add capture id to occurrence detections using a different key or something, this is where we need to change how to access it @mihow

mihow:
@annavik great call with using capture.id to target frames instead of dectection.id, that's much more direct and reliable! I made the changes you suggested to make that possible:

- The `capture` param now takes priority in the backend function that calculates the offset. The priority order is: capture.id, timestamp, detection.id, occurrence.id. Which also matches the order of query efficiency (select by capture is the fastest, select by occurrence takes a couple database joins).
- The API response for occurrence detail now contains nested data about the capture for each detection
- The detections look for `capture.id` in `occurrence-details.ts`

Now the linking from detection thumb to the session detail is much more satisfying. You can click on the last thumb and it will actually go to the last frame (and still select the right occurrence). 

image


One small todo (for later) would be to scroll down to the right detection when you navigate from the capture view to the occurrence detail modal. 

Also the back button wasn't working for me at one point, but now it is! So maybe something to look out for.

mihow (approved):

PR #197, opened 2023-07-10
Add some adjustments for small screens

annavik:
I realised the mobile version has gotten minimal attention last week! Here comes some quick tweaks:
- Fix broken session detail view
- Consider the dynamic view port size when we want "full height"
- Adjust paddings and margins here and there (32px -> 16px)
- Adjust typography (go down one step in size and line height)

Still a lot of improvements to made for mobile devices, for example find an alternative to our hover interactions and make table views even more compact? Also the session detail might need some alternative designs? I hope this will take care of most urgent issues though :)

### DIFF
--- a/ui/src/app.module.scss
+++ b/ui/src/app.module.scss
@@ -6,6 +6,7 @@
   background-color: $color-neutral-50;
   width: 100%;
   height: 100vh;
+  height: 100dvh;
   overflow-x: auto;
   overflow-y: auto;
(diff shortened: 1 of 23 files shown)


mihow:
Heck yeah! Fantastic, thank you. I didn't want to say anything but I really wanted the session detail to work on my phone.

mihow (approved):

PR #200, opened 2023-08-01
Add logo and info pages

annavik:
### In this PR:
- Add top bar with AMI logo.
- Prepare info pages. To keep things simple for now, I suggest we use dialogs for this. ~~For now we are using hard coded data in FE app. Happy to change this to a BE call once the endpoint is available :)~~ Updated to use API!

### Screenshots:

Screenshot 2023-08-02 at 11 33 16

Screenshot 2023-08-02 at 11 33 30


ui/src/components/header/infoPages.ts
```diff
@@ -0,0 +1,15 @@
+// TODO: Replace with BE data
+export const infoPages = [
+  {
+    id: 1,
+    name: 'About',
+    content:
+      '<h1>Heading 1</h1><h2>Heading 2</h2><h3>Heading 3</h3><p>Paragraph</p><p>Paragraph with <a href="https://google.com">link</a></p>',
```
annavik, line 7: @mihow please edit this or let me know when the endpoint is available, so we can show the real content

ui/src/components/info-dialog/info-dialog.module.scss
```diff
@@ -0,0 +1,33 @@
+@import 'design-system/variables/colors.scss';
+@import 'design-system/variables/typography.scss';
+
+.content {
+  min-width: 640px;
+  max-width: 100%;
+  padding: 24px 32px;
+  color: $color-neutral-800;
+
+  h1 {
```
annavik, line 10: Some simple styling of HTML elements, let me know if we need more and we can easily extend. The app global styles are a bit limited since we are using CSS modules (making each component define its own styles, rather than using global CSS).

mihow:
Okay I merged the pages endpoints and they are available now!

https://api.dev.insectai.org/api/v2/pages/
https://api.dev.insectai.org/api/v2/pages/about/
https://api.dev.insectai.org/api/v2/pages/help/

Just a couple things if you have any more time:
- Can we use the version of the logo that isn't perfectly square? This one has the little bird on the right cut off. Or is it a limitation of the styles? I think you used the full one for the favicon app icons. But I attached what I have here as well.
- Can we update the URL routes for the pages? So /about & /help work?

Otherwise the styles look just fine! This is so great. Thank you thank you thank you.

Sticker mule cut line
AMI Logo
AMI logo traced

annavik:
> Okay I merged the pages endpoints and they are available now!
> 
> https://api.dev.insectai.org/api/v2/pages/ https://api.dev.insectai.org/api/v2/pages/about/ https://api.dev.insectai.org/api/v2/pages/help/
> 
> Just a couple things if you have any more time:
> 
> * Can we use the version of the logo that isn't perfectly square? This one has the little bird on the right cut off. Or is it a limitation of the styles? I think you used the full one for the favicon app icons. But I attached what I have here as well.
> * Can we update the URL routes for the pages? So /about & /help work?
> 
> Otherwise the styles look just fine! This is so great. Thank you thank you thank you.
> 
> Sticker mule cut line AMI Logo AMI logo traced

- Updated to use API data!
- Good call about the logo, I changed it. Just remembered I added a "Logo material" tab in Figma before, with all the different versions I used for favicons. Makes it easy to adjust export size and format, depending on what you need.
- Regarding routing, can we do it in a follow up PR? I think I would like to give the routing some more thought since the info pages are dynamic and also because the info dialogs can be triggered from anywhere in app.

mihow (approved):

PR #201, opened 2023-08-07
Control info dialogs from URL

annavik:
Since info dialogs can be shown from anywhere in app (but still we want to be able to link to them), I suggest we control their visibility using query params. Examples:
- https://deploy-preview-201--ami-web.netlify.app/?page=about
- https://deploy-preview-201--ami-web.netlify.app/overview?page=help
- https://deploy-preview-201--ami-web.netlify.app/sessions/4?capture=99&page=about

### DIFF
--- a/ui/src/components/info-dialog/useActivePage.ts
+++ b/ui/src/components/info-dialog/useActivePage.ts
@@ -1,0 +1,19 @@
+import { useSearchParams } from 'react-router-dom'
+
+const SEARCH_PARAM_KEY = 'page'
+
+export const useActivePage = () => {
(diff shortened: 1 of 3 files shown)


ui/src/components/info-dialog/useActivePage.ts
```diff
@@ -0,0 +1,19 @@
+import { useSearchParams } from 'react-router-dom'
+
+const SEARCH_PARAM_KEY = 'page'
```
annavik, line 3: Maybe this key is too generic? Let me know if you have other suggestions

PR #204, opened 2023-08-21
Tweak project map

annavik:
Some small tweaks to the project map:
- Enable map dragging
- Tweak min zoom value (we do set initial zoom and bounds based on marker positions, but in order to fit markers on different continents we had to decrease min zoom value)

**Before:**
Screenshot 2023-08-21 at 12 11 44

**After:**
Screenshot 2023-08-21 at 12 12 11

### DIFF
--- a/ui/src/design-system/map/config.ts
+++ b/ui/src/design-system/map/config.ts
@@ -10,7 +10,7 @@ export const TILE_LAYER_URL =
 
 export const DEFAULT_ZOOM = 10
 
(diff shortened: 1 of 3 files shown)


mihow (approved):
This looks great! It encouraged me to optimized the speed of the project overview page as well. Now that we have 1M+ captures the counting is getting very slow. So I disabled captures_count and some others on the map for the time being, until I add some caching methods.

PR #205, opened 2023-08-21
Investigate issues in session details

annavik:
There seems to be some confusions with captures in session detail view. Some images are forever loading and also we have some infinite scroll issues. I have looked in to it and I think the data is a bit strange for some sessions... We could make some things clearer on the FE side though.

**Issue 1: Images not showing**
This issue is related to some session images missing full URL in the response. I have updated the frame loader to stop the loading on error, to make it more clear in UI what happens. In order to solve it, I think we need some BE updates though.

See example: https://api.dev.insectai.org/api/v2/captures/?ordering=timestamp&limit=20&offset=20&event=1 (the item with id 38 does not have a full URL for the image, compared to item with id 128).

**Issue 2: Infinite scroll**
Sometimes we get into a state where the app will continue to load previous items. This is because while the top item is visible, we will continue to check if any previous items are to be loaded. Since the captures look very similar (without detections), it's not obvious that more items are in fact appended.

I suggest we do a (temporary?) update to the UI here, so that we instead load more items on button click. While we are testing things out, I think it might be nice to give users control of the loading. Later on, we could perhaps also look into updating scroll position when items are appended, to make things more clear.

What is also strange though, is that we end up on the offset that we do. For example, if we specify capture id to be 1 (https://api.dev.insectai.org/api/v2/events/1/?capture=1), we get "capture_page_offset" set to 5824 in the response, which seems wrong. I think this is why the linking to a specific capture is not working properly atm! :)


mihow:
Thanks for investigating this. Already the fix for reloading on error is a big help.

I say definitely increase the page size, how about to 200? I think we had it at 20 to test pagination for short sessions.

I fixed some of the images that had a missing full path, but there is one session where the whole folder is missing in the object store (June 20th). I think it only existed on my local machine when I created the test dataset. 

I will look into the offset issue. I think it may be because some queries are filtering by project now and others are not. 

I am okay to disable the auto scroll for the time being. Especially since this component will be changed.

annavik:
Strange with the missing images! I could default to a host if URL is just a path, but it sounds like that would not help us then?

Updated page offset to 200 and removed some unused code. Since we have a separate task for capture offset, I think this can be merged then?

mihow (approved):
This is working for me except for the missing captures and bad offset for which I have made tickets for myself!
https://github.com/RolnickLab/ami-platform/issues/209
https://github.com/RolnickLab/ami-platform/issues/206

PR #210, opened 2023-08-23
Skip proxy and update geo search URL

### LINKED ISSUE #207: Location search API no longer working
When editing a deployment, searching for a location no longer works. It looks like the API may have changed, or do we need to setup a proxy on the backend? 

image

I see it's trying to hit a URL relative to the current host rather than an external provider:
https://app.preview.insectai.org/nominatim?format=json&q=vr

annavik:
Fixes #207 

Issue was related to this: https://github.com/osm-search/Nominatim/issues/3134

Also seems we can skip the proxy now!

### DIFF
--- a/ui/netlify.toml
+++ b/ui/netlify.toml
@@ -31,14 +31,6 @@ force = true
 headers = { X-From = "Netlify" }
 
 [[redirects]]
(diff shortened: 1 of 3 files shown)


mihow (approved):
Works! that's great.

PR #211, opened 2023-08-23
Add projects to page hierachy

annavik:
In this PR we are starting to consider different projects. The main view will no longer be the overview for a single project. Instead, the main view will be a collection of all projects and the user gets to decide what project to have a closer look at.

### Screenshots:
New page for showing all the projects (I have improvised a bit for the UI here, will check with Andre later):
Screenshot 2023-08-23 at 19 40 43

Project overview can now be found one level down the page hierarchy. Breadcrumbs are updated with link back to projects.
Screenshot 2023-08-23 at 19 40 55

This is how it will look for pages further down the hierarchy. Project name will link to project overview.
Screenshot 2023-08-23 at 19 41 39


mihow (approved):
Big moves! This is great. Thanks for designing the initial project view, breadcrumbs and approach to filtering by project. I will add the additional filters on the backend asap.

mihow:
And yes let's run this by Andre when we have a chance!

Issue #212, opened 2023-08-24
Add project filters


annavik:
When we fetch a group of items, we sometimes want to filter based on project id. Deployments and jobs endpoints have this filter, I could need a similar one for:
- /status/summary
- /occurrences
- /events
- /species

Deployments example: https://api.dev.insectai.org/api/v2/deployments?project=2

mihow:
Thanks for finding these! I added the filter to `/status/summary/?project=2` but will come back to the others.

mihow:
These endpoints can now be filtered by project:
- /occurrences
- /events
- /status/summary

each species can belong to multiple projects, so that filter is slightly different. coming soon.

PR #213, opened 2023-08-25
Add project relationship to additional models

mihow:
For many objects, the associated Project was only accessible through multiple joins. But we will be filtering on project in almost every query. This adds a project attribute and filter to all models that need it. 

- Deployment
- SourceImage
- Occurrence
- Taxon

Functions have been added to multiple save() methods so that if a Deployment's project changes, the project of its children will be updated as well.

Taxon has a special project filter that will only return Taxa which have actually been _detected_ within a project.


mihow:
Closes #212

PR #214, opened 2023-08-29
Setup UI for identifications

annavik:
In this PR we setup UI for occurrence details -> identifications. Also we are adding information related to determinations and identifications in table view.

**List identifications:**
Screenshot 2023-09-22 at 17 45 02

**Suggest new ID:**
Screenshot 2023-09-22 at 17 44 51

**Table view:**
Screenshot 2023-09-22 at 17 45 57


annavik:
@mihow I have tweaked the component API:s to better match how we intend to use them. Example:

Screenshot 2023-08-30 at 10 38 32

mihow:
This is actually working pretty well as-is! Can we merge soon and then add your new UI ideas in a follow up PR?

One feature request: a status indicator when searching species. I can't tell if my search has zero results or if there is a lag. I will work on reducing the lag, but there will always be cases where the network might lag.

mihow:
- Fixed issue when adding more than 2 identifications for an occurrence
- Added "withdrawn" attribute to Identifications. When a user adds a new identification, all of their previous IDs will have `withdrawn: true`. This may no be necessary but giving it a try
-  Added new nested `determination_details` attribute to Occurrences, which looks like:
```
occurrence: {
  determination_details: {
      taxon: // taxon determined by all IDs & backend logic
      identification: // best identification used for final ID, if there is one
      prediction: // best prediction, if there is no identifications
      score: // repeat of the best prediction score if using predictions. but otherwise 1 if identified by human, 0 if no identifications or predictions.
      }
}
````

`determination: Taxon` and `determination_score: float` attributes remain for temporary backwards compatibility. We can rename `determination_details` to `determination` when/if ready.

**Comment by mihow** [comment]
Will you sort the identifications & predictions by most recent first? They should be listed in that order in the API response as well.  

image

**Comment by mihow** [comment]
In the mini inline taxonomy, the nodes should be ordered by "deepest" parent first. Right now it seems to be in reverse. We can revise how parents are returned in the API response if that's tricky to render right now. 

image

**Comment by annavik** [comment]
Hey @mihow, think we can merge this beast now and add continue in upcoming PR! The things you mentioned above should be fixed.

**Next up:**
- Test some UI variants of "Suggest ID", not 100% happy with it
- Add button "Agree" to table view, occurrence details header and to list of identifications
- Make it possible to withdraw identification

File: ui/src/data-services/models/occurrence-details.ts [hunk 1]
```diff
@@ -47,4 +110,19 @@ export class OccurrenceDetails extends Occurrence {
       }),
     }
   }
+
+  private _isIdentificationOverridden(identification: any) {
```
annavik, line 114: @mihow here is the logic for when to display an identification as overridden. I'm not seeing withdrawn being set in BE response, which is why I also have some FE logic here. Maybe we can remove the FE stuff later here

annavik: Oh and FIY the response I'm talking about is the occurrence details one, with the list of identifications

mihow:
Looking good! I pushed fixes to the limit parameter and the missing withdraw attribute.

Here are the top todo's IMO for the next revision:

- Add button "Agree" to table view
- Add permissions check for creating new identifications
- Get feedback on UI, bring back auto focus into search box?

mihow (approved):

PR #221, opened 2023-09-03
Implement auth on FE

annavik:
### In this PR
- Setup page for login
- Show login/logout info in top header
- Setup basic auth logic:
  - Store auth token in browser
  - Clear auth token on auth error
  - Pass auth header with all request to server if logged in
  - Setup global context for user info
- Refactor data services to avoid some repetitive code (see hook useAuthorizedQuery for example)
- Invalidate queries on auth status update (both on login and logout)
- Tweak UI based on permissions
- Setup page for sign up

### Screenshots
Screenshot 2023-09-06 at 17 06 19

Screenshot 2023-09-06 at 17 06 09

### Next step
- #225 
- #226


mihow:
Looking really great! The sign in works for me! Thank you for updating the data services and adding the general `useAuthorizedQuery` method.

mihow:
I tried to test the ability to edit/create deployments when logged in or logged out. I am getting an error however. The API routes all currently require a trailing slash at the end. It will automatically redirect GET requests. But POST, PATCH, etc requests cannot be redirected. Will you enforce a trailing slash on all API calls?

> RuntimeError at /api/v2/deployments/19
You called this URL via PATCH, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining PATCH data. Change your form to point to api.dev.insectai.org/api/v2/deployments/19/ (note the trailing slash), or set 

image

annavik:
> I tried to test the ability to edit/create deployments when logged in or logged out. I am getting an error however. The API routes all currently require a trailing slash at the end. It will automatically redirect GET requests. But POST, PATCH, etc requests cannot be redirected. Will you enforce a trailing slash on all API calls?
> 
> > RuntimeError at /api/v2/deployments/19
> > You called this URL via PATCH, but the URL doesn't end in a slash and you have APPEND_SLASH set. Django can't redirect to the slash URL while maintaining PATCH data. Change your form to point to api.dev.insectai.org/api/v2/deployments/19/ (note the trailing slash), or set
> 
> image

Oops! I must have missed this when doing some cleanup in data services. Should be fixed now!

Also making sure to refetch on auth status update (and clear all cached data).

mihow:
Great! I can save changes if I am logged in now, and not when logged out. But I am not longer seeing any errors from the backend (`{"detail":"Authentication credentials were not provided."}`). So maybe check that out.

annavik:
> Great! I can save changes if I am logged in now, and not when logged out. But I am not longer seeing any errors from the backend (`{"detail":"Authentication credentials were not provided."}`). So maybe check that out.

I think this was because when we get 403 in a response, I will logout the user and clear all queries, which will trigger a data refetch. This probably triggered UI to update and error never got visible? Let's talk more about the logic here.

Now I'm considering user permissions in FE, so editing should never be presented as an option to begin with in your case.

annavik:
Some updates since yesterday:
- Considered user permissions for creating and editing deployments. I think for the create case, it would be nice to not have the permission set on object level though. Logic got a bit weird.
- Added page for sign up. A valid sign up ends up in 500 though, something with email verification perhaps?
- Tweaked the UI a bit for login and sign up
  - Made it possible to skip to projects
  - Made it possible to navigate from login<->sign up without re rendering (video will for example not restart)

For the new sign up I spent some time on making sure server errors connected to form fields end up in the correct place. Example:
Screenshot 2023-09-05 at 17 59 51
Screenshot 2023-09-05 at 18 00 04

ui/src/pages/deployments/deployments.tsx
```diff
@@ -19,6 +19,7 @@ export const Deployments = () => {
     items: deployments,
     defaultSort: { field: 'name', order: 'desc' },
   })
+  const canCreate = deployments?.some((deployment) => deployment.canCreate)
```
annavik, line 22: This logic ended up a bit strange, for example what if you have no deployments? Can we put it on collection level you think? :)

mihow: Ah yes, the create permission makes way more sense on the collection level! I will do that

mihow:
> > Great! I can save changes if I am logged in now, and not when logged out. But I am not longer seeing any errors from the backend (`{"detail":"Authentication credentials were not provided."}`). So maybe check that out.
> 
> I think this was because when we get 403 in a response, I will logout the user and clear all queries, which will trigger a data refetch. This probably triggered UI to update and error never got visible? Let's talk more about the logic here.
> 
> Now I'm considering user permissions in FE, so editing should never be presented as an option to begin with in your case.

If you get a 403 in a response it doesn't mean the user should be logged out, it just means they tried to access something they don't have permissions for. And you are correct, they shouldn't be presented with options they don't have permissions to use. But they might find ways to try to access things even when there isn't a link (e.g. a friend shares a link to a detection in a private project) -- in which case it's okay to see the permission denied error! But don't log them out.

annavik:
> > > Great! I can save changes if I am logged in now, and not when logged out. But I am not longer seeing any errors from the backend (`{"detail":"Authentication credentials were not provided."}`). So maybe check that out.
> > 
> > 
> > I think this was because when we get 403 in a response, I will logout the user and clear all queries, which will trigger a data refetch. This probably triggered UI to update and error never got visible? Let's talk more about the logic here.
> > Now I'm considering user permissions in FE, so editing should never be presented as an option to begin with in your case.
> 
> If you get a 403 in a response it doesn't mean the user should be logged out, it just means they tried to access something they don't have permissions for. And you are correct, they shouldn't be presented with options they don't have permissions to use. But they might find ways to try to access things even when there isn't an option visible (e.g. a friend shares a link to a detection in a private project) -- in which case it's okay to see the permission denied error! But don't log them out.

Oh I see what you mean! What I need to do though is clear the token if it's not valid anymore. I'm getting 403 in that case as well, but you are correct there could be others reasons behind the 403. Do you think we can use 401 instead when credentials are not valid (and in that case I will clear token on FE), or do you have other suggestions on the logic? Maybe we can have a chat :)

Why I want to have this error handling on each request is because we are not continuously refreshing the token in our currently auth flow.

mihow:
1) I opened a ticket to the auth framework we are using about returning a 401 vs. a 403. That makes a lot of sense! For now, one workaround is to check the content of the response. You should get 
`{"detail":"Invalid token."}` if the user is logged out or was deactivated. OR another alternative is to poll the /auth/users/me endpoint to check for logged out status. That seems very reasonable to me. 
https://github.com/sunscrapers/djoser/issues/760

2) I moved the "create" permission to the collection level
https://github.com/RolnickLab/ami-platform/pull/222

ui/src/data-services/constants.ts
```diff
@@ -1 +1,21 @@
 export const API_URL = '/api/v2'
+
+export const API_ROUTES = {
```
mihow, line 3: Very glad to see the routes being centralized!

ui/src/utils/constants.ts
```diff
@@ -0,0 +1,5 @@
+export const LINKS = {
```
mihow, line 1: Should this be part of ROUTES?

mihow (approved):
Tested signup, login, logout & permissions to edit Deployments. Feels great from the UI! Very dynamic and fast.

mihow:
@annavik I'm not seeing the "Add deployment" option. Did the permissions on the collection work? I'm going to go ahead and merge this and we can follow up. P.S. let's change the string for this action to "Register new deployment"

PR #230, opened 2023-09-07
Tweak logic for register new deployment

annavik:
Show "Register new deployment" if logged in user has create permissions.


mihow:
It works! I believe the only thing missing is the project assignment. When you create a deployment it "disappears" because it is not assigned to a project. Can you pass the current project ID to the POST request when creating a Deployment? (the property is just called `project` and accepts a project ID). You can test that in the API test page as well: https://api.dev.insectai.org/api/v2/deployments/

mihow (changes requested):
It works! I believe the only thing missing is the project assignment. When you create a deployment it "disappears" because it is not assigned to a project. Can you pass the current project ID to the POST request when creating a Deployment? (the property is just called project and accepts a project ID). You can test that in the API test page as well: https://api.dev.insectai.org/api/v2/deployments/

annavik:
> It works! I believe the only thing missing is the project assignment. When you create a deployment it "disappears" because it is not assigned to a project. Can you pass the current project ID to the POST request when creating a Deployment? (the property is just called project and accepts a project ID). You can test that in the API test page as well: https://api.dev.insectai.org/api/v2/deployments/

Ah yes, sorry, forgot about that. Code is updated!

I believe the project param expects a full url, would it be possible to just pass the id here?

ui/src/data-services/hooks/deployments/useCreateDeployment.ts
```diff
@@ -14,7 +13,7 @@ const convertToServerFieldValues = (fieldValues: DeploymentFieldValues) => ({
   latitude: fieldValues.latitude,
   longitude: fieldValues.longitude,
   occurrences: [],
-  project: null,
+  project: `http://api.dev.insectai.org/api/v2/projects/${fieldValues.projectId}/`, // TODO: Can we pass just id here?
```
annavik, line 16: I'm getting 400 if I just pass the ID here @mihow

mihow: I just fixed this! The deployments endpoint now accepts a project id for the project field, and it returns a nested representation of the project instead of just the details URL.

image

annavik: Thank you! Just pushed a FE fix

mihow (approved):
Works!

PR #232, opened 2023-09-07
[Draft]: Setup user info form

annavik:
Using same form structure as when editing deployment details (but without the wizard).

Left to do:
- Look into why BE is not updating
- Get rid of hard coded strings
- Move form styles to a global place where all forms can share them

### DIFF
--- a/ui/src/components/header/user-info/user-info-form/user-info-form.module.scss
+++ b/ui/src/components/header/user-info/user-info-form/user-info-form.module.scss
@@ -1,0 +1,40 @@
+@import 'design-system/variables/colors.scss';
+@import 'design-system/variables/typography.scss';
+
+.section {
+  margin: 32px;
+  box-sizing: border-box;
+
+  &:not(:last-child) {
+    padding-bottom: 32px;
+    border-bottom: 1px solid $color-neutral-100;
+  }
+}
+
+.sectionContent {
+  display: grid;
+  grid-template-columns: 1fr;
+  gap: 32px;
+
+  &:not(:last-child) {
+    margin-bottom: 32px;
+  }
+}
+
+.sectionRow {
+  display: grid;
+  grid-template-columns: minmax(0, 1fr) minmax(0, 1fr);
+  gap: 32px;
+}
+
+.formActions {
+  display: flex;
+  gap: 16px;
+}
+
+.formError {
+  @include paragraph-x-small();
+  padding: 16px 32px;
+  color: $color-destructive-600;
+  border-bottom: 2px solid $color-neutral-100;
+}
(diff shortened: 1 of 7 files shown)


mihow:
I realized the problem is that we need to post to `/auth/users/<pk>/` instead of `/auth/users/me/` 

BUT I decide to delete the old /users/ endpoints and merge the new Djoser auth endpoints into those.

So that means `/auth/users/` is no more! everything has been moved to `/users/` (`/api/v2/users` to be exact) except for `/auth/token/login/` & `/auth/token/logout`

So will you update the endpoints to use:

`/users/me/` (check token and current user info, including profile info like name.)
`/users/<pk>/` (profile detail and profile updates)
`/auth/token/login/`
`/auth/token/logout/`

mihow:
It works! I added the current email address to the /users/me response. Will you update the profile modal to show the email? And under password it could say "Contact an administrator to change your email or password". Or if it looks better, you can put that message somewhere else and put a fake obscured password value (*****) under the Password heading.

mihow:
I cherry-picked the backend change to main and deployed it to our dev site so it's available now.
image
https://api.dev.insectai.org/api/v2/users/me/

annavik:
Hey @mihow, just marked this PR as ready and updated the PR description, check it out!

ui/src/components/header/user-info-dialog/user-info-form/user-info-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { InputContent, InputValue } from 'design-system/components/input/input'
+import { useEffect, useState } from 'react'
+import { useForm } from 'react-hook-form'
+import { STRING, translate } from 'utils/language'
+import { parseServerError } from 'utils/parseServerError/parseServerError'
+import { UserInfo } from 'utils/user/types'
+import { UserInfoImageUpload } from '../user-info-image-upload/user-info-image-upload'
+import styles from './user-info-form.module.scss'
+
+const IMAGE_MAX_SIZE = 1024 * 1024 // 1MB
```
mihow, line 16: Nice

ui/src/components/header/user-info-dialog/user-info-form/user-info-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+}
+
+const config: FormConfig = {
+  name: {
+    label: 'Name',
+  },
+  image: {
+    label: 'Image',
+    description:
+      'The image must smaller than 1MB. Valid formats are PNG, GIF and JPEG.',
```
mihow, line 30: Can you use MAX_IMAGE_SIZE in this string?

mihow (approved):
We have user profiles! And hooray for icons. Looks great.

 I added a follow up ticket about a possible progress indicator later: #242

ui/src/components/header/user-info-dialog/user-info-form/user-info-form.tsx
annavik: Yes good point!!

PR #238, opened 2023-09-11
Add unit tests for auth related code

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/226

Also did some small lint tweaks, so that we can start include scripts for both (frontend) lint and test in pipelines, if we want?

As part of this PR, we also include 2 new libs:
- [nock](https://github.com/nock/nock): for mocking server calls
- [@testing-library/react-hooks](https://react-hooks-testing-library.com/): for making it possible to test React hooks and contexts without too much hassle

Choice of method and libs is inspired by [React Query docs](https://tanstack.com/query/latest/docs/react/guides/testing).


mihow:
I added the linting & tests to the github workflows and split up the frontend & backend tests so you don't have to wait for django stuff on every frontend change. I also added a pre-commit hook for the ESlint checks so you can catch linting errors before you commit them. Installation instructions here if you don't use it already! https://pre-commit.com/

mihow (approved):
Thanks for so many tests! I think this will help us sleep at night and help us dev without fear!

PR #240, opened 2023-09-12
Update auth URLs

mihow:
Update endpoints for /users/ and /users/me/ used for registration and checking login state to match new backend routes


mihow:
@annavik This works for me but I want to make sure I'm not missing anything! Merge away if it looks good on your end.

annavik:
Looks good to me! Only think I noticed is that now when email is not part of the /me response, this can happen:

Screenshot 2023-09-12 at 09 47 12

Let's talk about if we can add it back, or if we should do some FE workaround here.

PR #253, opened 2023-09-28
Test using new image resizing service

mihow:
This requires the `public_base_url` to be updated on the StorageSource in the database


annavik:
Exciting! This will be very useful

PR #262, opened 2023-10-02
Taxonomic hierarchy

annavik:
**In this PR:**
- Create a taxonomic tree from search result data
- Update taxon search UI to be more similar to label studio
- Highlight selected item
- Add loading to geo search as well
- Update file structure for combo box variants, to make things a bit more clear

**Some backend suggetsions:**
- Backend limit param seem to not work (I pass limit=5 but get around 10 results)
- Backend taxonomy parent list is not always complete, can we include more parent levels? That would make it easier for me to create a full tree on FE! :) See https://api.dev.insectai.org/api/v2/taxa/suggest/?q=Baileya%20ophthalmica&limit=5 for example. Can we have all items nest parents all the way back to Lepidoptera?

**Screenshots:**
Screenshot 2023-10-02 at 17 40 46


annavik:
Note to self:
- Make popover wider than input
- Bring back rank labels

mihow (changes requested):
I see your notes from our Slack conversation, but here they are again in a review to help keep track!
- Bring back rank labels
- Make popover wider than input
- Backend: Add more taxonomic ranks
- Backend: always return all parents for a taxon

mihow (approved):
Looks great! I fixed the limit parameter on the BE. I will revisit the additional taxon ranks. I need to represent hierarchical data in a more efficient way in the database. Thanks for adding the loading status to the geo search as well!

mihow:
I love how when I click back in the search field after making a selection, it shows me my selection in context! And actually reveals other species in the same genus. 
image

PR #266, opened 2023-10-03
Pass "agreed with" params (if applicable) when creating an identification

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/251

### DIFF
--- a/ui/src/data-services/models/occurrence.ts
+++ b/ui/src/data-services/models/occurrence.ts
@@ -38,6 +38,22 @@ export class Occurrence {
     return `${this._occurrence.determination.id}`
   }
 
+  get determinationIdentificationId(): string | undefined {
+    const determinationIdentification =
+      this._occurrence.determination_details?.identification
+
+    return determinationIdentification
+      ? `${determinationIdentification.id}`
+      : undefined
+  }
+
+  get determinationPredictionId(): string | undefined {
+    const determinationPrediction =
+      this._occurrence.determination_details?.prediction
+
+    return determinationPrediction ? `${determinationPrediction.id}` : undefined
+  }
+
   get determinationScore(): number {
     const score = this._occurrence.determination_details.score
 
--- a/ui/src/data-services/hooks/identifications/useCreateIdentification.ts
+++ b/ui/src/data-services/hooks/identifications/useCreateIdentification.ts
@@ -4,7 +4,18 @@ import { API_ROUTES, API_URL } from 'data-services/constants'
 import { getAuthHeader } from 'data-services/utils'
 import { useUser } from 'utils/user/userContext'
 
-const convertToServerFieldValues = (fieldValues: any) => ({
+interface IdentificationFieldValues {
+  agreeWith?: {
+    identificationId?: string
+    predictionId?: string
+  }
+  occurrenceId: string
+  taxonId: string
+}
+
+const convertToServerFieldValues = (
+  fieldValues: IdentificationFieldValues
+) => ({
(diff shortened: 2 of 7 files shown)


ui/src/data-services/models/occurrence.ts
```diff
@@ -38,6 +38,22 @@ export class Occurrence {
     return `${this._occurrence.determination.id}`
   }
 
+  get determinationIdentificationId(): string | undefined {
```
annavik, line 41: Not super happy with this, I think a separate model representation for identifications/predictions could simplify, but perhaps not in this PR

mihow: Okay yes. I'm curious to know more about why you need to make additional prop functions like `determinationPredictionId()` instead of accessing `determination_details` directly

mihow (changes requested):
This looks great, but I am not seeing either agreed_with field saved in the API response. Were you able to see it working? For example:
https://api.dev.insectai.org/api/v2/identifications/210/

image

ui/src/data-services/models/occurrence.ts
annavik: I want to keep the interface separated from BE as far as possible. That is why I have setup a separate layer to handle the all backend communication, including converting data. This means, when you are making an API update (like changing the structure of data), I don't have make updates all over the FE app. Instead I can just update the FE model class. 

Also I'm using this layer to simplify a bit. Sometimes I'm getting more than needed from BE and somtimes in a format that requires a bit of proccessing before rendering (I think taxa is one good example of this). In this way, I can choose what to expose to the interface layer and how.

I do think the classes are sometimes a bit overkill and there are definitely room for improvements here and there (like in code above), but some kind of FE representation of the data is pretty nice for me to have :) Happy to discuss more, maybe you see things from a different perspective?

annavik:
> This looks great, but I am not seeing either agreed_with field saved in the API response. Were you able to see it working? For example: https://api.dev.insectai.org/api/v2/identifications/210/
> 
> image

Hmm, no I did not test that values are saved on BE, just that I pass them. Let's have a closer look at that!

ui/src/data-services/models/occurrence.ts
mihow: Thank you for that explanation! I really appreciate you keeping the frontend working while the backend has lagged behind. It makes sense to keep the objects in the frontend specific and simpler when they can be. But I also like the idea of eventually making types/models more consistent between the frontend & backend when the API is stable. At some point would even like to try using the types generated by the OpenAPI SDK directly in React. We can have multiple "view" types for a model (`OccurrenceIdentificationAgreeDetails`), but it would be awesome to define them in one place. Especially for validation, and for when we are all full stack developers!

annavik: I would love that, it would make the FE app less fragile also, if we can use a more type safe approach!

annavik:
> > This looks great, but I am not seeing either agreed_with field saved in the API response. Were you able to see it working? For example: https://api.dev.insectai.org/api/v2/identifications/210/
> > image
> 
> Hmm, no I did not test that values are saved on BE, just that I pass them. Let's have a closer look at that!

Strange! I had a closer look and works fine for me, I wonder what is might be? Some examples:

Screenshot 2023-10-05 at 16 54 05
Screenshot 2023-10-05 at 16 54 21

https://api.dev.insectai.org/api/v2/identifications/219/

mihow (approved):
Perhaps I was looking a the wrong identification before

PR #267, opened 2023-10-03
Fix bug with "About" and "Help" buttons

annavik:
Fixing small bug in header where "About" and "Help" were not possible to click. Related to click event not being passed from button component, which caused a dialog error.


mihow (approved):
Works!

PR #268, opened 2023-10-04
Make it possible to inspect job details

annavik:
Fixing issue with job details not being available. Issue was related to server data not being converted properly on FE.


mihow (approved):
Works! I love the job detail view!

PR #269, opened 2023-10-04
Setup (more) generic form helpers

annavik:
We already have some form helpers from before (see `FormController` and `FormField`), but in this PR we will also setup some help components to handle form layout. Also we setup a hook `useFormError` to streamline error handling further.

As part of this PR we are also updating current forms to use these new helpers. In the next one, we will create som new forms for projects. Curious to see if we can do that without adding a lot of code, thanks to this.


mihow (changes requested):
Awesome! It looks seamless, except for I am unable to open the dialog for "register new deployment"

`Uncaught TypeError: Cannot read properties of undefined (reading 'defaultPrevented')`

image

annavik:
> Awesome! It looks seamless, except for I am unable to open the dialog for "register new deployment"
> 
> `Uncaught TypeError: Cannot read properties of undefined (reading 'defaultPrevented')`
> 
> image

Oops! I had forgot to sync with main where I fixed this

Also added some new form stuff for delete, in that case we probably want to have a complete form to reuse

mihow (approved):

PR #270, opened 2023-10-05
Setup project forms

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/233

Note: Let's change the base branch before merge!

### Summary
- Make it possible to create a new project
- Make it possible to edit projects
- Make it possible to delete projects

### Screenshots
Screenshot 2023-10-05 at 15 29 32
Screenshot 2023-10-05 at 15 29 52
Screenshot 2023-10-05 at 15 30 02


ui/src/pages/project-details/project-details-form.tsx
```diff
@@ -0,0 +1,120 @@
+import { FormController } from 'components/form/form-controller'
```
annavik, line 1: @mihow, here is the form, using all the helpers from previous PR:s. I have tried to find a good balance between flexibility and simplicity, by using smaller building blocks rather than one massive super generic form component. I'm a bit inspired by the Radix UI components (checkout the code example): https://www.radix-ui.com/primitives/docs/components/accordion

annavik:
@mihow is BE ready to assign current user as project owner? I saw something about that in the related issue

mihow:
Exciting! Yes we should do a few permissions checks here:

- Only show create if the user has   `"user_permissions": ["create"]` on the projects list endpoint (which should be available for any logged-in user)
- Only show edit & delete if you have the update/delete user_permissions (which _will_ be project owners only but yes that needs BE. Use the existing permissions right now). 
- Same for deployments: the edit & delete permission will only be available if the current user is an admin if the deployment's project.
- Perhaps move the "Register new project" button to the bottom of the projects grid?

mihow (changes requested):
See comment about permissions checks. Otherwise looks awesome!

PR #271, opened 2023-10-05
Make it possible to delete deployment

annavik:
Make it possible to delete a deployment.

Screenshot 2023-10-05 at 16 45 14

### DIFF
--- a/ui/src/data-services/models/deployment.ts
+++ b/ui/src/data-services/models/deployment.ts
@@ -10,8 +10,7 @@ export class Deployment {
   }
 
   get canDelete(): boolean {
-    // TODO: Check delete permission instead of update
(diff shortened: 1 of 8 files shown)


ui/src/data-services/models/deployment.ts
```diff
@@ -9,6 +9,11 @@ export class Deployment {
     this._deployment = deployment
   }
 
+  get canDelete(): boolean {
+    // TODO: Check delete permission instead of update
+    return this._deployment.user_permissions.includes(UserPermission.Update)
```
mihow, line 14: The delete permission is actually available in the API! It shows for superusers only
image

mihow (changes requested):
Switch update permission to delete permission.

mihow (approved):

PR #274, opened 2023-10-11
New model & views for manually uploading example captures

mihow:


mihow:
@mihow decide what to do with the SourceImageUpload instances after they are uploaded & converted in to regular SourceImages. Some options:

- Delete the instances (caveat, they will disappear from the deployment Edit modal)
- Update the delete behavior to CASCADE and keep the instances. If they are deleted, then the created SourceImage will be deleted as well.
- Change the model name & purpose to `TestSourceImage` and let the user always treat them differently (e.g. for testing new models)
- Delete them, but the created SourceImages get a new `test_image=True` or `manually_uploaded=True` attribute. And then update the API / frontend to show regular SourceImages, but filtered by that attribute.


@annavik some decisions for the backend!

annavik:
> @mihow decide what to do with the SourceImageUpload instances after they are uploaded & converted in to regular SourceImages. Some options:
> 
> * Delete the instances (caveat, they will disappear from the deployment Edit modal)
> * Update the delete behavior to CASCADE and keep the instances. If they are deleted, then the created SourceImage will be deleted as well.
> * Change the model name & purpose to `TestSourceImage` and let the user always treat them differently (e.g. for testing new models)
> * Delete them, but the created SourceImages get a new `test_image=True` or `manually_uploaded=True` attribute. And then update the API / frontend to show regular SourceImages, but filtered by that attribute.
> 
> @annavik some decisions for the backend!

Hmm, good question! I'm not really sure what I think here yet... Perhaps we need to define more clearly what is an "example capture". Maybe we should skip that term and just call them "manually uploaded captures"?

From a user perspective, if images are about to be permanently deleted perhaps we should use the trash icon instead of the cross. Also introduce a confirm step?

mihow:
@annavik I merged this branch into the UI branch so that I can test both together. Closing in favor of #281

Issue #219, opened 2023-08-31
Add object-level permissions for Projects & related entities


mihow:
All API endpoints should only list or provide access to objects that the current user is allowed to see. And each object should specify if the current user can edit or take other actions on it. 

Currently each object provides a simple list of permissions in the API response based on if the user is logged in or is a super user.

Consider using the Guardian package to help handle this
https://www.django-rest-framework.org/api-guide/permissions/#djangoobjectpermissions
https://github.com/rpkilby/django-rest-framework-guardian

Reminder to consider adding an Organization entity for which users belong to and inherit permissions from. Multiple projects belong to an Org. Users can modify any project in org by default. 

Consider looking at https://github.com/keycloak/keycloak
https://django-keycloak.readthedocs.io/en/latest/
https://github.com/Ouranosinc/Magpie

mihow:
Related to #354

PR #275, opened 2023-10-11
Improve Project-User Management and Accessibility

mihow:
This pull request enhances how projects and project members (users)  are managed and accessed. It introduces a many-to-many relationship between projects and members, ensuring that users are automatically assigned to a project when they create it. Django admin panel is updated to allow manual assignment of members to a project. A filtering by `user_id` is added to the `/projects` to filter projects that the current user is involved in.My Projects tab  is added at the top of the home page to make it easier for users to find and access their projects.


mihow:
- Every view query and nested serializer query should have a filter to enforce that the current user is owner of the associated project
- Consider prefixing all api endpoints by /project/<id>/

mihow:
All queries should include objects from public projects as well.

mohamedelabbas1996:
### Changes in This PR
- [x ] Add many-to-many relationship between projects & users - make a new migration & commit file
- [ x] Auto-assign a user to a project when they create it
- [ x ] Expose the projects a user belongs to in the user profile API endpoint (`/me`)
- [ x ] Update Django admin for projects to manually add users to a project until public UI is ready
- [ x] Add tests:
  - [ x] Ensure the current user is added to a project when it is created via the API
  - [ x] Ensure the current user gets only the projects they're involved in as (owner or user) through /users/me/ endpoint.
- [ ] See possible UI for "My projects" section on top of the home page
  - [ ] Make it easier to find your project ([#518](#))

---

## Follow-Up Tasks
- [ ] Use the upcoming project edit screen for managing project members
- [ ] Implement full permission checks for access to projects and related data
  - [ ] Add true object-level permissions ([#219](#))
  - [ ] Consider moving permissions to the user object ([#354](#))
  - [ ] Add tests for permissions and access to different actions
- [ ] Rework project structure and clarify the collection concept ([#639](#))
- [ ] Plan roles for projects (e.g., annotator, data processor, admin) - roles can come later

mohamedelabbas1996:
Closes #402

mihow:
Thanks for collaborating on this PR @mohamedelabbas1996 and @annavik! I look forward to reviewing and thinking about your suggestions.

annavik:
Frontend updates pushed. Me and @mohamedelabbas1996 solved it by adding tabs to the header:

Screenshot 2025-01-22 at 17 46 55
Screenshot 2025-01-22 at 17 47 06

ui/src/pages/projects/projects.tsx
```diff
@@ -2,17 +2,32 @@ import { useProjects } from 'data-services/hooks/projects/useProjects'
...
 
 export const Projects = () => {
+  const { user } = useUser()
+  const [selectedTab, setSelectedTab] = useState(
+    user.loggedIn ? TABS.USER_PROJECTS : TABS.ALL_PROJECTS
+  )
   const { pagination, setPage } = usePagination()
+  const filters =
+    user.loggedIn && selectedTab === TABS.USER_PROJECTS
+      ? [{ field: 'public', value: 'false' }]
```
annavik, line 27: @mohamedelabbas1996 here is where the params we pass to the backend can be tweaked

annavik:
For the filter param name `public`, I think it's slightly misleading since user projects can also be public. I think this filtering is more related to membership rather than visibility. But I'm also fine keeping things as is, just a minor! :)

If you change the filter name, I added a comment (to the PR) so you can see where in the FE code to tweak.

@mihow @mohamedelabbas1996

mohamedelabbas1996:
Thanks @annavik for adding the changes on the frontend. For the filter name,  we can rename it to `current_user_projects`. I think this  makes more sense. @mihow  what are your suggestions?

ami/main/admin.py
```diff
@@ -68,6 +86,12 @@ def _remove_duplicate_classifications(self, request: HttpRequest, queryset: Quer
 
     actions = [_remove_duplicate_classifications]
 
+    # Make the 'users' field optional
+    def formfield_for_manytomany(self, db_field, request, **kwargs):
+        if db_field.name == "users":
```
mihow, line 91: This shouldn't be needed. Try adding `blank=True` to the users ManyToManyField on the Project model definition. The `blank` parameter is for Django validation only. https://github.com/RolnickLab/antenna/pull/275/files#diff-107580e5c15abf62fdda9c22724aeee42526be32b13db3d615f085a61693b457R106

ami/main/admin.py
```diff
@@ -56,7 +56,25 @@ class BlogPostAdmin(admin.ModelAdmin[BlogPost]):
...
+    filter_horizontal = ("users",)
+
+    @admin.display(description="Project Users")
+    def get_users(self, obj):
+        return ", ".join([user.email for user in obj.users.all() if user])
+
+    fieldsets = (
+        (None, {"fields": ("name", "description", "priority", "active")}),
+        (
+            "Ownership & Access",
```
mihow, line 71: Nice I like this fieldset!

ami/main/api/views.py
```diff
@@ -121,6 +122,15 @@ class ProjectViewSet(DefaultViewSet):
     serializer_class = ProjectSerializer
     pagination_class = ProjectPagination
 
+    def get_queryset(self):
+        qs: QuerySet = super().get_queryset()
+        # Filter projects for the current user if `public` query parameter is `False`
```
mihow, line 127: I agree that filtering by `user_id` instead of public feels more intuitive at this time. And also we don't have a good definition of what makes a Project "public". But only the `request.user` should be able to filter by the `user_id` OR a superadmin. Check out this permissions check helper here:

https://github.com/RolnickLab/antenna/blob/14e9ec011ffffd66a76a15863f0543ea22a6fc79/ami/base/permissions.py#L11-L13

ami/main/models.py
```diff
@@ -102,7 +102,8 @@ class Project(BaseModel):
     name = models.CharField(max_length=_POST_TITLE_MAX_LENGTH)
     description = models.TextField()
     image = models.ImageField(upload_to="projects", blank=True, null=True)
-
+    owner = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name="projects")
+    users = models.ManyToManyField(User, related_name="user_projects")
```
mihow, line 106: what about `members` instead of users? we will add roles later for editor, annotator, admin, etc

ami/main/admin.py
mohamedelabbas1996: Thanks for pointing this out

mihow:
@mohamedelabbas1996 there is a chance for some inconstancy when retrieving "my projects" since there is both a project owner and project members. Some ideas to address this:

- When filtering projects by `user_id`, search both members and the project owners
- Add project owner as a member in the Project save method, if they are not already a member

My recommendation is to add a new reusable method to to Project QuerySet class that can be used everywhere. `Project.objects.can_view(user)` and `Project.objects.can_edit(user)` or `can_access()` or `filter_by_user()`

It takes a little more code to setup a QuerySet for a model the first time, but it's a pattern that I would like to continue using because it keeps the logic in one place. Here is one example for the SourceImage model. You define a QuerySet, a Manager and then assign the manager to the Model.

https://github.com/RolnickLab/antenna/blob/14e9ec011ffffd66a76a15863f0543ea22a6fc79/ami/main/models.py#L1170-L1171

https://github.com/RolnickLab/antenna/blob/14e9ec011ffffd66a76a15863f0543ea22a6fc79/ami/main/models.py#L1194-L1196

https://github.com/RolnickLab/antenna/blob/14e9ec011ffffd66a76a15863f0543ea22a6fc79/ami/main/models.py#L1233

mohamedelabbas1996:
Screenshot 2025-01-22 at 18 32 28


Screenshot 2025-01-22 at 18 30 12

mohamedelabbas1996:
Now project members can be managed from the Admin page

mohamedelabbas1996:
> @mohamedelabbas1996 there is a chance for some inconstancy when retrieving "my projects" since there is both a project owner and project members. Some ideas to address this:
> 
>     * When filtering projects by `user_id`, search both members and the project owners
> 
>     * Add project owner as a member in the Project save method, if they are not already a member

I opted to the owner to members if not already exists

mohamedelabbas1996:
> My recommendation is to add a new reusable method to to Project QuerySet class that can be used everywhere. `Project.objects.can_view(user)` and `Project.objects.can_edit(user)` or `can_access()` or `filter_by_user()`
> 
> It takes a little more code to setup a QuerySet for a model the first time, but it's a pattern that I would like to continue using because it keeps the logic in one place. Here is one example for the SourceImage model. You define a QuerySet, a Manager and then assign the manager to the Model.

Added the model manager. Thanks for the recommendation!

ami/main/api/views.py
```diff
@@ -121,6 +123,18 @@ class ProjectViewSet(DefaultViewSet):
     serializer_class = ProjectSerializer
     pagination_class = ProjectPagination
 
+    def get_queryset(self):
+        qs: QuerySet = super().get_queryset()
+        # Filter projects by `user_id`
+        user_id = self.request.query_params.get("user_id")
+        if user_id:
+            user = User.objects.filter(pk=user_id).first()
+            if not (user == self.request.user or is_active_staff(self.request.user)):
```
mihow, line 132: Active "staff" is actually everyone who has edit permissions. Superusers are a different boolean field. But how about we simplify and just check for the matching user for now? Since superusers can see anything in the Django admin.

ami/main/api/views.py
```diff
@@ -130,6 +144,37 @@ def get_serializer_class(self):
...
+    def perform_create(self, serializer):
+        # Check if user is authenticated
+        if not self.request.user or not self.request.user.is_authenticated:
+            raise PermissionDenied("You must be authenticated to create a project.")
+
+        # Add current user as project owner
+        project = serializer.save(owner=self.request.user)
+
+        # Add owner to project members if not already a member
+        if not project.members.filter(id=self.request.user.id).exists():
```
mihow, line 156: I would add this to the Project.save() method instead, because there are other places where the owner can be set or changed. Like if you add a project in the Django admin, the owner is not made a member.

ami/main/models.py
```diff
@@ -95,14 +95,28 @@ def create_default_research_site(project: "Project") -> "Site":
     return site
 
 
+class ProjectQuerySet(models.QuerySet):
+    def filter_by_user(self, user):
+        """
+        Filters projects to include only those where the given user is a member.
+        """
+        return self.filter(members=user)
```
mihow, line 103: Since you made this queryset method (thank you!), it's pretty easy to filter by member or owner.

`self.filter(models.Q(members=user) | models.Q(owner=user)).distinct()`

ami/main/models.py
```diff
@@ -95,14 +95,28 @@ def create_default_research_site(project: "Project") -> "Site":
...
+
 @final
 class Project(BaseModel):
     """ """
 
     name = models.CharField(max_length=_POST_TITLE_MAX_LENGTH)
     description = models.TextField()
     image = models.ImageField(upload_to="projects", blank=True, null=True)
-
+    owner = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name="projects")
```
mihow, line 118: I'm glad we kept owner in this PR

ami/main/tests.py
```diff
@@ -882,3 +882,25 @@ def test_project_devices(self):
         exepcted_device_ids = {device.id for device in Device.objects.filter(project=project)}
         response_device_ids = {device.get("id") for device in response_data["results"]}
         self.assertEqual(response_device_ids, exepcted_device_ids)
+
+
+class TestProjectOwnerAutoAssignment(APITestCase):
```
mihow, line 887: Thanks for the tests! remember that `is_staff` is different than `is_superuser`. We are using `is_staff` to give a user edit permissions to everything in the app. `is_superuser` are the only users who can delete some objects (like jobs). We should review this in the next permission PRs.

ui/src/pages/projects/projects.tsx
```diff
@@ -1,18 +1,36 @@
...
+export const TABS = {
+  USER_PROJECTS: 'user-projects',
+  ALL_PROJECTS: 'all-projects',
+}
 
 export const Projects = () => {
+  const { user } = useUser()
+  const { userInfo} = useUserInfo()
+
+  const [selectedTab, setSelectedTab] = useState(
```
mihow, line 24: @annavik can we add a URL path or param so we can link to all projects when necessary? With the new feature, if someone sends someone else the projects link, they likely won't see the same projects.

ami/main/tests.py
mohamedelabbas1996: @mihow  Thank you for the clarification!

mohamedelabbas1996 (commented):
@mihow  I 've tried to move adding project owner to members to Project.save method. It works well from the api but it doesn't auto-add the owner if we create a project from the admin page

ami/main/migrations/0007_project_owner.py
```diff
@@ -0,0 +1,25 @@
+# Generated by Django 4.2.2 on 2023-10-11 02:15
```
mihow, line 1: One more thing, will you squash these migrations into one or two? `python manage.py squashmigrations first_one last_one`. That will remove some clutter and extra steps when we merge into main (instead of adding the field, then renaming the field). I only do this for migrations that haven't modified live data yet (migrations in an branch like these).

ami/main/models.py
```diff
@@ -116,6 +130,12 @@ class Project(BaseModel):
     devices: models.QuerySet["Device"]
     sites: models.QuerySet["Site"]
     jobs: models.QuerySet["Job"]
+    objects = ProjectManager()
+
+    def ensure_owner_membership(self):
```
mihow, line 135: @mohamedelabbas1996 this is what I was thinking, but also  things may change in your next PRs related to permissions. I think this is good to go now!

PR #277, opened 2023-10-11
[Minor]: Fix delete permission for projects

annavik:
Also did some testing with different types of users, BE permissions works like a charm @mihow!

Not logged in:
Screenshot 2023-10-11 at 16 45 01

Logged in with regular user:
Screenshot 2023-10-11 at 16 45 40

Logged in as super user:
Screenshot 2023-10-11 at 16 46 18


mihow (approved):
Awesome!

PR #281, opened 2023-10-13
Upload example captures from web interface

annavik:
In this PR we make it possible to add and remove example captures from deployments. It's possible to upload 
one or many at a time.

Screenshot (last card shows upload in progress):

Screenshot 2023-10-16 at 14 02 49


mihow:
I love how the errors are handled from backend validation! Thank you
image

ui/src/pages/deployment-details/deployment-details-form/section-source-images/section-source-images.tsx
```diff
@@ -71,26 +68,7 @@ export const SectionSourceImages = ({
...
         </FormRow>
-        {status === Status.Connected ? (
-          <>
-            <FormRow>
-              <InputValue
-                label={translate(STRING.FIELD_LABEL_CAPTURES)}
-                value={deployment.numImages}
-              />
-              <InputValue
-                label={translate(STRING.FIELD_LABEL_EXAMPLE_CAPTURES)}
```
mihow, line 82: I'm thinking we should remove this count. The "Example captures" become regular captures after you upload them and I don't think we should expose info about this proxy model.

image

mihow (changes requested):
Works great! One small request left in my comments. Then also one lager design decision we need to make related to the delete behavior. Currently deleting the example images will have no effect on the created source images after they are uploaded. See https://github.com/RolnickLab/ami-platform/pull/274#issuecomment-1763718715

annavik:
PR is updated with some tweaks tweaks!

Some remaining questions:
- Should we restrict file size client side?
- Should we restrict max number of images client side?
- On upload/delete error, we want to show error info and make it possible to retry. Which button option do you prefer? In both cases we show the error info on hover and trigger retry on click.
Screenshot 2023-10-16 at 13 52 50
<br/>
Screenshot 2023-10-16 at 13 54 13

mihow:
> PR is updated with some tweaks tweaks!
> 
> Some remaining questions:
> 
> * Should we restrict file size client side?
> * Should we restrict max number of images client side?
> * On upload/delete error, we want to show error info and make it possible to retry. Which button option do you prefer? In both cases we show the error info on hover and trigger retry on click.
> 
> Screenshot 2023-10-16 at 13 52 50
> Screenshot 2023-10-16 at 13 54 13

- You can use the same maximum file size constant as for the user icon and project image. I increased it to 10MB! We will have auto thumbnailing eventually... 
- Let's set the limit based on when the UI starts to break down. 20? Then provide a message saying "To upload more than N images you must configure a data source. See Help."
- For the Retry method. I like seeing "Retry" in a button. But I kind of like button in white. What about a red warning triangle on a white button that just says Retry?
- Will you make the data source field not-required?

mihow:
Oh new issue! The Deployment needs to exist before the Example Capture is saved. Can we hide or disable the image upload section and say "Save deployment before uploading images"?

mihow:
I don't know how to recreate this one, but it happened once where I uploaded 2 successfully, then the 3rd failed, but all of them showed the Retry button!
image

mihow:
@annavik 

> From a user perspective, if images are about to be permanently deleted perhaps we should use the trash icon instead of the cross. Also introduce a confirm step?

I think yes! Let's rename to "Test captures" and use the trash icon with a confirm. The backend is now deleting the uploaded capture if the source image is deleted and vice versa. 

The "Test captures" collection could also be used when creating a job (Select the "Test captures" collection to test models on this hand-picked subset)

annavik:
> I don't know how to recreate this one, but it happened once where I uploaded 2 successfully, then the 3rd failed, but all of them showed the Retry button! image

Good catch, thanks!! Will check it out

mihow:
Small design thought: Use a white button for the Retry buttons, but with the red warning triangle inside. I could go either way.

mihow (approved):

PR #285, opened 2023-10-19
Initial support for object detection ML backend

mihow:
- Allow filtering captures in the API that have or don't have detections yet (this is a basic, initial way for the ML backend to know which captures remain to be processed). 
- Show detection ID label in the session captures detail if a detection does not yet have a determination


ui/src/data-services/models/capture.ts
```diff
@@ -20,8 +20,8 @@ export class Capture {
       this._detections = capture.detections.map((detection: any) => ({
         bbox: detection.bbox,
         id: `${detection.id}`,
-        label: detection.occurrence.determination.name,
-        occurrenceId: `${detection.occurrence.id}`,
+        label: detection.occurrence ? detection.occurrence.determination.name : detection.id,
```
mihow, line 23: @annavik How did I do?

annavik: Perfect! 👌

ui/src/data-services/models/capture.ts
```diff
@@ -20,8 +20,8 @@ export class Capture {
...
+        occurrenceId: `${detection.occurrence?.id}`,
```
annavik, line 24: If occurrence is not defined, occurrenceId will be set to string "undefined" here. That is probably why could see all of the detections being selected yesterday. Haha but perhaps it's a feature??

annavik: I pushed a commit where we make occurrence ID optional for capture detections.

In UI this would mean if no occurrence is present, we can't mark the detection as selected (since selections are currently based on occurrence ID). Also, tooltip label will not be a link to occurrence details if we have no occurrence.

Feel free to nuke the commit if not helpful!

mihow: Thank you!

PR #286, opened 2023-10-19
Playback controls + trigger job for capture + job forms and more!

annavik:
### Summary of changes:
- Make it possible to create new jobs from UI
- Make it possible to delete jobs from UI
- Some new columns in jobs collection view
- Make it possible to start and cancel jobs from UI
- Update job details view with more fields and elements
- Update session details view with "Process now" quick action
- Update session details view with capture job details
- Adapt FE representation of jobs, for example to not rely on config
- Add playback control to session details view
    - Setup threshold slider
    - Keep track of current settings in browser
    - Pass threshold value with with capture collection request
    
This PR also includes some BE changes to support things mention above! 🌟
    
### Screenshots:
Screenshot 2023-11-08 at 17 27 05
Screenshot 2023-11-08 at 17 28 29
Screenshot 2023-11-08 at 17 29 06
Screenshot 2023-11-08 at 17 29 18


annavik:
Progress so far. Controls are not affecting anything yet, I have just prepared the UI. What do you think about the slider thumb "shadow" @mihow? On click it will set the thumb to the recommended value.

Screenshot 2023-10-19 at 16 41 25
Screenshot 2023-10-19 at 16 41 41

annavik:
Some backend related questions:
- Do we need logic for when to show the confidence threshold control and not? For example if data is not processed yet, does it make sense to show?
- We talked about the recommended value should be decided by current model. Can I get that value somehow, or is just 0.6 ok to use for now?
- What should I do with the confidence threshold value? Do some client side filtering or pass with BE calls?
- Let me know when and how I can trigger that simple capture job :)

mihow:
This is looking great!!

Let's change the word "confidence" to "score" to satisfy the critics for now. And change "recommended threshold" to "default threshold"

- Do we need logic for when to show the confidence threshold control and not? For example if data is not processed yet, does it make sense to show?

Great question. Let's check the `detections_count` property on the session detail (or list) response. If it's greater than zero, show the threshold bar.

- We talked about the recommended threshold value should be decided by current model. Can I get that value somehow, or is 0.6 ok to use for now?

It will come from the settings of currently selected algorithm. But that's not available yet. I think it's fine to use a constant set to 0.6 and use that as a the global default.

- What do you suggest I do with the confidence threshold value? Some client side filtering or pass with BE calls?

It should be added as a parameter to the API call for captures under a session:
`/api/v2/captures/?limit=200&offset=0&event=1&classification_threshold=0.6`

This doesn't do anything, yet! but I will add asap. 

- Let me know when and how I can trigger that simple capture job :)

For now you can create a job by posting to https://api.dev.insectai.org/api/v2/jobs/
Set the status to PENDING. 
Save the returned job ID and add a status indicator to the "Process capture" button  
You should also see it show under the Jobs section
Play with changing the status by automatically PUTing to the job detail endpoint and updating the status 

image

annavik:
**Some updates:**
- Updated copy and logic for the slider as we discussed above.
- On process button click, we now call backend. After a job is created, a button to inspect job details will show up next to the process button.
- In job table view, new jobs will be listed. It's also possible to delete jobs from there (if user has permissions).
- Removed project column from job table since we are already in the context of the project.

**Thoughts:**
I wonder if you have ideas on what user permissions to check for creating a job in session details view. Might be unnecessary to check job collection endpoint here, which I would normally do?

**Next:**
I think the next step for us might be to pass capture details when the job is created? I also need a way to know if a capture is associated with a job.

Also, I will make sure to pass confidence threshold with the capture call, not doing that yet.

mihow:
Will you add a permissions check for creating a new job?

image

mihow:
@annavik I implemented the confidence threshold filtering on the backend. The UI makes a new query when you move the slider, however only the image in view is updated. Can you update the list on the right so that the green bars change? We should update the charts as well, although I haven't tested that update on the BE yet. 

We could get tricky to make it feel much more responsive:
- When the slider is moved, don't make an API call, just filter the detections on the current capture with javascript. 
- Add a debounce function, so after the slider hasn't moved for a 1-2 seconds, call the backend and then update the list of captures and other data. 

image

This is much less priority then getting jobs working! So I'm okay to move it to it's own PR after this one is merged. The slider is working now!

mihow:
@annavik I've merged the new backend for background jobs. So you can now create jobs and see the progress bar changed -- if you put a number in the Delay field and check Queue Immediately.

mihow:
@annavik I added a new job status "Failed" and fix some error handling in the background tasks. I think some tasks you saw getting stuck had actually failed, but they weren't reporting. 

image

I also added basic logging to jobs! There are two new fields, `logs[]` and `errors[]` on each job instance. 

image

I tried to keep it simple and lumped all logs under the main job, rather than underneath stages. Can you add a section under job details that displays these log strings? Or another idea? For the error messages, we could show the latest error when you hover over the Failed status icon. But what ever is simple to implement.

mihow:
@annavik `source_image_single` now has an `event_id` you can use to link to the capture from the jobs detail page
image

mihow:
@annavik Will you remove references to the `job.config` object and see if stages can still render? Then I will remove that from the backend when ready.

annavik:
> @annavik Will you change the permissions check for delete to only check for "delete". That permission is only available to super users. And I am realizing that it could lead to chaos of users delete jobs that are attached to results.

I think I do this, are you seeing it differently? Let me know how to reproduce! :)

annavik:
> @annavik Will you remove references to the `job.config` object and see if stages can still render? Then I will remove that from the backend when ready.

Yes this is handled! I realised without the config I have no way to access stage params, so for now stages are empty. Let 's discuss what info to show here instead.

annavik:
Here is a suggestion on how to show job logs and errors, let me know what you think! Error log is simulated.

Screenshot 2023-11-08 at 16 02 40
Screenshot 2023-11-08 at 16 02 48
Screenshot 2023-11-08 at 16 04 27

What do you think of the log order, I think I prefer reversed?

annavik:
I have made some small tweaks to status details. This what we show now:
Screenshot 2023-11-08 at 16 26 19
Screenshot 2023-11-08 at 16 30 02

Do you still want me to replace with last message from error logs if present, or is that something better for BE to handle?

mihow:
That's a great idea to use the status_label, I will update that value from the BE!

mihow (approved):
I think this is plenty to work with! I added the small remaining tasks to a new ticket. https://github.com/RolnickLab/ami-platform/issues/310


Merge when you are ready!

PR #241, opened 2023-09-13
Initial models & endpoints for human identifications

mihow:
Model & endpoints dedicated to human Identifications separate from machine learning Predictions / Classifications. We can find a way conveniently retrieve them both in the same list later.

New generic endpoints for listing & creating identifications. This is how a user will "suggest a new species" for an occurrence
- /identifications/ 

User names and images are available in the identifications responses

Adds identifications to Occurrence response
- /occurrences/<id>/

Adds basic search endpoint for Taxa. Hopefully fast enough to be used for autocomplete
- /taxa/suggest?q=<name part>

Adds parent taxa to most taxa representations to be used for the name tree component. Eventually this will include more parts.
- Family > Species

New attributes on an Occurrence:
```json
 "identifications": [
        {
            "id": 2,
            "details": "http://api.dev.insectai.org/api/v2/identifications/2/",
            "taxon": {
                "id": 6,
                "name": "Campaea perlata",
                "rank": "Species",
                "details": "http://api.dev.insectai.org/api/v2/taxa/6/",
                "parent": {
                    "id": 49,
                    "name": "Campaea",
                    "rank": "Genus",
                    "details": "http://api.dev.insectai.org/api/v2/taxa/49/",
                    "parent": {
                        "id": 48,
                        "name": "Geometridae",
                        "rank": "Family",
                        "details": "http://api.dev.insectai.org/api/v2/taxa/48/",
                    },
                },
            },
            "user": {
                "id": 2,
                "name": "MB",
                "image": null,
                "details": "http://api.dev.insectai.org/api/v2/users/2/",
            },
            "created_at": "2023-09-14T20:47:18.344420",
        }
    ],
    "predictions": [
        {
            "id": 10,
            "details": "http://api.dev.insectai.org/api/v2/classifications/10/",
            "taxon": {
                "id": 6,
                "name": "Campaea perlata",
                "rank": "Species",
                "details": "http://api.dev.insectai.org/api/v2/taxa/6/",
                "parent": {
                    "id": 49,
                    "name": "Campaea",
                    "rank": "Genus",
                    "details": "http://api.dev.insectai.org/api/v2/taxa/49/",
                    "parent": {
                        "id": 48,
                        "name": "Geometridae",
                        "rank": "Family",
                        "details": "http://api.dev.insectai.org/api/v2/taxa/48/",
                    },
                },
            },
            "score": 0.9270092845,
            "algorithm": {
                "id": 1,
                "name": "Latest Model",
                "version": "1.0",
[...]

PR #288, opened 2023-10-23
Add detailed page titles

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/196

We are basing the page titles on the current breadcrumbs, to be consistent. We just reverse them.

Example of some page titles after this update:

Screenshot 2023-10-23 at 16 58 37

### DIFF
--- a/ami/main/api/serializers.py
+++ b/ami/main/api/serializers.py
@@ -407,6 +407,8 @@ class TaxonListSerializer(DefaultSerializer):
             "occurrences",
             "occurrence_images",
             "last_detected",
+            "created_at",
+            "updated_at",
         ]
 
     def get_occurrences(self, obj):
--- a/ami/main/api/views.py
+++ b/ami/main/api/views.py
@@ -189,6 +189,22 @@ class SourceImageViewSet(DefaultViewSet):
         .order_by("timestamp")
         .all()
     )
+
+    def get_queryset(self) -> QuerySet:
+        queryset = super().get_queryset()
+        has_detections = self.request.query_params.get("has_detections")
+
+        if has_detections is not None:
+            has_detections = BooleanField(required=False).clean(has_detections)
+            queryset = (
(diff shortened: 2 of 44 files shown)


annavik:
We could easily be even more detailed here, for example include more information for all details views (not just sessions). Do you think that is a good idea or are titles starting to get too complex? @mihow

mihow:
Thanks so much @annavik. I do think we should make titles specific to each detail page. Then it will be much easier to navigate through your history and it will be helpful when pasting links in Slack, etc. And it will be much better for search indexes when we go live.

For many detail pages the title could just use the object id "#341 | Occurrences | Vermont ML | AMI Platform". For projects, deployments & species you can use their nice names. Actually even occurrences could be "Actias Luna #241"

annavik:
> Thanks so much @annavik. I do think we should make titles specific to each detail page. Then it will be much easier to navigate through your history and it will be helpful when pasting links in Slack, etc. And it will be much better for search indexes when we go live.
> 
> For many detail pages the title could just use the object id "#341 | Occurrences | Vermont ML | AMI Platform". For projects, deployments & species you can use their nice names. Actually even occurrences could be "Actias Luna #241"

Fixed! Added more specific titles for jobs, deployments, occurrences and species. Example:

Screenshot 2023-10-25 at 14 57 58

annavik:
The update also affects detail breadcrumb, although most of the time it will not be visible behind dialog. If we decide to go from dialog -> full page for more detail views it could be nice to have though.

Screenshot 2023-10-25 at 15 01 02

annavik:
@mihow I also included a small change to update description when a project is selected. Looks a bit nicer when we share links. From Slack:
Screenshot 2023-10-25 at 17 25 26

See #294

mihow (approved):
Love it! Thank you for adding the meta previews. We can add images eventually when we get thumbnailing working!

Issue #290, opened 2023-10-24
Fix column sorting for table views


mihow:
- Sorting by date & duration no long seem to have an effect (backend)
- Enable sorting to detections_count, captures_count, deployment & species (backend & frontend)
- Allow "Session" column to sort by date (frontend)

mihow:
@annavik I fixed sorting on the API level, I don't know when that broke! can you update the column header sorting in the UI when you have a chance?

annavik:
Some issues I noticed:
- https://api.dev.insectai.org/api/v2/events/?ordering=duration
- https://api.dev.insectai.org/api/v2/occurrences/?ordering=timestamp

Not using these sort settings for now!

annavik:
To see what sort settings we have, I have checked API docs:

<img width="1312" alt="Screenshot 2023-10-25 at 16 27 08" src="https://github.com/RolnickLab/ami-platform/assets/11680517/3ae8c48d-071e-4475-ac91-3e6f95c39224">

mihow:
@mihow to add additional sorting params in the API:

https://api.dev.insectai.org/api/v2/events/?ordering=duration
https://api.dev.insectai.org/api/v2/occurrences/?ordering=timestamp

mihow:
> @mihow to add additional sorting params in the API:
> 
> https://api.dev.insectai.org/api/v2/events/?ordering=duration https://api.dev.insectai.org/api/v2/occurrences/?ordering=timestamp

Fixed in #296

PR #291, opened 2023-10-25
Fix sorting in API

mihow:
Sorting was disabled in the backend due to the order of inheritance in our base view class.


mihow:
Part of #290

PR #292, opened 2023-10-25
Taxonomy import & display v2

mihow:
- Add new taxon rank choices
- Enable importing a taxa list from CSV
- Enable importing a list from a URL
- Accept any taxon rank in the import file and appropriately assign parents & ranks
- Better error handling & logging
- Add additional fields like BOLD taxon BIN id
- Better support for using standardized TaxonRank Enum objects instead of just strings

Updates for the exported taxon tree choices for Label Studio annotators:
- Exclude taxon rank from the display name (display name is used for exact lookups)
- Enable filtering the taxa tree to include or exclude specific ranks (e.g. make me a tree with only family, genus, species)
- Add tests for taxa tree and filtering

This will continue to be refined in subsequent PRs. Tests


mihow:
Feature idea after receiving feedback: Add method to process a species list with all of the cleaning methods, but don't apply the fixes and import. Instead flag the entries that would change and return the list to the user. This way the taxonomists can confirm each edit and fix themselves.

mihow:
Feature request: Add "region" field. For now we can use the list field.

PR #294, opened 2023-10-25
Update meta description on project change

annavik:
In future we might want to do more advanced meta tags (add Open Graph tags, etc.). Think there might be some libs we could look into as well here.

This is just a small thing that will update page description when project changes.


annavik:
Merging into title branch, I want to test the deployed preview :)

PR #295, opened 2023-10-30
Update pagination

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/289

### In this PR:
- Make page index part of URL (when changing page from default)
- Update pagination bar with more detailed page controls

### Logic for pagination bar:
- We always show first and last page.
- Current page is not clickable.
- We show current page together with 4 nearby pages. We try keep the current page in center. If current page is close to first or last page, we will sift the center to still make it possible to show 4 nearby pages.
- Dividers are shown after first page and before last page. The exception is if the nearby pages are close.

### Examples:
Page 0 of 33 is selected:
Screenshot 2023-10-30 at 16 51 47

Page 5 of 33 is selected:
Screenshot 2023-10-30 at 16 52 10

Page 32 of 33 is selected:
Screenshot 2023-10-30 at 16 52 43

### Questions:
- Should we make "per page" part of URL as well? First I thought yes and code is prepared, but since we do not provide such settings from UI, perhaps it's not necessary? We could consider some control in UI here, but not sure we want this to be tweaked by users?
- First page have index 0 (to have settings closer aligned with BE calls), but perhaps it makes more sense to users if we start on 1? Should I bump it (both for URL params and in UI)?


annavik:
I have replaced the info text with page settings as part if this PR, but actually maybe we could keep it. Do you think it's useful or is page info enough? @mihow 

Example:
Screenshot 2023-10-30 at 17 27 04

mihow:
This works great, thank you! In response to your questions.

- I agree about omitting `per_page` from the URL params. Let's keep it a fixed value for now.
- Yes let's start the page index at 1 for users sake
- I like keeping "showing 81-95 results out of 95". It's very helpful to see the total and which subset you are looking at, especially for sequential data (e.g. for sessions you can interpret it as "showing nights 81-95"). Also I prefer the page links on the right side.

annavik:
> This works great, thank you! In response to your questions.
> 
> * I agree about omitting `per_page` from the URL params. Let's keep it a fixed value for now.
> * Yes let's start the page index at 1 for users sake
> * I like keeping "showing 81-95 results out of 95". It's very helpful to see the total and which subset you are looking at, especially for sequential data (e.g. for sessions you can interpret it as "showing nights 81-95"). Also I prefer the page links on the right side.

PR is updated! :)

mihow (approved):
Works great! And so so helpful for our growing database! Thank you.

PR #315, opened 2023-11-13
Setup UI for pipelines (+ related tweaks)

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/309.

### Summary of changes:
- Add pipelines collection to project overview
- Setup detail view for pipelines, including stages
- Update job details to show stages params (semi related...)
- Add pipeline picker to session details
- Add capture details popover

### Screenshots:
Screenshot 2023-11-13 at 17 33 54
Screenshot 2023-11-13 at 17 34 20
Screenshot 2023-11-13 at 17 34 26
Screenshot 2023-11-22 at 16 10 40
Screenshot 2023-11-22 at 16 08 35


mihow:
Rather than showing the pipeline name in the select component, what if it always just says "Pipeline" and when you open it, it has a checkmark next to the selected name. Or I do like the short name as well. We can use it in URLs too. For now, maybe use the ID, then I can swap the fields later.

I like the 2 lines for the controls.

Also, in the designs there is a control button for filters. What if we hide _all_ controls until you click that? Then maybe we don't have to be so clever in fitting it all in. 

image

annavik:
Updated this PR with some new stuff:
- Make it possible to expand/collapse detailed controls (also we will not fetch capture details and list of pipelines until detailed controls are expanded)
- Show capture details in popover
- Require pipeline to be set before capture can be processed (do we want this?)
- Pass pipeline id when job is created from session details
- Make it possible to see detailed controls while list of captures is loading (for example when changing score threshold)

PR desc is updated with some screenshots!

mihow:
Thank you for adding the stage details to the jobs panel. I'm already using it for new stages!

image

mihow:
Can you link the pipeline name to the pipeline details modal?

mihow (approved):

PR #316, opened 2023-11-15
Migrate from CRA -> Vite + update Storybook + update Yarn

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/254

Followed pretty much this process: https://www.freecodecamp.org/news/how-to-migrate-from-create-react-app-to-vite/

Hade to do some special handling for SVG:s, Jest and ESLint. Also updating Storybook and Yarn.


annavik:
Some notes:
- I ended up updating Yarn as well. There are a lot of file changes because of this. See: https://yarnpkg.com/getting-started/qa I have tried updating `.gitattributes` as suggested to avoid massive diffs showing, but with no success.
- Deploy is failing because after the update, prod files will end up in folder /dist, not /build.

annavik:
Some notes:
- I ended up updating Yarn as well. There are a lot of file changes because of this. Some binary files are supposed to be checked in. See section "Which files should be gitignored?" in https://yarnpkg.com/getting-started/qa. I have tried updating `.gitattributes` as suggested to avoid massive diffs showing for binary files, but it did not help a lot. We also have huge diffs in `yarn.lock`.
- Deploy is failing because after the update, prod files will end up in folder `/dist`, not `/build`.

mihow (approved):
Thank you for taking the time to make these updates & cleanups.

PR #318, opened 2023-11-20
Lazy load Plotly

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/239

In this PR, we lazy load the plot component. We do this to extract Plotly lib from main bundle. Since loading the main bundle is render blocking, this will let users interact with the app quicker.

Chunks before:
Screenshot 2023-11-20 at 09 41 44

Chunks after (notice that plot chunk is way larger than rest of app):
Screenshot 2023-11-20 at 09 38 07

In UI, we will show loading spinners instead of plots while the plot chunk is loading (this will only happen first time the user visits a page where plots are present).
Screenshot 2023-11-20 at 09 42 51

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -11,11 +11,13 @@ import { CollectionDetails } from 'pages/collection-details/collection-details'
 import { Deployments } from 'pages/deployments/deployments'
 import { Jobs } from 'pages/jobs/jobs'
 import { Occurrences } from 'pages/occurrences/occurrences'
+import Overview from 'pages/overview/overview'
 import { Projects } from 'pages/projects/projects'
+import SessionDetails from 'pages/session-details/session-details'
 import { Sessions } from 'pages/sessions/sessions'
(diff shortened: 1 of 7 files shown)


ui/src/app.tsx
```diff
@@ -61,25 +58,11 @@ export const App = () => {
               </Route>
               <Route path="projects" element={<ProjectsContainer />} />
               <Route path="projects/:projectId" element={<ProjectContainer />}>
-                <Route
```
annavik, line 64: Tried out a similar approach in previous PR, but in that case I lazy loaded pages instead of just the plot component.

annavik:
Also, see improved Lighthouse score (before vs. after code splitting). First score is captured from https://github.com/RolnickLab/ami-platform/pull/305.

Screenshot 2023-11-20 at 10 09 59

Screenshot 2023-11-20 at 10 10 42

mihow (approved):
This is awesome, thank you! Can't believe plot.ly is 4mb!

PR #324, opened 2023-11-22
Setup error boundaries

annavik:
In this PR we setup a generic error boundary component. We can wrap this around any component to catch errors instead of having the application crash when something unexpected happens. It can be used for larger components, like page components, but also for smaller ones for more granular control. In the future, this can be nice to wrap around components we know can be fragile.

Example where error is catched on page level. The button "Retry" will reset the component to its initial state.
Screenshot 2023-11-22 at 22 37 32

Example where error is catched on component level (error message is visible on icon hover):
Screenshot 2023-11-22 at 22 38 08

Error message is visible on icon hover:
Screenshot 2023-11-22 at 22 38 10

### DIFF
--- a/ui/src/components/error-boundary/error-boundary.tsx
+++ b/ui/src/components/error-boundary/error-boundary.tsx
@@ -1,0 +1,39 @@
+import { Button } from 'design-system/components/button/button'
+import { Icon, IconTheme, IconType } from 'design-system/components/icon/icon'
+import { Tooltip } from 'design-system/components/tooltip/tooltip'
+import { ErrorInfo, ReactNode } from 'react'
+import { ErrorBoundary as _ErrorBoundary } from 'react-error-boundary'
+import { STRING, translate } from 'utils/language'
+import styles from './error-boundary.module.scss'
+
+const logErrorToService = (error: Error, info: ErrorInfo) => {
+  // TODO: Pass error to Sentry here
+  console.error(error, info)
+}
+
+const FallbackComponent = ({
+  error,
+  resetErrorBoundary,
+}: {
+  error: { message: string }
+  resetErrorBoundary: () => void
+}) => (
+  <div className={styles.wrapper}>
+    <Tooltip content={error.message}>
+      <div className={styles.iconWrapper}>
+        <Icon type={IconType.Error} theme={IconTheme.Error} size={24} />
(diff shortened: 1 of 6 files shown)


ui/src/components/error-boundary/error-boundary.tsx
```diff
@@ -0,0 +1,39 @@
+import { Button } from 'design-system/components/button/button'
+import { Icon, IconTheme, IconType } from 'design-system/components/icon/icon'
+import { Tooltip } from 'design-system/components/tooltip/tooltip'
+import { ErrorInfo, ReactNode } from 'react'
+import { ErrorBoundary as _ErrorBoundary } from 'react-error-boundary'
+import { STRING, translate } from 'utils/language'
+import styles from './error-boundary.module.scss'
+
+const logErrorToService = (error: Error, info: ErrorInfo) => {
+  // TODO: Pass error to Sentry here
```
annavik, line 10: @mihow here is where we would log errors to Sentry, so that you and me can get info about what happened and get a chance to make a proper fix :) Info objects should be pretty detailed

mihow: Oh that's great! Check for an invitation from Sentry. I just sent you a new one from our sponsored account.

mihow (approved):
This is fantastic, thank you. It is so much more helpful to have a single component crash rather than the whole page.

ui/src/components/error-boundary/error-boundary.tsx
annavik: Thanks! Sorry I forgot the last invite. FE app is now hooked up with Sentry 🚀

Screenshot 2023-11-23 at 09 56 42

ui/src/index.tsx
```diff
@@ -1,7 +1,13 @@
+import * as Sentry from '@sentry/react'
 import ReactDOM from 'react-dom/client'
 import { BrowserRouter } from 'react-router-dom'
 import { App } from './app'
 
+Sentry.init({
+  dsn: 'https://bdacb11d18ccce7135e11de82e017632@o4503927026876416.ingest.sentry.io/4505909755838464',
```
annavik, line 7: As far as I understand Sentry DNS is no secret, which is why I'm checking it in

PR #326, opened 2023-11-27
Add control for marking captures as starred

annavik:
Not starred capture:
Screenshot 2023-11-27 at 23 25 14

Starred capture (when possible to unmark as starred, we can have this clickable):
Screenshot 2023-11-27 at 23 25 25


mihow:
@annavik I added an unstar endpoint and attempted to implement it in the UI. I also tried to use a tooltip instead of a label. But looks like label is required on a Button, so maybe there is a better way to do that for accessibility?

Let me know how I did. 

Also what do you think about moving the heart to the main controls row, before the user clicks expand?

annavik:
> @annavik I added an unstar endpoint and attempted to implement it in the UI. I also tried to use a tooltip instead of a label. But looks like label is required on a Button, so maybe there is a better way to do that for accessibility?
> 
> Let me know how I did.
> 
> Also what do you think about moving the heart to the main controls row, before the user clicks expand?

Looks great! We have a separate `IconButton` component, I set it up like that to not make regular button component too complex. PR is updated!

mihow (approved):
I love it. Also the loading indicator works great, thank you.

PR #331, opened 2023-12-05
Introduce related entities

annavik:
**Summary of changes:**
- Set up generic entity type (to use for storage, sites and devices)
- Make it possible to list entities
- Make it possible to create, update and delete entity
- Setup entity dropdown component (to use in other forms, for example in deployment form)
- Hook up deployment with new entities

**Screenshots:**
Screenshot 2023-12-11 at 23 13 11

Screenshot 2023-12-11 at 23 13 01

Screenshot 2023-12-11 at 23 17 34

Screenshot 2023-12-11 at 23 14 26

**Notes for future:**
The entity type is actually something we could reuse also for other objects and delete some code! All objects seems to have at least id, name, created at, updated at and user permissions?

Also, after starting to extend the entity model for other objects, we can replace the pipeline picker completely with the newly created entity picker.


annavik:
I have started try hook up site and device with deployment, I'm keeping it simple for now. Is it something like this you had in mind?

Screenshot 2023-12-06 at 11 42 28

Screenshot 2023-12-06 at 11 42 37

If so, can we make BE accept `device_id` and `site_id` on deployment create and patch? And return device and site in deployment details response (including both id and name). @mihow 

Next up: data source!

annavik:
PR is updated with storage entity. Also we can connect entities to deployments now. I have updated PR desc and marked for ready for review, please give it a test run when you get a chance :) @mihow

PR #332, opened 2023-12-05
Conditionally enable Sentry

annavik:
Ref: https://docs.sentry.io/platforms/javascript/configuration/options/


mihow (commented):
Is it possible to specify NODE_ENV when building in Netlify?

PR #333, opened 2023-12-11
Add device, site and data source to Deployment responses

mihow:
Device & Research Site are available in the list view. Data Source is only available in the detail view.

Update the related entities with device_id, research_site_id, and data_source_id

`deploynent.data_source_uri` was moved to the nested attribute `deployment.data_source.uri`

image


annavik:
Perfect, thank you!!

mihow:
Fixes #344

PR #338, opened 2024-01-26
Add Google Analytics

mihow:


mihow:
@annavik I haven't added GA to a React UI before. Let me know if there is a better way to integrate the tag.

mihow:
@annavik it would be preferable to make the gtag code a variable (as well as the Sentry DSN). Even though those aren't sensitive values, the project should be configurable in different environments (and for other groups, since it is open source).

annavik:
I think this lib could be worth checking out: https://www.npmjs.com/package//react-gtm-module

Then we can easily pass GTM id as a variable (and have different values for different envs). Also have GTM more integrated with rest of the React app, for example if we want to configure custom events from other places in code, etc.

Let me know if you want me to give it a go! :)

annavik:
Depending on where app will launched, we might have to make sure to only initialize the tracking if we have cookie consent. I n Europe we have laws about it so always in the back of my mind, but probably not to worry about at this early stage!

mihow:
Hi yes that react-gtm-module looks great @annavik! It's very nice to trigger custom actions as well (like start/cancel Job). Hopefully this library will trigger page views automatically when there is a URL location change, because that will cover most interaction events.

mihow:
For the GDPR cookie consent, we will need that before people can even log in. Is there a React or Radix module for such consent forms? Since it's ubiquitous now.

annavik:
> For the GDPR cookie consent, we will need that before people can even log in. Is there a React or Radix module for such consent forms? Since it's ubiquitous now.

I have only made custom solutions in the past, seems the law is still being interpreted in some different ways. However for the popup UI there are definitely Radix stuff we can use!

There are also different categories of cookies to consider. Necessary (login), preferences (could be a language setting for example) and statistics (like the GTM cookie). Some you just need to inform about, some must be opt out by default, etc. It ca be a bit of a jungle!

I think we should perhaps get some legal advice here first?

mihow:
I think the basic GDPR cookie acceptance modal will be enough. I haven't had any legal problems with using google analytics in Europe on other apps, as long as we have that GDPR question. I prefer the ones that say "By using this site you agree to... " with a button that says "Okay" or "I understand" rather than "I agree" & "Disagree". Because you can't actually continue to use the website if you disagree.

mihow:
Merging this as-is so we can troubleshoot interactions during the beta preview phase this month. I added two tickets with TODOs based on the conversation in this PR.

https://github.com/RolnickLab/ami-platform/issues/351
https://github.com/RolnickLab/ami-platform/issues/350

PR #347, opened 2024-02-10
Fix session details navigation

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/329

Also including two small related fixes:
- Do not fetch capture details until capture id is set (I noticed we did one "capture/undefined" fetch on session details load)
- Fix broken session link in occurrence table


mihow (approved):
Thank you!

PR #348, opened 2024-02-10
Make thumbnails clickable

annavik:
Fixes https://github.com/RolnickLab/ami-platform/issues/313

Fixed in tables sessions, occurrences, species and captures. Only thumbnail itself will be clickable, buttons for previous/next thumbnail should work as before.


mihow:
Fantastic! Works for me and is very satisfying. @kpmcfarland

mihow (approved):

PR #359, opened 2024-02-20
Fix counts, Don't show occurrences from other projects

mihow:


mihow:
@annavik this includes two frontend fixes that could be nicer. We need to pass the project id to the API request in a species detail modal so it only shows occurrences from the correct project. Also the date & time links were broken, but I believe passing the timestamp to get the capturePageOffset is working!

PR #362, opened 2024-02-27
Fix timeout on session detail pages with many species

mihow:
It turns out a single query that uses annotations to count the number of taxa and occurrences for a night is much much slower than using separate queries.

This also was a chance to standardize how the `confidence_threshold` parameter is used. Most views should now look for this query parameter in the current request when counting or displaying occurrences or detected taxa. It falls back to the `DEFAULT_CONFIDENCE_THRESHOLD` in the Django settings, which is also an env var.


mihow:
@kpmcfarland This should fix the issue you are facing with May 31, 2022! https://app.preview.insectai.org/projects/18/sessions/2575

PR #366, opened 2024-03-12
Fix various onboarding issues

mihow:
Users have been unable to create a new Deployment on their own. This is mostly because a Device, Research Site and Data Source need to be created first before creating a Deployment. Those related models are not actually required, but the form was requiring something to be selected. This PR makes several updates to address this issue.

- Allow the Deployment form to be submitted when Device, Site & Data Source are blank
- Set the fields to not be required on the backend
- Create a default Device and Research Site for a Project when it is first created.

Other changes:
- Increase the file size of user uploads from 1MB, and make standard across the site
- Run existing doctests on utility functions, like the date parser from image file names.


ui/src/data-services/hooks/deployments/utils.ts
```diff
@@ -12,7 +12,9 @@ export const convertToFormData = (fieldValues: DeploymentFieldValues) => {
     longitude: fieldValues.longitude,
     research_site_id: fieldValues.siteId,
   }).forEach(([key, value]) => {
-    data.append(key, `${value}`)
+    if (value && value !== 'undefined') {
```
mihow, line 15: @annavik I am curious how these variables are getting converted to the string "undefined". Any ideas?

ui/src/pages/deployment-details/deployment-details-info.tsx
```diff
@@ -83,7 +83,7 @@ export const DeploymentDetailsInfo = ({
         <FormSection title={translate(STRING.FIELD_LABEL_SOURCE_IMAGES)}>
           <FormRow>
             <InputValue
-              label="Data source"
+              label={translate(STRING.FIELD_LABEL_DATA_SOURCE)}
```
mihow, line 86: @annavik The other select fields for Site & Device have a placeholder "Pick a value", but Data Source does not. I can't see why, any ideas?

mihow:
Hi @jonbeu this should fix the issue you faced when creating a new deployment. The site, device & data source are no-longer required -- but also a default device & site are created for each project now. I am anxious to see some test images from the AMI-Mini! -Michael

ui/src/pages/deployment-details/deployment-details-info.tsx
annavik: This is because if project has no data sources registered, it's not possible to "pick a value". Then we instead disable the control with label "Not available". Does it make sense? No problem to tweak this logic, let me know :)

If project has a storage registered:
Screenshot 2024-03-18 at 22 25 07

If project has no storage registered:
Screenshot 2024-03-18 at 22 28 13

ui/src/data-services/hooks/deployments/utils.ts
annavik: Actually, the problem is the line below. JavaScript behaves like this:
Screenshot 2024-03-18 at 22 38 53

You can try it out in the console!

annavik: If we want to make sure we will not pass the string "undefined" to server, we must make sure the value is not undefined before we try make a string out of it. Pretty much what you did, but the string comparison is not necessary. I would also make sure we will not ignore numeric values equal to 0 by making the check a bit more specific.

```suggestion
    if (value !== undefined) {
```

annavik: You know because
Screenshot 2024-03-18 at 22 44 50

mihow: > If we want to make sure we will not pass the string "undefined" to server, we must make sure the value is not undefined before we try make a string out of it. Pretty much what you did, but the string comparison is not necessary. I would also make sure we will not ignore numeric values equal to 0 by making the check a bit more specific.

This is what I started with, but somehow the undefined value is converted to a string before this function is called. So `value !== undefined` does not catch it :(

PR #367, opened 2024-03-18
Small fixes for deployment form logic

annavik:
Includes three small fixes:
- Cleanup logic for form data
- Ignore selected values not matching available options in pickers
- Do not treat data source as existing if no ID is set

### DIFF
--- a/ui/src/pages/overview/entities/entities-picker.tsx
+++ b/ui/src/pages/overview/entities/entities-picker.tsx
@@ -15,7 +15,7 @@ export const EntitiesPicker = ({
   const { entities = [], isLoading } = useEntities(collection, {
     projectId: projectId as string,
   })
-  const hasValue = entities.some((e) => e.id === value)
+  const validValue = entities.some((e) => e.id === value)
 
   return (
     <Select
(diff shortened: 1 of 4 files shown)


ui/src/pages/overview/entities/entities-picker.tsx
```diff
@@ -24,7 +25,7 @@ export const EntitiesPicker = ({
         label: e.name,
       }))}
       showClear={false}
-      value={value}
+      value={hasValue ? value : undefined}
```
annavik, line 28: This is because I noticed if no data source is set, backend still sets a value for it. But in FE it looks a bit strange to show "empty string" as selected.

Screenshot 2024-03-18 at 22 53 04

mihow: > This is because I noticed if no data source is set, backend still sets a value for it. But in FE it looks a bit strange to show "empty string" as selected.
> 
> Screenshot 2024-03-18 at 22 53 04

I fixed the bogus the data source! merging this PR, thank you.

mihow (approved):

PR #375, opened 2024-04-12
[AMI-267] Enable collections to be defined & populated from UI

mihow:
image

image

Description TBD!


mihow:
Hi @annavik! This started as a "simple" change so users can make and refresh populate Collections in the UI, since that is a missing piece for the batch processing workflow. But it turned out to be my largest contribution to the UI code yet! And I had a lot of catching up to do. I know this could be cleaner, but I am hoping to compromise and make something available to demo soon. We should probably discuss more about collections. They may be outgrowing the sub-tab under projects and ready for their own list view page!

After this, I am definitely look forward to re-using the backend types in the frontend. I tested that using our autogenerated openAPI schema -> typscript and it worked amazingly well. I will make a PR with an example soon. Also I am curious about using react-hook-form library more directly. It seems like there are some layers of abstraction/simplification that could be skipped? It was difficult for me to find where values are being transformed before being passed to the server. Anyway, take a look if you have a chance, but no worries if we have to wait until June.

mihow (commented):
Doing it! This also fixes an issue where the edit forms appear for non-logged in users

PR #377, opened 2024-04-17
[Draft] Explore and test the use of pixel area to determine detected insect size

mihow:
See comments


mihow:
This is a nice-to-have but needs additional work

**Must dos:**
- Move query from the view to a QuerySet method, or a dedicate function
- Test that it doesn't noticeably impact the list views
- Review the changes in #858 related to filtering by size. Assume we will have a post-processing step that filters by size. Can these two queries use the same function? If not okay, but check.

**Enhancements**:
- Add size filter to UI 
- Add fields for physical dimensions to each device type for the size of the camera target.
- Automatically convert pixel percentage to physical size if the physical size of the target is configured in the device type and display/filter by centimeters of moth size instead of pixel percentage. 
- Show average / max / min size in the taxon observed detail view

PR #379, opened 2024-04-20
Allow data sources to be synced from the UI

mihow:
Bugs to fix:

- [x] BE Update data source test endpoint to accept "subdir" & "regex" params. Return full S3 URI, and some connection details ("Found at least 1 file with this configuration")
- [x] Most Costa Rica bucket paths have a space in them (likely unintentionally)
- [x] Add real Test button under deployment data sources
- [x] Data source Regex field gets filled in with string "null"
- [x] Complete tests for MinIO (s3 compatible storage backend for local / docker-compose stack)
- [x] Zip files are added as captures when syncing? Should filter by file type.
- [x] Ensure presigned urls are not used if the public base url is specified (rather than new boolean field)
- [x] Fix MinIO implementation, add default bucket.
- [x] Document how to use MinIO in local dev. (use default user/pass, add entry to hosts file for now. add notes about debug logging and running tests with -K, --pdb and --failfast as well, upload files with minio:9001)
- [x] Creating a new processing jobs adds stages for "data_storage_sync"
- [x] Multiple copies of "data_source_sync" stage are added to all job types
- [x] Confirm that images are working with public_base_url. The bucket name appears to be missing if the base url does not end in a trailing slash and the deployment uses a subdir.
- [x] Make subdir required? or not?
- [ ] Storage source tab on project overview page needs to be updated to newer style of other entity tabs

Discovered bugs, may no longer be relevant:
- [ ] If a capture does not belong to an Event, the UI crashes (null does not have attribute 'id')
- [ ] Sorting sessions by session name or "most recent" doesn't seem to work. Sorting by "date" column does.

Several issues raised during this PR have been moved to related issues:
https://github.com/RolnickLab/ami-platform/issues/400
https://github.com/RolnickLab/ami-platform/issues/455
https://github.com/RolnickLab/ami-platform/issues/310
https://github.com/RolnickLab/ami-platform/issues/456
https://github.com/RolnickLab/ami-platform/issues/236

Configure S3 storage source for a project
image

Test the connection, see which deployments are using this storage
image

Configure the subdirectory for each deployment  & test the config
image

Start the sync
image

View status of sync job
image

See the number of captures increase for the deployment
image


albags:
Created a new issue [Sign up button #385](https://github.com/RolnickLab/ami-platform/issues/385) in GitHub

annavik:
Great work Michael! 🙌

I am looking into some FE tweaks now. Here are some questions that showed up:
- The "Total size" for the storage is always 0 bytes for me, is it a bug?
- I see the description field is gone from form, but present in table view, is the intention to keep or remove? Let me know and I fix!
- For a beginner like me, the difference between "Endpoint URL" and "Public base URL" wasn't super clear. I felt like I was guessing a bit when I added the values. Not really sure what to do about it haha (maybe a placeholder for the value?), but just wanted to mention it.
- What do you think about testing the connection from detail view, instead of from table view?
- In deployment details, the "Full URI" is missing some things, right? Also what do you think about updating it on edit, instead of on save?
- In deployment details, the button "Test" is disabled. Is the intention to check the connection here? Just want to double check before I fix.
- Regarding "Sync", I noticed the deployment must be saved before this can happen. I'm thinking I will add a similar description, like the one we have for sample captures.
- What do you think about calling "Sync" -> "Import captures" or "Sync captures"? To me that would be a bit more clear.
- The job details for the sync ended up a bit strange for me, but I think you are on this already from reading the PR description?

annavik:
@mihow I'm working on some tweaks on the UI to sync data source, I wonder if there is a way to know if the deployment is currently  up to date with the data source or not? Perhaps more "nice to have"!

mihow:
Thanks so much for looking at this one Anna! Here are responses to your questions:

> * The "Total size" for the storage is always 0 bytes for me, is it a bug?

This only populates once you sync. We could change the label to "Total size indexed" or something like that? 

> * I see the description field is gone from form, but present in table view, is the intention to keep or remove? Let me know and I fix!

Oh hmm, let's see if we can live without it! One less field?

> * For a beginner like me, the difference between "Endpoint URL" and "Public base URL" wasn't super clear. I felt like I was guessing a bit when I added the values. Not really sure what to do about it haha (maybe a placeholder for the value?), but just wanted to mention it.

Yes, a good place for descriptions or tooltips! "Endpoint URL" is the custom endpoint required if you are using anything other than Amazon's S3 storage service (DigitalOcean, Google, Azure, Compute Canada, etc). This is only used by the backend to talk to the S3 service. "Public base URL" is for the actual image URLs. This is optional, but if your images are public then they will load much faster if you provide a public base URL. Otherwise the backend assumes they are private and has to generate a temporary signed-URL for every image. We might be able to avoid this field and provide a checkbox instead ("public bucket") or something. 

> * What do you think about testing the connection from detail view, instead of from table view?

Yes please!!

> * In deployment details, the "Full URI" is missing some things, right? Also what do you think about updating it on edit, instead of on save?

Yes please!!!

> * In deployment details, the button "Test" is disabled. Is the intention to check the connection here? Just want to double check before I fix.

Yes!! I think we need an endpoint first though

> * Regarding "Sync", I noticed the deployment must be saved before this can happen. I'm thinking I will add a similar description, like the one we have for sample captures.

Okay, I suppose we could also save before sync? But fine to require saving first and provide messaging. It's a safe workflow to fall back on (saving & validating before taking related actions) but we may be able to make the workflow it smoother in some cases.

> * What do you think about calling "Sync" -> "Import captures" or "Sync captures"? To me that would be a bit more clear.

Absolutely! "Import captures" is good. But a couple notes: we aren't copying the images, only scanning them and updating our index/reference (I call it "indexing"). And If you sync again it will update & index any new images it finds. Right now I am not deleting any references if they are deleted from the data source. But I think we probably should. Otherwise the images will never load.

> * The job details for the sync ended up a bit strange for me, but I think you are on this already from reading the PR description?

Yes that's a todo for me! But also open to feedback on how they should be displayed. 

Another idea: 
We could consider adding a sync button at the data source level on the project detail page, which would create a sync job for all deployments that use that data source.

mihow:
We only can tell if it’s up to date when it syncs. We do store when it last
synced. So maybe “up-to-date, last checked 6hr ago”. If it’s within 24 hrs
it’s green?

mihow:
@annavik I finally added a method to pass a Deployment's subdir and/or regex filter to the test endpoint. I also added more data to the response including the `full_uri` that you can display to the user. Perhaps the "first_file_found" URL is nice to show as well? If a regex is passed it will search a maximum number of files. So in that case the connection can be successful, but not matching files will be found during the test. 

To check a subdir and regex_filter pass those as params in the POST data to the test endpoint:
https://api.beluga.insectai.org/api/v2/storage/1/test/

```
{
    "connection_successful": true,
    "prefix_exists": true,
    "latency": 0.06636643409729004,
    "total_time": 0.440143346786499,
    "error_code": null,
    "error_message": null,
    "files_checked": 10000,
    "first_file_found": "https://static.dev.insectai.org/ami-trapdata/vermont/RawImages/LUNA/2022/movement/2022_06_20/20220620220231-00-11.jpg",
    "full_uri": "s3://ami-trapdata/vermont/RawImages/LUNA/?regex=2023"
}
```

image

mihow:
I think we should merge this one soon and then move any remaining issues to a new ticket. The backend has already been deployed for some time now and isn't breaking anything we've noticed.

annavik:
> I think we should merge this one soon and then move any remaining issues to a new ticket. The backend has already been deployed for some time now and isn't breaking anything we've noticed.

Agreed! I synced the PR with main today and also made use of your BE updates to /test endpoint (thank you for that!). I think ok to merge this from FE side!

mihow:
I just noticed I can view these edit actions and open the edit form even when logged out

image

mihow:
@annavik I didn't get to merge this one yet, so adding some more feedback!

When testing the data storage connection on the Deployment edit screen, the results look the same even if the path doesn't exist. I think we should show the red icon with the question mark if the path isn't found, even if the connection is okay. Or! another state that is something in between? It would be nice to show a few more details too, like the latency and perhaps even a link the first image found (already in the API response see screenshots attached). Because the data syncing can take a long time, it's good to confirm everything before you start. Also there's one typo in the regex parameter, it should be `regex_filter` instead of `regexFilter`.

Good connection, path found
image

Good connection, path not found
image

Example of current error state
image

mihow:
@annavik I had to do a fair bit of troubleshooting to get the local S3 storage service (called "MinIO") working. But now we are a big step closer to being able to run the full stack locally and run end-to-end tests. I will document more about that later, but I had to celebrate after some hair pulling! We should be able to quickly troubleshoot more issues and scenarios related to data syncing now. 

image

image

image

annavik:
> @annavik I had to do a fair bit of troubleshooting to get the local S3 storage service (called "MinIO") working. But now we are a big step closer to being able to run the full stack locally and run end-to-end tests. I will document more about that later, but I had to celebrate after some hair pulling! We should be able to quickly troubleshoot more issues and scenarios related to data syncing now.
> 
> image
> 
> image
> 
> image

Nice, great work!! Very satisfying to see the captures showing up

annavik:
I added some updates to the connections status test:
- Fixed regex typo
- Check the prefix status
- Tweaked the layout a bit
- Added some more details (see screenshots below)

**Screenshots:**
Everything ok:
Skärmavbild 2024-07-10 kl  18 01 03

Data source ok but sub dir strange:
Skärmavbild 2024-07-10 kl  18 01 19

Data source ok but regex strange:
Skärmavbild 2024-07-10 kl  18 02 21

annavik:
Some questions (sorry in advance for the long comment!):
- I noticed my sub directory `anna` got `prefix_exists: false`, I don't understand why? Same for directory `test`. I could still sync the captures though! The `vermont/RawImages/LUNA` one works fine!
- Should we disable syncing of captures if the connection has any problems?
- Is the sub dir required for deployment data source, or is it ok to leave empty?
- Any chance we can make the job details more user friendly for the sync job? Perhaps you are on it! :) That last stage(s) looks weird and also the labels. Also the list view job status and the job detail status is not always adding up. 
- When syncing captures, they show up for the deployment, but I noticed it takes a while for details about width and height to appear. The result will be a collapsed playback view. Why is that and is it and edge case I should to handle in UI? I could handle it, but since captures might take a while to load, it's helpful to know the size beforehand to jumpy interface. 

Here is my subdir with some sample captures:
Skärmavbild 2024-07-10 kl  19 14 44

I could still successfully sync the 4 captures to the deployment though:
Skärmavbild 2024-07-10 kl  19 10 32

Collapsed playback view when captures do not have defined size:
Skärmavbild 2024-07-10 kl  19 28 41

Job status is 67% done in detail view:
Skärmavbild 2024-07-10 kl  19 33 43

Job status for same job is done in list view:
Skärmavbild 2024-07-10 kl  19 33 51

annavik:
FE code wise I'm still using some hard coded strings, I feel we are still iterating a bit on the UI here. Do you think we are showing too much details about the connection status now (making the view a bit cluttered), or is ok?

mihow:
@annavik Thank you for the comments and observations. I agree the UI is looking a but cluttered, but I think it is necessary while we work out the kinks. I want to provide all the info we need to troubleshoot. Especially because syncing data can take a long time, and you may not notice an error right away. So it's great that we are making it possible to check the status of everything ahead of time that may break the sync.

1) I agree we should disable the Sync button if the connection status is not okay, great call.
2) Let's show the full URL of the first_image_found, maybe on tooltip if it's too long. Then you can see if the filter is grabbing an image from an incorrect sub folder. 
3) Height & width calculation - Let's make a ticket for that. We need to decide when to calculate. I am trying to avoid reading the source images until necessary. It may actually be a deployment-wide setting. 
4) I believe I fixed the issues with your anna & test directories. There were files in those subdirs with blank names, and the connection test was stopping on those. I increased the limit from 1 file to many even when there is no regex filter, and that seems to have solved the issue.
5) I fixed the num files checked so it shows the actual number of files checked before finding a match.
6) The sub-directory isn't technically required, but it probably should be. It's risky if you leave it blank you could end up syncing everything in the bucket. This is okay if you are in control of your storage and you really create a separate bucket for every deployment's images. But typically we are making a subdir for each deployment. 
7) Job details are funky yes, that's the last to-do I have for this PR! At least a slight improvement for now.
8) I added additional error handling to the backend and realized I can send the error messages for subdir not found & prefix not found the same way we send other error messages. So I removed the message for that from the frontend. Here are some shots of some of the errors you may encounter!

Screenshot from 2024-07-10 19-29-54

Screenshot from 2024-07-10 19-29-35

Screenshot from 2024-07-10 19-31-22

Screenshot from 2024-07-10 19-32-30

annavik:
Ok some more FE updates!
- Update connection status UI using an accordion layout (to make it possible to show connection details, but still keep the UI clean in the default case)
- Disable "Sync now" if connection was not successful
- Add tooltip to ongoing job icon button (after a sync job was started)
- Make sub directory required for the connection in the deployment case

**Screenshots:**

Collapsed accordion:
Skärmavbild 2024-07-12 kl  11 38 23

Expanded accordion:
Skärmavbild 2024-07-12 kl  11 38 31

Connection status also visible on icon hover:
Skärmavbild 2024-07-12 kl  11 41 15

First file URL visible on hover, click will open the capture in a new tab.
Skärmavbild 2024-07-12 kl  11 38 46

Same connection status component in the storage case, but with less details (and no requirement about sub directory): 
Skärmavbild 2024-07-12 kl  11 39 19

mihow:
@annavik @joseepoirier 

I just fixed the last large todo for this mega PR (related to data sync job status and interfering with the ML job status). I will return to the smaller remainders asap. So close to merging this one!

annavik:
Great job @mihow ! 🙏

PR #396, opened 2024-06-05
Fix failing pre commit hook

annavik:


mihow:
@annavik I tried a few things. But now it looks like there is a different package that is more popular for frontend/React projects to run linting on pre-push, called [Husky](https://typicode.github.io/husky/)

Issue #385, opened 2024-05-02
Sign up button should not be visible when logged in


albags:
### Description of issue
You can see the "Sign up" button after you logged in.
 

```
Path:      /projects/44
Browser:   Chrome 124.0.0.0 on Windows 10
Viewport:  1920 x 919 @1x
Language:  en-GB
Cookies:   Enabled
```

[Open in BrowserStack](https://www.browserstack.com/user/try-live?url=https%3A%2F%2Fdeploy-preview-379--ami-web.netlify.app%2Fprojects%2F44&os=Windows&os_version=10&browser=Chrome&browser_version=124.0&resolution=1920x919&speed=1&start=true&ref=netlify-source)

Submitted from [Netlify Deploy Preview #379](https://deploy-preview-379--ami-web.netlify.app/projects/44)

mVedr:
Hi, I want to contribute. Are there any guidelines for contribution.

annavik:
> Hi, I want to contribute. Are there any guidelines for contribution.

Hi @mVedr! Happy to hear you want to contribute! 🎉 We are still working on guidelines for external contributions, so probably a bit early... We will keep you updated!

PR #401, opened 2024-06-06
Fix small UI bugs after feedback

annavik:
Fixing some smaller problems that showed up during feedback sessions:
- When editing user info, the email field is called “Name” as well as the name field
- Make name a required field?
- Highlight required fields with *
- Show "10+" in the slideshow count if there are more than 10, to help troubleshoot mismatched counts
- Show seconds in the timeline component
- Show timestamps all the time in timeline component (not just on hover)
- Show that saving was successful in user info form
- Do not show "Sign up" if user is logged in (captured in #385)
- Fix issue with "not available" label showing up as a number in session details

### DIFF
--- a/ui/src/components/form/form-field.tsx
+++ b/ui/src/components/form/form-field.tsx
@@ -30,7 +30,7 @@ export const FormField = <
         <Input
           {...field}
           type={type}
-          label={fieldConfig.label}
+          label={fieldConfig.rules?.required ? `${fieldConfig.label} *` : fieldConfig.label}
           error={fieldState.error?.message}
           description={fieldConfig.description}
           onBlur={(e) => {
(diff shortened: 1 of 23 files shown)


mihow:
This is awesome, thank you for finding all of these comments. I have heard them from multiple people. One thing I noticed right away is that the localized AM/PM time breaks to a new line in the timeline view, even with a wide window size. But there is plenty of room on the left side. 

image

annavik:
> This is awesome, thank you for finding all of these comments. I have heard them from multiple people. One thing I noticed right away is that the localized AM/PM time breaks to a new line in the timeline view, even with a wide window size. But there is plenty of room on the left side.
> 
> image

Oh good catch, I missed it because of my locale. Fixed!

Skärmavbild 2024-06-10 kl  14 05 07

mihow (approved):
I love the 10+, and thanks for the other details!

PR #408, opened 2024-06-10
Fix: Single bar looks very wide

### LINKED ISSUE #405: Single bar looks very wide
In diagrams where we only have one bar, it looks very wide. I suggest we give bars a max width!

Skärmavbild 2024-06-10 kl  10 45 10

annavik:
Fixes #405

**Before:**
Skärmavbild 2024-06-10 kl  14 48 53

**After:**
Skärmavbild 2024-06-10 kl  14 48 32


mihow (approved):
Nice thanks, that was funny bar chart!

PR #409, opened 2024-06-10
Fix: Support sort params in URL

### LINKED ISSUE #371: Support sort parameters in URL
When clicking on the column headers in a list view, the sort parameter should be added as a param to the URL.

The filter params are already added to the URL when viewing a filtered list view (occurrences for one species)
https://app.preview.insectai.org/projects/24/occurrences?deployment=74

The sort params are already sent as URL params to the API 
https://app.preview.insectai.org/api/v2/occurrences/?project=24&ordering=-updated_at&limit=20&offset=0

While we are at it, the List vs. Gallery tab choices should stick in the URL as well!
https://app.preview.insectai.org/projects/18/occurrences?determination=24
image

annavik:
Fixes #371 

Also making table/gallery view setting part of URL. One question remains in parent issue.


mihow (approved):
Works! Awesome thank you!

PR #417, opened 2024-06-11
Tweak UI for identifications

annavik:
Now when identifications may include both algorithm details and comment I thought could be helpful to tweak the card layout slightly, to distinguish the taxon from other details. Also I disabled the button "✓ Agreed" to avoid confusions. If already agreed, pressing it would not have any effect.

Let me know what you think, I just did these tweaks quickly after testing submitting a comment! :)

**Before:**

Skärmavbild 2024-06-11 kl  11 42 46

**After:**

Skärmavbild 2024-06-11 kl  11 42 28

### DIFF
--- a/ui/src/data-services/models/algorithm.ts
+++ b/ui/src/data-services/models/algorithm.ts
@@ -15,7 +15,7 @@ export class Algorithm {
     })
   }
 
-  get description(): string {
+  get description(): string | undefined {
     return this._algorithm.description
   }
 
--- a/ui/src/design-system/components/identification/identification-summary/identification-summary.module.scss
+++ b/ui/src/design-system/components/identification/identification-summary/identification-summary.module.scss
@@ -1,12 +1,19 @@
 @import 'src/design-system/variables/colors.scss';
 @import 'src/design-system/variables/typography.scss';
 
+.wrapper {
+  display: flex;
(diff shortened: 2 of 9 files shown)


mihow:
This is great thank you @annavik! I stuffed in the ML model name in the easiest place I could. I like the Agree button change.

I did notice an empty tooltip shows if the model description is empty
image

image

At some point we will show more information about each prediction:
- A link to the job that was used to process the image (so you can review errors or see the collection it was apart of, and what other processing steps were run).
- A link to the model page 
- The accuracy of this species in this model
- The number of images the model was trained on for this species

mihow (approved):

annavik:
> This is great thank you @annavik! I stuffed in the ML model name in the easiest place I could. I like the Agree button change.
> 
> I did notice an empty tooltip shows if the model description is empty image
> 
> image
> 
> At some point we will show more information about each prediction:
> 
> * A link to the job that was used to process the image (so you can review errors or see the collection it was apart of, and what other processing steps were run).
> * A link to the model page
> * The accuracy of this species in this model
> * The number of images the model was trained on for this species

Oops, tooltip thing fixed!

Good to know! I guess we could introduce some expand feature on the card, to fit more details... Or some other solution!

PR #429, opened 2024-06-14
Some improvements to table columns (occurrence and species)

### LINKED ISSUE #427: Show quick actions even though we have verification by human
Comment from Lars:
"For the case where human A has provided the not Lepidoptera ID, but human B happens to know more than that. Would be good if human B doesn't have to go through the openingen image pop-up etc."

annavik:
Some tweaks to species and occurrence table after feedback from Lars! Also including some related tweaks. See the changes in this PR as a suggestion! :)

### Summary of changes
- Tweak column order and labels for occurrences and species.
- Move identification status to score column.
- Never hide identification quick actions in occurrence table view (fixes #427).
- Show detailed taxon info in species table.

### Notes
In order to show the correct "agreed" status in table view, user info details has also been made globally available from a context as part of this PR. This makes it possible for us to easily access user info from any component in app, without having to refetch it. This behaviour can in general also be achieved by react query cache, but for user info I think it makes sense to put it in a context. This hook has some special auth behaviour I don't want to mess with.

### Screenshots
**Before:**

Skärmavbild 2024-06-14 kl  11 47 26
Skärmavbild 2024-06-14 kl  11 47 15

**After:**

Skärmavbild 2024-06-14 kl  11 46 29
Skärmavbild 2024-06-14 kl  11 46 46


annavik:
Let me know what you want to do this with @mihow. I'm fine with reverting moving verification status to score column if you feel we should do some more thinking here and we can just keep the 3 other tweaks :)

mihow (approved):
I like it let's try it!

mihow (approved):

mihow:
After merging I noticed that we could add the same score component to the Species table as well!

image

PR #432, opened 2024-06-17
Fix: Make reject options easy to access

### LINKED ISSUE #431: Make reject options easy to access
Reject ID options can for example be "Not Lepidoptera" and "Not identifiable". At the moment the user needs to search for them using the taxon search, which is tricky if you don't know these options exist beforehand.

We probably want a more fancy UI solution here, but to start with the goal is to just present these options in a separate list somehow. Both in table view (probably the most common place to reject id) and also in identification view.

annavik:
Fixes #431 

We solve this temporarily with a popover and hardcoded values for reject options.

**Screenshots:**

Skärmavbild 2024-06-17 kl  15 44 20

Skärmavbild 2024-06-17 kl  15 44 56

Skärmavbild 2024-06-17 kl  15 46 07


annavik:
What do you think about the icon for this? We could also use a more neutral "options" icon and put more stuff in the popover, not just reject actions.

Skärmavbild 2024-06-17 kl  15 37 37

ui/src/pages/occurrence-details/reject-id/contants.ts
```diff
@@ -0,0 +1,4 @@
+export const REJECT_OPTIONS = [
```
annavik, line 1: Hardcoding this for now @mihow, happy to fetch from backend if we have an endpoint! :) Let me know if we have more reject options, these are the ones I could find atm.

mihow:
I made a reminder ticket to revisit this later here #438

mihow (approved):
I think this is an awesome quick fix. We can see if people like this approach. I think the Reject Options will be configurable per project and eventually come from the API. I like the icon too.

PR #444, opened 2024-06-25
Update page layout to support new features

annavik:
### Summary of changes:
- Setup sticky page header and fixed page footer
- Add loading and fetching state to page header
- Setup new toggle group component and move view settings to page header
- Make table headers sticky
- Use the new layout components on all main pages 

### Some comments:
- Making the top bar being "pushed away" without messing too much with the current element structure was a bit tricky. I ended up rendering the menu component using a portal here. In this way we can put it next to the top menu in the DOM tree, but still render it from a page component.
- Combining sticky table header with horizontal table scrolling turned out to be difficult due to some CSS limitations. I tested a few different approaches to this using JS. Ended up with a table clone + a small library for sync without flickering. See the help component `StickyHeaderTable`. Core issue is described in this W3C thread: https://github.com/w3c/csswg-drafts/issues/865


mihow (approved):
Great improvements. Keyboard accessibility and narrow viewports (mobile) seem to still work as well. Fantastic

PR #448, opened 2024-07-02
Make applied filters more visible in UI

annavik:
### Summary of changes
- Show applied filters directly in page header in same styles as bubble links
- Skip filter settings popover for now

### Future plans for filters
- Return filters in a more readable format from backend (so that we can show a label instead of ID in UI)
- Add custom filter controls

### Screenshots
Before:
Skärmavbild 2024-07-02 kl  10 44 33

After:
Skärmavbild 2024-07-02 kl  10 43 43


mihow (approved):
So cool! I added a ticket for me to add more details about the active filters in the API response #452

Issue #447, opened 2024-07-01
Allow agreeing to higher taxon ranks


mihow:
The quick options for rejecting an ID were added in #432 & #441. We also need some quick actions for Accepting / Applying a suggested ID:

- Apply species
- Apply genus 
- Apply tribe
- Apply subfamily
- Apply family
- Apply order

How close is the new design for this? Perhaps, for now, the Accept button becomes a drop-down like the Reject button is. And we remove the green background. The choices can come from the suggestion (all taxon ranks available for the suggested taxon). 

These should work in the upcoming bulk actions bar as well.

PR #450, opened 2024-07-02
Agree to rank

annavik:
Here is a suggestion to a solution for #447:
- Group both apply and reject options in a "more" popover
- Keep agree option as is
- Make suggest ID button compact (with tooltip) in table view to avoid visual overload


annavik:
**Some questions:**
- What do you think of the overall approach, does it work for a first version?
- For the ranks, I'm currently using the ones I'm getting with the occurrence list response (`determination_details.taxon`). Can we get more levels, or is this causing too much load on DB?
- Do you think we should include the name of what will be applied somehow, not just the rank?

mihow:
Thanks so much for finding a way to support this feature. I think it's working great, except I do think we should say which name will be applied as you suggested. 

I like that you use the current taxon ranks that are returned. I will add the additional levels in an upcoming PR, but there are a few things that need to happen in the DB first (which will unblock a couple of other tickets as well)

annavik:
> Thanks so much for finding a way to support this feature. I think it's working great, except I do think we should say which name will be applied as you suggested.
> 
> I like that you use the current taxon ranks that are returned. I will add the additional levels in an upcoming PR, but there are a few things that need to happen in the DB first (which will unblock a couple of other tickets as well)

Great! I have added some more details to the apply options. Do you prefer tooltip or inline? I pushed the inline variant for now. It's similar to taxon search and also I thought it was a bit more clear, although not as compact.

Skärmavbild 2024-07-03 kl  11 09 17

Skärmavbild 2024-07-03 kl  11 08 50

Also tweaked button styles to better match Andres suggestion for future.

mihow (approved):
I really like how the current inline version looks!

Issue #361, opened 2024-02-27
Show all detections in session detail view


mihow:
There is confusion about what objects has been detected or not in a frame. 

Let's update the view to:
1) Show the detections overlay without hovering
2) Show all detections, regardless of confidence threshold selected
2) Add emphasis to the detections that do meet the confidence threshold 

"Show all detections, regardless of confidence threshold selected" is specifically about the bounding boxes on the big image in the session detail view. The confidence score comes from the just species ID classifier, but it the threshold is currently applied to BOTH the species ID and the original bounding box detection. Users are mistakenly thinking that the object detector is just terrible and can't find 1/2 of the moths in the image. Maybe we show all boxes with a dim/gray style on hover, but then only the moths that pass the selected threshold get the current style treatment and are visible before you hover.

mihow:
@annavik I added that new detail to the description here. Also I like the idea of the backend returning all detections always, and then the filtering happening in real time as you move the slider. You can test that now by always passing a threshold of 1.0 to the backend endpoint

mihow:
Closed by #453!

Issue #263, opened 2023-10-03
Need faster way to browse through captured images


mihow:
It's a bit tedious to scroll through images of a session before they have been processed. 

I'd like to try an interactive chart of the images captured per hour, on which you can click and jump to images at different points in the night. Likely just a horizontal bar chart on the session detail page.

mihow:
@annavik I wonder if we can come up with an interim solution for this one. I believe I added support in the backend endpoint to return a page of results based on a time in the night (as a query param). Currently the session detail view is not usable unless you navigate to an interesting image by clicking on an Occurrence first, or from the Collection list view. If you just start from the beginning of the night, you will never get to the moths!

mihow:
Related to #198 & #273

mihow:
Closed by #453!

PR #453, opened 2024-07-04
Refactor session details

annavik:
We are moving towards a new, more scaleable, way of navigating captures. This is the start to adress both #263 and #361. 

For now, the timestamp slider is not possible to interact with (we need some BE changes to select capture based on timestamp first). It shows the idea though. Next and previous button works, but we are only considering the first 200 captures for now. I will update the logic there as well after the BE updates.

For capture detections, they are now all be visible by default (but can be hidden from sidebar). I have also update the frames to better match future design from Andre. Selected captures will always be visible and will get a primary border and the tooltip.

### Notes:
Maybe for later, just putting it here:
- I think we can improve the sidebar content a bit. Perhaps show capture info inline, instead of in popover? Also the processing settings feel a bit strange to me, for example the expand button is a bit unclear? _Update: this is fixed now!_ 
- More view settings? Make it possible to toggle the "score frames"? Make it possible to only show detections that matches the score, like before?
- Would be nice to store all view settings in browser, not just the threshold.
- Maybe the sidebar should be possible to show/hide if you want a nice full view of the capture?
- Showing all verification labels at once is a bit tricky for now, we need some collision handling.

### Screenshots:
Skärmavbild 2024-07-04 kl  15 41 34

Skärmavbild 2024-07-04 kl  15 41 58


mihow:
Really exciting. I will respond your thoughts soon and provide an endpoint for the timeline feature. I'm thinking a dedicated, lightweight array of tick marks & source image IDs just for generating the timeline component -- on the session detail endpoint. Unless you have another idea!

annavik:
> Really exciting. I will respond your thoughts soon and provide an endpoint for the timeline feature. I'm thinking a dedicated, lightweight array of tick marks & source image IDs just for generating the timeline component -- on the session detail endpoint. Unless you have another idea!

That sounds really great! For example, a list of ticks like this would be helpful! Like splitting the session into "sub sessions"? This would make it possible for me to take any timestamp (either specified in URL or on timeline) and move the user to that point in time. Also select a capture, if present. The field `detections_count` we can leave for later, but if present it would make it possible for me to include activity info more similar to the design. Also the processing status for the sub session would be helpful later on!

```
  ticks: {
    start: string
    end: string
    first_capture?: string
    detections_count: number
  }[]
```

Also, if possible, including next and previous capture with the capture details would be helpful 🙏 I know I can probably get that from captures page endpoint, but after some thought I think it would be nicer to only call capture details. We could avoid one extra network call and keep the code a bit more simple.

mihow:
I see your suggested data structure for the ticks, I will try that asap! I can add a `last_processed` timestamp, which for the sub session would have to be some sort of judgement based on all of the captures in that tick. Perhaps if 1 out of 3 images have been process, then the status is not processed, but if 2 out of 3, then return the latest date processed. So >= 50% of images in tick means show the "has been processed" color. We might have to see how it feels. It could also just be the status of the first capture in that tick.

```
  ticks: {
    start: string
    end: string
    first_capture?: string // object of capture or something else?
    detections_count: number
    last_processed: null | datetime
    
    // other ideas for processing status:
    processing_status: enum
    in_queue:  bool
    jobs: Job[] 
  }[]
```

mihow:
> Maybe for later, just putting it here:
> 
> I think we can improve the sidebar content a bit. Perhaps show capture info inline, instead of in popover? Also the processing settings feel a bit strange to me, for example the expand button is a bit unclear?
> More view settings? Make it possible to toggle the "score frames"? Make it possible to only show detections that matches the score, like before?
> Would be nice to store all view settings in browser, not just the threshold.
> Maybe the sidebar should be possible to show/hide if you want a nice full view of the capture?
> Showing all verification labels at once is a bit tricky for now, we need some collision handling.

- I think showing capture info inline instead of popover is the most helpful out of all of these. Although I like all of these!
- I'm curious what collision / "label auto layout" libraries exist for React. I know web maps have to deal with this issue. 
- Storing all settings in the browser session seems important too
- The score slider is there now, what do you mean by toggling the score frames?

annavik:
> * I'm curious what collision / "label auto layout" libraries exist for React. I know web maps have to deal with this issue.
Yes, I hope we can use some existing solution!

> * The score slider is there now, what do you mean by toggling the score frames?
I think I meant show detections (and possible hide ones below threshold), but skip the coloring based on score.

mihow:
Hi @annavik I realized the data structure won't quite work. Because I think we want the tick groups / sub-sessions to be determined by the frontend, based on the width of the chart and duration of the night, right? So perhaps the data should be a list of all timestamps for the night, with number of detections & image ID for each. Then you can aggregate them together on the frontend. The plotly chart library does this aggregation automatically, and even provides a zoom & drag tool for timeline-like charts. Rechart probably does too. I do wonder if a bar chart is more appropriate (with one tick per minute or groups of minutes, depending on zoom level). 

But! I am fine to keep this super simple. Important: The main point of this component is to be able to see the whole night from start to finish, see where the images were captured, and see where moths were detected, and be able to click anywhere in the night to be able to jump there. 

```
const data = [
  { datetime: "2024-07-31T23:00:00", detections: 0, image_id: 123, processed: true },
  { datetime: "2024-07-31T23:10:00", detections: 1, image_id: 124, processed: true },
  { datetime: "2024-07-31T23:20:00", detections: 0, image_id: 125, processed: false },
  { datetime: "2024-07-31T23:30:00", detections: 1, image_id: 124, processed: true },
  { datetime: "2024-07-31T23:40:00", detections: 1, image_id: 126, processed: true },
  { datetime: "2024-07-31T23:50:00", detections: 0, image_id: 127, processed: false },
  { datetime: "2024-08-01T00:00:00", detections: 3, image_id: 128, processed: true },
  { datetime: "2024-08-01T00:10:00", detections: 2, image_id: 129, processed: true }
];
```

https://claude.site/artifacts/00823cc2-8f67-42b0-9a5c-8408c2045f14
image
image

annavik:
Good point about the data structure! I have made some updates here today using Plotly, since we are already using this lib for graphs. I now assume we will get some basic info for each capture. I'm not doing any aggregation of data yet, but I think the current chart is giving a pretty good overview of detections. We can now also use the slider to move to a point in time and we will snap to the most nearby capture. You can also set active capture by clicking on a point in the graph.

Skärmavbild 2024-08-01 kl  16 45 54

I suggest we leave zooming, processing status, etc. for later. I want to keep it simple and merge this first step in the update first.

**Needed from BE:**
Just the overview data for the full night (right now I use data from 1 page of captures with page size 1000, but I don't need all these details). We can use the format you suggested!

mihow:
That looks great! I will work on a new endpoint for fetching the timeline data using the data structure similar to the one above.

mihow:
Okay we have an endpoint! 

`/events/1/timeline/?resolution_minutes=1`

I went ahead and made an interval system after all! haha. Since our images can be taken every second I decided to make intervals of a minimum of 1 minute. That means we have to make each tick be a summary of images & detections after all. So the data looks almost identical to your original proposal. I hope this works!

The backend for this branch is now deployed on staging:
https://api.beluga.insectai.org/api/v2/events/1/timeline/?resolution_minutes=1

You can see there is an optional parameter for resolution_minutes, which defaults to 1. The other common values might be 1, 10, 30, and 60. But I don't know if its actually helpful to make a toggle.



```json
         {
            "start": "2022-06-21T03:02:00",
            "end": "2022-06-21T04:02:00",
            "first_capture": {
                "id": 1886338,
                "details": "http://localhost:8000/api/v2/captures/1886338/",
                "user_permissions": []
            },
            "captures_count": 10,
            "detections_count": 24
        },
        {
            "start": "2022-06-21T04:02:00",
            "end": "2022-06-21T05:02:00",
            "first_capture": {
                "id": 8,
                "details": "http://localhost:8000/api/v2/captures/8/",
                "user_permissions": []
            },
            "captures_count": 12,
            "detections_count": 25
        },
        {
            "start": "2022-06-21T05:02:00",
            "end": "2022-06-21T05:23:00",
            "first_capture": {
                "id": 2,
                "details": "http://localhost:8000/api/v2/captures/2/",
                "user_permissions": []
            },
            "captures_count": 2,
            "detections_count": 5
        }
  ```
  
  I also added a meta attribute 
```
   "meta": {
        "total_intervals": 10,
        "resolution_minutes": 60,
        "max_detections": 30,
        "min_detections": 0,
        "timeline_start": "2022-06-20T20:02:00",
        "timeline_end": "2022-06-21T05:23:00"
    }
```

Probably nothing super helpful now, but I think it's a good pattern that we should use for most endpoints (e.g. for the active filter & sort information!)

**Comment by annavik** [comment]
Yay, thank you so much, it works great! I have updated the PR to use this endpoint instead.

Some things I noticed:
- Since we are not using info for each capture, could the capture details endpoint return id for previous and next capture? Also current capture index would be helpful. I need it for the control below.
- The last capture seem to be excluded from the ticks, I noticed it for small test sessions. For example the session #3602 has 4 captures, but the last one (capture #9427942) is excluded from timeline. I would expect it to be included in the last tick. The effect will be it looks as if the session ends earlier, see screenshot below. Example: https://api.beluga.insectai.org/api/v2/events/3602/timeline/. I guess just some interval logic tweak?

Skärmavbild 2024-08-02 kl  12 10 59

Skärmavbild 2024-08-02 kl  12 02 34

**Comment by annavik** [comment]
Do you think it makes sense to filter out ticks with no captures FE side for a more consistent less spiky graph? This is the same session, but first one is without ticks that has no captures.

Skärmavbild 2024-08-02 kl  11 32 04
Skärmavbild 2024-08-02 kl  11 32 49

**Comment by annavik** [comment]
What do you think about line shape "spline"? With vs. without:

Skärmavbild 2024-08-02 kl  12 27 32
Skärmavbild 2024-08-02 kl  12 28 04

**Comment by annavik** [comment]
Also if you could have a quick look at this I think could improve the user experience a bit 🙏  https://github.com/RolnickLab/ami-platform/issues/473

**Comment by annavik** [comment]
Another thought: do we need to support selection of capture by occurrence ID in the URL? We did this before. When we fetched list of captures we could search capture detections for the selected occurrence. Since we only fetch the timeline now, this is currently possible without BE tweaks, but also perhaps not necessary? From occurrence details view, we include capture ID in URL, which means the correct capture will still be selected for this type of navigation.

**Comment by mihow** [comment]
I went for it and added a new line for the number of captures per tick! I see this as important because it shows where the camera fired in the night (and where it didn't). Also this gives the chart some data to render before the images are processed. 

The chart looks way different depending on whether it is capturing based on time intervals or motion detection. But I think that is great, it makes it so the data sampling settings are super obvious! 

Motion detection camera:
image

Timelapse camera:
image

- Can we add vertical markers every hour or more time labels along the X axis?
- Can the height of the chart be shorter? 

As-is this is very useful and allows me to navigate the session event, which is just what we needed!

**Comment by mihow** [comment]
Okay some more fixes are in!

- The last capture in the timeline is available (not missing)
- Num detections is available for the event / session detail view
- The capture detail includes the next/prev capture in the event, and the captures index within its event

```
    "event_next_capture_id": 1436801,
    "event_prev_capture_id": 1436799,
    "event_current_capture_index": 3,
    "event_total_captures": 29,
```
https://api.beluga.insectai.org/api/v2/captures/1436800/

annavik:
Woho, thank you so much for the fixes!

Pushed some updates:
- Updated capture stepper to be functional and tweaked the stepper layout to better match new design.
- Tweaked the plot layout:
  - Changed color from green -> grey for captures. Since we are using the green color for the score in this context, I'm thinking might be nice to use another one for the captures. Also could be nice to put the visual focus on detections?
  - I skipped the markers for the capture graph, I think the line only looks a bit cleaner, what do you think?
  - Tweaked the plot height from 200 -> 100.
  
After changes:

Skärmavbild 2024-08-04 kl  17 48 20

annavik:
Created some issues for your remaining suggestions @mihow:
- #482 
- #483 

I will try fix these really soon because I think they are good improvements, but perhaps they don't need to block this PR though? Marking this one for review!

PR #454, opened 2024-07-08
Add tooltips to page header

### LINKED ISSUE #399: Clarify entity types and terminology in UI
After processing some of the user feedback, it seems to be some confusions around the entity types and also the terminology in some cases. For example, what defines a "deployment"? What is a "site"? What does a "collection" include?

To prevent this, I suggest we include some clarifications in the UI here. To not disturb seasoned users with information, I suggest we show this information on demand, for example using tooltips. We could show a short description and also include a link to more detailed information, when we have a wiki?

annavik:
Fixes #399 

Example:
Skärmavbild 2024-07-08 kl  15 48 59

### DIFF
--- a/ui/src/design-system/components/page-header/page-header.tsx
+++ b/ui/src/design-system/components/page-header/page-header.tsx
@@ -1,12 +1,16 @@
 import { AppliedFilters } from 'components/applied-filters/applied-filters'
 import { ReactNode } from 'react'
 import { STRING, translate } from 'utils/language'
+import { IconButton, IconButtonTheme } from '../icon-button/icon-button'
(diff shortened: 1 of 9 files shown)


mihow:
Tooltips, yes!! Are the strings copied from the google doc? I want to update a few of them but I should change them in both places and make sure there isn't existing discussion about a string.

annavik:
> Tooltips, yes!! Are the strings copied from the google doc? I want to update a few of them but I should change them in both places and make sure there isn't existing discussion about a string.

Yes I have copied them! Feel free to update the Google doc and I can sync to this PR if you want :)

PR #459, opened 2024-07-12
Handle missing capture size

annavik:
As we are scaling up, we will not always get the capture size in the backend response. To avoid a collapsed playback view when this happens, we will start inspect the capture image asset after it has been fully loaded.

Following logic is applied in this PR:
- If the image **has** been loaded, we will render the frame (including detection boxes) using the natural size of the image.
- If the image **has not** been loaded, we check for size info in the backend response. If this info is not present, we will wait for the image to load before rendering any detection boxes (without any size info, the boxes will not make any sense). However we will still render a frame using a 16/9 fallback ratio, to avoid a collapsed view.

### DIFF
--- a/ui/src/data-services/models/capture.ts
+++ b/ui/src/data-services/models/capture.ts
@@ -76,7 +76,7 @@ export class Capture {
     return this._detections
   }
 
-  get height(): number {
+  get height(): number | null {
     return this._capture.height
   }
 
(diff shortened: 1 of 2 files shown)


annavik:
Related to #379

mihow (commented):
Thank you for jumping on this!

PR #460, opened 2024-07-12
Bulk actions

annavik:
In this PR we setup bulk actions for occurrences. At the moment it's possible to agree in bulk and also to apply ID in bulk. The apply options are currently a list of all the common ranks among the selected occurrences. Also we will show the recently applied identification as an apply option.

Comments for later:
- Sticky header has a bug. When in sticky mode, click actions are ignored. Annoying when wanting to "select all". Created the issue #489.
- Make it impossible to apply same id, if the ID was already applied by the current user? Perhaps not very important...?
- Backend plans to return more ranks, this will make the list of overlapping ones more complete and also solve some other issues. Related to https://github.com/RolnickLab/ami-platform/issues/411

### Screenshots:
Skärmavbild 2024-07-12 kl  18 22 17
Skärmavbild 2024-07-12 kl  18 22 28


mihow:
TODO for @mihow : At least add Family level taxa choices

mihow:
This one is working great. However one change I purpose is that if you select multiple Occurrences and then use the Apply ID quick action, it should only show the taxa choices that are _common_ among all the choices. Right now it shows all of the taxa from all of the choices. I'm hoping there is a reduce function of some sort that can make that easy!

image

And will you go ahead and try to implement the Suggest ID search for the bulk select? 

OR!

As an alternative, we could try a new feature:

Under Apply ID, add a "Recent" section of taxa that were recently applied (maybe saved in the current browser session). This would be pretty handy in multiple cases. But also you could use it as an alternative to search. Because you can search & apply a custom ID to a single Occurrence, then it will show up in the recent column.

Thoughts?

ui/src/data-services/hooks/identifications/useCreateIdentifications.ts
```diff
@@ -0,0 +1,20 @@
...
+export const useCreateIdentifications = (
+  params: IdentificationFieldValues[]
+) => {
+  const { createIdentification, isLoading, isSuccess } =
+    useCreateIdentification()
+
+  return {
+    isLoading,
+    isSuccess,
+    createIdentifications: async () => {
```
mihow, line 13: This is great! The async really pays off here. I see all 16 requests fire off at once and the backend seems to respond quickly.

mihow:
One experience detail: if you are sorting by score, as soon as you apply a batch of IDs, the IDs disappear because they now have a score of 1. Perhaps this is a feature! because then you are presented with the next batch, but at first it was confusing. But here are some ideas:

- Show the results of the identification calls in a modal before the table refreshes (similar to an error modal, but it shows the successes as well). 
- Add an "identified_at" column so it's easy to see what was recently identified 
- Don't change the score field to 1 when a user verifies a prediction. Perhaps this is confusing for other reasons. We can change the icon, but leave the original machine score? (which would be the most recent machine score). But also I was thinking that this would become a variable score even for humans (the identifier can choose their own score based on confidence).
- It's not an issue! Once you learn the behavior

annavik:
> This one is working great. However one change I purpose is that if you select multiple Occurrences and then use the Apply ID quick action, it should only show the taxa choices that are _common_ among all the choices. Right now it shows all of the taxa from all of the choices. I'm hoping there is a reduce function of some sort that can make that easy!
> 
> image
> 
> And will you go ahead and try to implement the Suggest ID search for the bulk select?
> 
> OR!
> 
> As an alternative, we could try a new feature:
> 
> Under Apply ID, add a "Recent" section of taxa that were recently applied (maybe saved in the current browser session). This would be pretty handy in multiple cases. But also you could use it as an alternative to search. Because you can search & apply a custom ID to a single Occurrence, then it will show up in the recent column.
> 
> Thoughts?

Ah yes, I should definitely only show the common choices! Also once BE is returning more levels, we can do this without risking excluding some choices that are actually common. But anyhow, I will go ahead and fix this!

For suggest ID, I think yes, but could probably need a bit of thinking through UX wise and I don't see it as a blocker for this feature. I think the thing you pointed out above, error handling and the sticky header thing I mentioned in PR descriptions are the most important things to fix, so will do that first! :)

I like your suggestion with showing recent options, seems handy! Probably pretty straight forward design wise as well. I'll see if I include here, or in a separate branch...

annavik:
### Pushed some updates:
- Updated logic for apply options (only show common ranks for the selected occurrences)
- Error handling (show how many requests were rejected and make it possible to retry those)
- Present recent identifications as apply options

### Screenshots:
Common ranks visible for the selected occurrences and also the recently applied taxon:
Skärmavbild 2024-08-05 kl  13 18 23

Case when we have no common ranks for selected occurrences:
Skärmavbild 2024-08-05 kl  13 18 32

Error example:
Skärmavbild 2024-08-05 kl  13 20 28

More info on button hover. On button click, we will retry only the failed requests:
Skärmavbild 2024-08-05 kl  13 20 34

annavik:
I have updated the PR description with current status of this task. I think this one could get merged as is and we can handle what is left as separate issues. Hopefully it's still useful even though we of course also want the taxon updates in order to show better choices :)

annavik:
Reminders for Anna:
- Do not close popup on selection change
- Filter selections based on what is currently visible (for example if sort order is changed and some item is no longer visible, clear that selection)

mihow:
I love using the new persistent modal while adding to the selection! Merging!

mihow (approved):

PR #478, opened 2024-08-02
Disable populate for starred collection

annavik:
If I'm not misunderstanding, I think this should not be possible and could be a bit confusing to users? Removing! 

Skärmavbild 2024-08-02 kl  17 00 51

### DIFF
--- a/ui/src/data-services/models/collection.ts
+++ b/ui/src/data-services/models/collection.ts
@@ -7,6 +7,10 @@ export class Collection extends Entity {
     super(entity)
   }
 
+  get canPopulate(): boolean {
+    return this.canUpdate && this._data.method !== 'starred'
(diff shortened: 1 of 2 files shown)


mihow (approved):

mihow:
@annavik thanks for catching! If we make the sampling methods shared types, we could add a `populationType`. These are:
- Manually curated (starred collections)
- Statically updated (important if you want to refer back to a certain state of a collection)
- Dynamically updated (stay up-to-date based on rules: all images in a project, all 10 minute intervals, etc). 
- 
Okay I went ahead and made a ticket for this idea!  https://github.com/RolnickLab/ami-platform/issues/487

mihow (approved):

PR #479, opened 2024-08-02
Show pipeline ID

### LINKED ISSUE #475: Show pipeline ID
Pipeline ID:s are present in database, but we no not display them. Let's do it!

annavik:
Fixes #475. Also including some other small tweaks for default sort order and dialog trigger apperance.

**Before:**
Skärmavbild 2024-08-02 kl  17 33 53

**After:**
Skärmavbild 2024-08-02 kl  17 31 29

### DIFF
--- a/ui/src/pages/overview/pipelines/pipelines-columns.tsx
+++ b/ui/src/pages/overview/pipelines/pipelines-columns.tsx
@@ -7,6 +7,7 @@ import { STRING, translate } from 'utils/language'
 export const columns: (projectId: string) => TableColumn<Pipeline>[] = () => [
   {
     id: 'id',
+    sortField: 'id',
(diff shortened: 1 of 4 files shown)


mihow (changes requested):
Thank you! Will you make the column sortable too?

annavik:
> Thank you! Will you make the column sortable too?

I actually tried, but it seemed to have no effect on the fetch call. Perhaps sorting by id supported by backend? Or I used the wrong key, I tried `id` and `_id` ...

annavik:
This seems to be updated, thanks! We can now sort by ID and I also made this the default sort order ✅

mihow (approved):
Thanks for noticing that I added the option in the other PR! I intended to mention that.

PR #480, opened 2024-08-02
[Draft]: Display access key as a regular string

annavik:
Frontend is prepared now, we just need to include it from API response @mihow! :)


mihow:
Note for myself: I think the detail view for the data sources should not be public, only the list view.

PR #481, opened 2024-08-02
Fix deployment date labels

annavik:
If deployment has no captures, we show no start or end labels. Workaround for https://github.com/RolnickLab/ami-platform/issues/461 to avoid confusions in upcoming launch, but feel free to do a proper fix BE side when there is time :)


mihow (approved):
Thanks! I think the backend behavior is what is expected. If there are no images, then the first & last image dates of the deployment are `null`. When we have shared types then it will be easy to see which fields can be null!

```
            "first_date": null,
            "last_date": null,
```

Issue #485, opened 2024-08-04
Species detail links are often broken


annavik:
**Steps to reproduce:**
1. Go to https://app.preview.insectai.org/projects/18/species/5464
2. Click on the genus label
3. Detail view try load load, but then nothing renders (reason is no item with this ID was found)

We should handle this error more gracefully from the FE side, but also look into the reason. Seems to happen for a lot of records, also when it comes to links from occurrence details. I have only noticed it for the lower ranks though if that can be a clue...

PR #486, opened 2024-08-04
Present error to users if fetching entity details fails

annavik:
When fetching entity details, we were not displaying any fetch errors to users. In this PR, we will use our current error component to display such errors. Also we will check the error object for info messages we can present to users.

Background is #485, I noticed the user experience was kind of bad when this happened!

After the fix we show the error like this (message string might vary depending on the response):

Skärmavbild 2024-08-04 kl  22 36 44


mihow (approved):
Thank you!

Issue #421, opened 2024-06-11
Always show family when displaying upper taxon levels, omit genus


mihow:
Currently we show the 2 taxon levels above species. Since moths commonly have tribe or subfamily, the family is usually never visible. Also the genus is redundant since it is part of the species name. The family is one of the most helpful taxon levels.

This likely depends on #264, because there are already too many SQL joins when populating the upper taxon levels for each species in an API response.

The UI likely doesn't need to make any changes for this to work.

mihow:
@annavik this one is gonna be helpful for us moth novices!

annavik:
Yes!! 🙏

PR #491, opened 2024-08-06
Expose all Taxon parents

### LINKED ISSUE #485: Species detail links are often broken

mihow:
This adds a new `parents` attribute to the Taxon object in all related API endpoints. This new attribute is an ordered list of all ranks and is intended to be used instead of the recursive `parent` attribute. 

Most UI views have been updated to use the new `parents` list, except for the species list view! I can't crack that one.

One `parent` is still available if needed. 

This enables the bulk ID actions to be much more useful because you can select multiple occurrences and choose the taxon rank that is common among all of them (family, superfamily, lepidoptera, etc)

Also fixes #485, you can click on any rank and the species detail page will not 404.

image

image

image

```json
{
"id": 1211,
"name": "Calliteara pudibunda",
"rank": "SPECIES",
"parent": {
"id": 1210,
"name": "Calliteara",
"rank": "GENUS",
"details": "http://localhost:8000/api/v2/taxa/1210.json",
"user_permissions": [
"update",
"delete"
]
},
"parents": [
{
"id": 1,
"name": "Lepidoptera",
"rank": "ORDER"
},
{
"id": 7743,
"name": "Noctuoidea",
"rank": "SUPERFAMILY"
},
{
"id": 50,
"name": "Erebidae",
"rank": "FAMILY"
},
{
"id": 8554,
"name": "Lymantriinae",
"rank": "SUBFAMILY"
},
{
"id": 11319,
"name": "Orgyiini",
"rank": "TRIBE"
},
...
```


mihow:
@annavik Can you help update the species list view to get the ranks back? I can't track that one down.

image

ui/src/data-services/models/taxa.ts
```diff
@@ -13,26 +27,26 @@ export class Taxon {
...
-      ranks.push({
-        id: `${current.id}`,
-        name: current.name,
-        rank: current.rank,
-      })
-      current = current.parent
+    if (taxon.parents) {
+      this.ranks = taxon.parents
+    } else if (taxon.parent?.parents) {
+      // TODO: Update this when species list is returning parents similar to other endpoints
```
annavik, line 38: @mihow seems in the species list case, data structure looked slightly different. I could still access all parents though from the `parent` object. We can skip this once we streamline the format :)

ui/src/data-services/models/taxa.ts
```diff
@@ -1,10 +1,24 @@
 export type ServerTaxon = {
-  id: number
+  id: string
   name: string
   rank: string
   parent?: ServerTaxon
+  parents?: ServerTaxon[]
 }
 
+const SORTED_RANKS = [
```
annavik, line 9: @mihow I noticed the ranks were not always sorted (you can for example see this on the occurrence table screenshot you posted in PR description). I added some client side sorting for now, based on this list of ranks. I hope it's correct!

We can skip this if backend plan to sort, but leaving it for now.

annavik:
I went on and updated the rank appearance in table view, to show max 3 ranks (least detailed and most detailed):

Skärmavbild 2024-08-06 kl  10 17 19

Also tweaked the row space, for the case when we need the wrap behaviour:

Skärmavbild 2024-08-06 kl  10 23 55

annavik:
After merging this I came to think of #421 and updated the compact logic to:

Skärmavbild 2024-08-06 kl  10 58 13

See commit https://github.com/RolnickLab/ami-platform/commit/b611a2705e12f537957c1e0951f1d2a7deb6b924

PR #494, opened 2024-08-06
Tweak tick config for plots

annavik:
The `dtick` prop was previously set to `1` to avoid decimal values on the x-asis for some plots, but this turned out to have bad consequences for other plots. Skipping this prop for now, I think it must be set conditionally.

Before:
Skärmavbild 2024-08-06 kl  12 17 29

After:
Skärmavbild 2024-08-06 kl  12 21 09


mihow:
Thanks I was wondering about those jammed letters!

PR #495, opened 2024-08-06
Make sticky table header clickable

### LINKED ISSUE #489: Click actions are ignored in sticky header
Sticky header has a bug. When in sticky mode, click actions are ignored. Annoying for selections and sorting.

annavik:
Fixes #489 

### Background
- The core issue: https://github.com/w3c/csswg-drafts/issues/865
- Previous solution: https://github.com/archfz/vh-sticky-table-header
- The new problem: turned out the sticky header click handling did not work as expected with the lib above. Tracked it down to the fact this library assumes `window` element will always handle the scrolling (which is often not the case).

### Solution
Had to copy and rewrite the class from the library a bit to support the case where scrolling is not happening on window. Removed the dependency for now.


mihow (approved):
It works! This is why open source software is great :) I recommend you open a pull request in the vh-stickt-table-header repo!

PR #496, opened 2024-08-07
Fixes & features for higher taxon ranks

mihow:
The occurrence table view can now be filtered by a taxon higher than species and will show all children.

For example, filter by all occurrences predicted to be under the Tortricidae family:
image

For now, the the species detail modal for higher taxa is a bit of a placeholder. But this is how you get to the occurrences table filtered by the Tortricidae family:
image


ami/main/models.py
```diff
@@ -2106,45 +2106,123 @@ def root(self):
...
+    def with_occurrence_counts(self) -> models.QuerySet:
+        """
+        Count the number of occurrences for a taxon and all occurrences of the taxon's children.
+        """
+        qs = self.get_queryset()
+
+        # Subquery to get all child taxa IDs
+        import json
+
+        child_taxa = qs.filter(
```
mihow, line 2174: Remove experiments until they work. Post woes to the django-pydantic-fields repo

ami/main/models.py
```diff
@@ -2288,21 +2379,25 @@ def list_names(self) -> str:
 
     def update_parents(self, save=True):
```
mihow, line 2380: If this Taxon parent changes, then all of its children need to be updated recursively

ami/main/api/views.py
```diff
@@ -884,6 +901,15 @@ def get_queryset(self) -> QuerySet:
 
         return qs
 
+    # def retrieve(self, request: Request, *args, **kwargs) -> Response:
```
mihow, line 904: This works and is very helpful in the UI. But consider if there is a cleaner way to do it. Basically we pre-calculate the occurrences count for list views, but bring back the occurrences_count method on the models for detail views, and omit the annotation in that case.

ami/main/models.py
mihow: @mihow Make a ticket for this issue

Issue #482, opened 2024-08-04
Add vertical markers every hour


annavik:
Or just more time labels along the X axis? Context is session details:

<img width="1208" alt="Skärmavbild 2024-08-04 kl  17 48 20" src="https://github.com/user-attachments/assets/d89105f4-6d62-4569-8be5-81ddeec3a3a6">

mihow:
I started playing with vertical lines with varying opacity for the captures. This feels more like a timeline to me, showing where captures occurred throughout the night. Especially if we add minor and major ticks for hours, 10 minute intervals? This should play nicely with the detections line chart on top as well.

To be continued!

![IMG_0543](https://github.com/user-attachments/assets/c546c56f-f94b-4a8c-9346-6ce0630419d8)

mihow:
Also consider making `detections_count` the mean/average count of detections from captures in a tick window rather than the sum.

PR #499, opened 2024-08-08
Fixes to the session captures timeline data

mihow:
- Fixes the issue where clicking on a "spike" (interval) activates the _first_ capture in the interval, which may have no detections and feels confusing. Now when clicking on a spike the capture with the most detections is activated instead of the first capture in the interval.
- Fixes the num captures displayed per interval. Now uses the mode average of detections from captures within the interval (ignoring zeros). Before it was a sum of all detections in the interval, which felt misleading. 
- Fixes the issue where capture ticks were not visible if the number of detections is high. Now the captures Y axis has its own range (centered around its average) that is separate from the range of the detections y axis.
- Added vertical grid lines on the hour marks! Just a quick way to do it until we can add labels in #482 

Before:
image
After:
image

Before:
image
After:
image

### DIFF
--- a/ui/src/data-services/models/timeline-tick.ts
+++ b/ui/src/data-services/models/timeline-tick.ts
@@ -70,6 +70,6 @@ export class TimelineTick {
       },
     })
 
-    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Avg. Detections: ${this.avgDetections}`
+    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Avg. detections: ${this.avgDetections}`
   }
 }
--- a/ui/src/data-services/models/timeline-tick.ts
+++ b/ui/src/data-services/models/timeline-tick.ts
@@ -11,6 +11,7 @@ export type ServerTimelineTick = {
   } | null
   captures_count: number
   detections_count: number
+  detections_avg: number
 }
 
 export class TimelineTick {
@@ -26,6 +27,10 @@ export class TimelineTick {
 
   get numDetections(): number {
     return this._timelineTick.detections_count ?? 0
+  }
+
+  get avgDetections(): number {
+    return this._timelineTick.detections_avg ?? 0
   }
 
   get numCaptures(): number {
@@ -65,6 +70,6 @@ export class TimelineTick {
       },
     })
 
-    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Detections: ${this.numDetections}`
+    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Avg. Detections: ${this.avgDetections}`
   }
 }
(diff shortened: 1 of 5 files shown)


ui/src/pages/session-details/playback/activity-plot/activity-plot.tsx
```diff
@@ -69,14 +90,30 @@ export const ActivityPlot = ({
...
+              showline: false,
+              showgrid: true,
+              griddash: 'dot',
+              gridwidth: 1,
+              gridcolor: gridLineColor,
+              showticklabels: false,
+              zeroline: false,
               fixedrange: true,
-              range: [session.startDate, session.endDate],
+              range: [new Date(session.startDate), new Date(session.endDate)],
```
annavik, line 115: `session.startDate` is already a date, any reason we are creating a new date? :)

annavik (commented):
Skärmavbild 2024-08-08 kl  17 34 10

When number of detections are 0, can we make the y axis still start at 0?

ui/src/pages/session-details/playback/activity-plot/activity-plot.tsx
```diff
@@ -69,14 +90,30 @@ export const ActivityPlot = ({
...
             },
-            xaxis: {
-              showline: true,
+            yaxis2: {
               showgrid: false,
               showticklabels: false,
               zeroline: false,
+              rangemode: 'nonnegative',
+              fixedrange: true,
+              range: [0, session.detectionsMaxCount],
```
annavik, line 102: For example (regarding comment above):

```suggestion
              range: [0, session.detectionsMaxCount ?? yAxisMax],
```

ui/src/data-services/models/timeline-tick.ts
```diff
@@ -50,6 +70,6 @@ export class TimelineTick {
       },
     })
 
-    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Detections: ${this.numDetections}`
+    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Avg. Detections: ${this.avgDetections}`
```
annavik, line 73: Nit (just to be consistent with other labels):
```suggestion
    return `${timespanString}<br>Captures: ${this.numCaptures}<br>Avg. detections: ${this.avgDetections}`
```

annavik (approved):
Thank for fixing this, it was bugging me as well, especially clicking a spike and the sometimes outshined capture graph. I think it looks great, just that case when no detections I thought looked a bit strange, but no biggie!

Pushed a tiny fix to format numbers in the capture picker.

ui/src/pages/session-details/playback/activity-plot/activity-plot.tsx
mihow: You are right, I was experimenting with date formats was trying different approaches. Reverting!

mihow:
> Skärmavbild 2024-08-08 kl 17 34 10
> When number of detections is 0, can we make the y axis still start at 0?

Fixed!

PR #502, opened 2024-08-10
Complete full stack environment for dev & CI

mihow:
Addresses https://github.com/RolnickLab/ami-platform/issues/390, a blocker for #349


mihow:
@annavik there aren't many frontend changes here, but I did update the Vite config to default to localhost:8000, but it will look for a `.env` file where you can point `API_PROXY_TARGET` to the remote env. Or create a `.env.staging` file and set `API_PROXY_TARGET` there, then run yarn with `yarn start --mode staging`. Let me know if you want to take a different approach. I also want to build a version of the frontend into the docker compose stack to make getting started easy.

PR #522, opened 2024-08-16
Update and clarify logic to fix capture upload duplicates

### LINKED ISSUE #407: When uploading multiple example captures, some captures might get duplicated
Steps to reproduce:
- Go to a deployment
- Press "Edit"
- Go to section "Source images"
- Press "+" in section "Sample captures"
- Upload 4 example captures in batch
- Note that 5 (not 4) example captures are added to the deployment (at least sometimes?)

Not sure if a FE issue or a BE issue yet, needs some more investigation!

annavik:
### Summary
We had some quirks in the upload logic for captures on the FE side, which made us sometimes call BE upload more than once for one file. In this PR we address this by including capture ids in the logic. Also we try to clarify the logic a bit by some variable name changes and comments. The upload logic is a bit complicated and the main reason for the bug was probably that the code here was a bit hard to follow.

Fixes #407 

### Updated logic
- When a file is about to get uploaded, we first add the data to an upload queue (see state variable `addQueue`).
- When a new queue item is mounted, we trigger the backend upload. The image shows up in UI with a loading spinner.
- When the upload is complete, we update the queue item with a capture id and status. We also trigger a refetch of deployment data. The image will still be visible in UI with a loading spinner (we are now waiting for deployment data to be updated).
- When deployment data is updated, we can re render the list of deployment captures. Now we can remove the item from the upload queue for good by comparing ids (deployment captures vs. queue items). The image will now be visible as a regular deployment capture.

### Notes
Comment for @mihow: while working on this, I noticed it seems we only get 10 example captures in the BE deployment response. Since the upload limit is 20, it could become a bit confusing to not actually see all captures you may have uploaded?


mihow:
@annavik Thanks for working on this! When I went to go increase the limit of example captures displayed, I realized we confounded "example captures" with "manually uploaded captures". I remember talking about that when we first made the feature but it was a rushed timeline so I think we made it super simple.

Anyway! I added a new attribute to the response and the manually uploaded captures now shows only captures that were actually manually uploaded, and they are sorted by creation date (so they shouldn't jump around so much). Whereas the example captures shows any captures from the deployment, sorted by filesize like we have had.

I made a few updates in the UI code to get there. Hopefully you can fix my dirty work! 

image
image

ui/src/utils/numberFormats.ts
```diff
@@ -0,0 +1,11 @@
+export const bytesToMB = (bytes: number) => bytes / (1024 * 1024)
+
+export const getTotalLabel = (
```
mihow, line 3: I thought this was used in multiple places so I moved to utils! But then realized it was only in on place. But maybe its still helpful to start a utils module for number formatting?

ui/src/pages/deployment-details/deployment-details-form/section-example-captures/section-example-captures.tsx
```diff
@@ -45,7 +45,21 @@ export const SectionExampleCaptures = ({
 }: {
   deployment: DeploymentDetails
 }) => {
-  const [files, setFiles] = useState<File[]>([])
+  const [addQueue, setAddQueue] = useState<
```
mihow, line 48: Nice work on adding a queue system for uploads! Perhaps we can use that pattern other places where we have a lot of single requests for bulk change.

ui/src/utils/numberFormats.ts
annavik: Yes that's great! I start adding some unit tests for these functions as well

annavik:
Thanks for separating uploaded captures from data source captures, I think it makes a lot of sense! Code wise I just tweaked the copy a bit to make it more clear:

Skärmavbild 2024-08-21 kl  12 23 43

Skärmavbild 2024-08-21 kl  12 24 00

annavik:
I tested both frontend and backend changes locally by uploading captures to a new deployment. Works well!

PR #525, opened 2024-08-19
Make it possible for logged in users to update email and password

annavik:
In this PR we make it possible for logged in users to update email and password. Before you had to contact an administrator to do so.

### Some issues to look into:
- BE response when updating email is successful, but the email is just partially saved (new email is not returned with user info, in admin the new email is only visible in some places).
  - Update: fixed!
- BE response when updating password is successful, but the password seems to be the same.
  - Update: fixed!

### User flow:
We show email and password as static values, but with edit buttons visible.
Skärmavbild 2024-08-19 kl  14 25 37

On edit button click, a mini form shows up. The mini form will have its own isolated state. This is because we are making special backend calls on email/password update, outside the user info update.
Skärmavbild 2024-08-19 kl  14 25 48


mihow:
@annavik there was one required setting missing in the Django Djoser settings (https://djoser.readthedocs.io/en/latest/settings.html#password-reset-confirm-url). I updated that and I am able to reset my password!

The one thing I noticed is that I couldn't use my password manager with the form fields. When I try to make a selection the focus jumps to the next field. 

image

annavik:
Thanks a lot @mihow!! Those URL:s seemed to fix things here and also some problems I had when I started #526 got solved.

I will look into that password manager problem, what manager is that do you use?

annavik:
@mihow I added an issue about the password manager thing, I suspect could be a similar situation for password reset, at least worth double checking. Marking this one as ready for review so we can get it out of the way! :)

PR #526, opened 2024-08-20
Setup UI for password reset

annavik:
In this PR we setup the UI for password reset. There are some small tweaks to be made to the reset instructions email, see comment sections.

### User flow

1) Link to reset password is now present on login page:
Skärmavbild 2024-08-20 kl  14 22 51

2) On reset click, user will first get asked to enter email:
Skärmavbild 2024-08-20 kl  14 22 58

3) On submit success, we hide the form and show a message.
Skärmavbild 2024-08-20 kl  18 28 17

4) From the email, the user will click a link including a temporary auth token (see query param). The link will lead to this form for updating password. The auth token will be passed with the update request, making it possible for a user that is not logged in to still change the the password.
Skärmavbild 2024-08-21 kl  10 14 16

5) On update success, we hide the form and show a message. The user can now login with the new password.
Skärmavbild 2024-08-21 kl  10 30 28

6) Example of the form in error mode (for example if the backend is responding with an error):
Skärmavbild 2024-08-21 kl  10 26 40


annavik:
Some questions!

For @mihow and @joseepoirier :
- Is the copy and UI making sense?
- Perhaps we can use a global email as the sender address here?

For @mihow :
- Can we update the email callback URL? The FE app is currently assuming **[APP URL]/auth/reset-password-confirm?token=[TOKEN VALUE]** but I can also be flexible here :)
- When making a [reset password confirm](https://api.beluga.insectai.org/api/v2/docs/#/users/users_reset_password_confirm_create) request, the param `uid` is required. I'm not fully sure what to pass here and also how to get this ID when I'm not logged in. Let's have a chat later!
- Is it possible to make the email callback URL dynamic? For testing, it could be nice to have it adjust depending on where the request is coming from (localhost, stage or prod)? Not a blocker though, since actual users will always use prod.
- The email I receive when I make a password reset usually ends up in the spam folder and looks a bit suspicious 😅 Is there anything we can do about that? If not at this point, we can also add some copy to the UI about checking the spam folder.

Skärmavbild 2024-08-20 kl  14 46 01

 It's cool I get it in Swedish though!

joseepoirier:
Great work @annavik ! I love that the little fury friend is absent when one is missing their password. 😄

1. The flow looks great. Would it be possible to hide the form upon successful submit of "send me instructions", much like you have upon setting a new password?
2. Could we remove "Please set a new password" in screen # 5 so only the green Success message box (and back to Log In) is visible?

annavik:
> Great work @annavik ! I love that the little fury friend is absent when one is missing their password. 😄
> 
> 1. The flow looks great. Would it be possible to hide the form upon successful submit of "send me instructions", much like you have upon setting a new password?
> 2. Could we remove "Please set a new password" in screen # 5 so only the green Success message box (and back to Log In) is visible?

Thanks Josee! Fixed ✅ Also updated PR description with new screenshots

mihow (approved):
It works! Great work! It's going to be a popular feature I know it :)

PR #527, opened 2024-08-20
Update password reset confirm URL

annavik:
To match the app updates in #526. 

Looked like this in email before, which is why I skipped the "#":
Skärmavbild 2024-08-20 kl  19 58 33

Is it possible to make base URL dynamic based on where the FE request is coming from?


mihow:
I realized the hostname is coming from the Site entry in the database. It defaults to the production hostname and can be changed here: http://localhost:8000/admin/sites/site/
image

I updated it on the staging site. But for local I will have to update our bootstrap script.

PR #530, opened 2024-08-21
Fix pagination bar appearance on info dialog open

### LINKED ISSUE #528: On info page open, pagination numbers get messed up
Skärmavbild 2024-08-20 kl  20 34 08

annavik:
Fixes #528 

Funny bug, the reason was clashing search param keys (both info page dialog and the pagination bar were checking the key "page"). Fixing by renaming "page" -> "info-page" for the info page dialog.


mihow (approved):
Ha yes that makes sense, clashing params! Thanks for fixing.

PR #534, opened 2024-08-21
Updates to the README for getting started

mihow:


mihow:
@mihow fix the need for creating `./backups/local/postgres` manually before running docker compose up. Add docs about sharing DB snapshots.

PR #536, opened 2024-08-22
Disable sign up

annavik:
Hide sign up from UI and make user registration disabled from API as well (make endpoint only possible to call for admin users). When we have the waitlist, we can add a link to it from the login page.

### DIFF
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -352,6 +352,9 @@ DJOSER = {
         "user": "ami.users.api.serializers.UserSerializer",
         "current_user": "ami.users.api.serializers.CurrentUserSerializer",
     },
+    "PERMISSIONS": {
+        "user_create": ["rest_framework.permissions.IsAdminUser"],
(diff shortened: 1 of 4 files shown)


config/settings/base.py
```diff
@@ -352,6 +352,9 @@
         "user": "ami.users.api.serializers.UserSerializer",
         "current_user": "ami.users.api.serializers.CurrentUserSerializer",
     },
+    "PERMISSIONS": {
+        "user_create": ["rest_framework.permissions.IsAdminUser"],
```
annavik, line 356: Let me know if you would do it like this @mihow? I had a look here: https://djoser.readthedocs.io/en/latest/settings.html

mihow: Thank you for finding this setting. Admin for user_create is perfect for the demo. I will make it environment-specific next week.

mihow (approved):

PR #537, opened 2024-08-22
Only call endpoint /me if user is logged in

annavik:
No need to poll this endpoint if we already know the user is not logged in.


mihow (approved):
Thank you!

Issue #541, opened 2024-08-22
Column selection in Occurrences view does not save


joseepoirier:
When selecting columns in the Occurrences View, the selection is lost if you navigate away from the page.

annavik:
I'll take this over since I believe we are going for a FE only solution here?

mihow:
That makes sense. I think we could start building a user preferences object on the frontend for all of the user preferences during a browser session, and then persist that entire object in the database. So make it available to read & write from via the /me endpoint. Thoughts?

mihow:
Another option for this one specifically is to save the column selections into URL params. So when you share a link the receiver sees the same columns you are seeing. Or both?

annavik:
That's a great idea @mihow, I fixed this and reworked the structure of all preferences in #544. In that PR we now use one single object for all preferences. When we are ready to sync these preferences to backend, we can do that from one place in code.

To share column preferences as part of URL params I feel could get a bit more messy...? Would those params override any preferences the user already has? I suggest we keep it simple for now, until we see a need for this, what do you think? :)

mihow:
Closed in https://github.com/RolnickLab/ami-platform/pull/544

Issue #351, opened 2024-02-16
Use React library for Google Analytics 


mihow:
Currently GA is added via a snippet in the main ui/index.html template, but there is a library that will better monitor interactions in React and allow us to use an env variable for the GA code.
https://www.npmjs.com/package//react-gtm-module

See discussion in https://github.com/RolnickLab/ami-platform/pull/338

mihow:
Closed in https://github.com/RolnickLab/ami-platform/pull/544

PR #544, opened 2024-08-27
Implement handling of cookie consent and related logic

### LINKED ISSUE #350: Add GDPR / cookie acceptance modal
See discussion in https://github.com/RolnickLab/ami-platform/pull/338

annavik:
### Summary of changes
- Setup cookie consent dialog
- Centralize handling of user preferences
  - Keep all preferences in one single object (threshold, recent identifications, column settings). This centralized handling will make it easier to also store preferences on backend as one object, when we are ready for that.
  - If user has given cookie consent, store these preferences in local storage
  - If user has not given cookie consent, handle these preferences as a global app state
- Conditionally load and unload GTM scripts, depending on cookie consent status

Fixes #350, #351 and #541 

### Summary of cookie dialog logic
- The cookie modal will be visible until the user has made a choice
- Necessary cookies cannot be disabled
- Cookies of category functionality or performance can be enabled or disabled
- "Refuse cookies" will disable all cookies, except necessary ones
- "Accept cookies" will enable all cookies

### Cookie categories
This is how we currently categorize the app's stored items.
- Necessary cookies
  - Auth token (ami-auth-token)
  - Cookie consent (ami-cookie-consent)
- Functionality cookies (all stored in ami-user-preferences)
  - Score threshold preference
  - Recent identifications
  - Visible table columns
- Performance cookies
  - Google tag manager (multiple with prefix _ga)
  
### Comments
We are awaiting response from Mila Legal on approval on both copy and logic for cookie handling. The current implementation is based on previous experiences and other sites from Mila. 

### Screenshots
Skärmavbild 2024-08-27 kl  16 28 31

Skärmavbild 2024-08-27 kl  16 28 58


ui/src/components/analytics.tsx
```diff
@@ -0,0 +1,39 @@
+import { useEffect } from 'react'
+import { useCookieConsent } from 'utils/cookieConsent/cookieConsentContext'
+
+const GTM_ID = 'G-EX26RXX0YT'
+
+export const Analytics = () => {
```
annavik, line 6: In #351 we said to use a React lib for this, however [the most popular one](https://github.com/alinemorelli/react-gtm) for Google Tag Manager did not work out of the box and after a closer look it seemed pretty outdated.

I suggest we use this custom implementation, at least until we have more advanced analytics. It's pretty much same as before, except we load and unload the scripts depending on cookie consent status.

mihow (approved):
Great work! I am very excited about the new `userPreferences` object!

PR #550, opened 2024-09-04
Setup page "Terms of service"

annavik:
### Summary
We export "Terms of service" from [this Google doc](https://docs.google.com/document/d/10MN7IPXfVFFqlHRCsm6LEGkzudJcSupL) in markdown file format. The content is rendered on a new route /terms-of-service, using a similar style as other info pages. We are also converting term titles to anchor links, as part of the rendering.

As part of this PR we are also converting info dialogs to full screen pages, to streamline the routing. I decided to remove the links from the header until the content is making sense here. The pages are still present on /about and /help though.

### Screenshots
Skärmavbild 2024-09-04 kl  12 02 52
Skärmavbild 2024-09-04 kl  12 03 03


ui/src/app.tsx
```diff
@@ -82,6 +83,7 @@ export const App = () => (
           <Route path="collections/:id" element={<CollectionDetails />} />
           <Route path="*" element={<UnderConstruction />} />
         </Route>
+        <Route path="/:slug" element={<InfoPageContainer />} />
```
annavik, line 86: Checkout this generic route, BE can return page details for any slug here and we will try render it.

ui/src/components/info-page/terms-of-service-page/terms-of-service-page.tsx
```diff
@@ -0,0 +1,64 @@
...
+        <ReactMarkdown
+          className={styles.content}
+          components={{
+            ol: (props) => {
+              const start = props.start ?? 1
+              const children = (props.children as ReactElement[]).filter(
+                (element) => element.type === 'li'
+              )
+
+              // TODO: Can we find a better way to detect what is a term title?
```
annavik, line 42: This is perhaps not an optimal way to detect what is a term title (this assumes all lists with one child is a term title). I did it to avoid modifying the markdown export, it works for now.

ui/src/components/info-page/terms-of-service-page/terms-of-service.md
```diff
@@ -0,0 +1,63 @@
+# **Terms and Conditions (August 1, 2024)**
```
annavik, line 1: Perhaps BE should return this? I think fine to keep as a FE asset for now though.

ui/src/app.tsx
mihow: This is great, thanks!

ui/src/components/info-page/terms-of-service-page/terms-of-service.md
mihow: I like that it's in the source code actually, at least for now.

ui/src/components/info-page/terms-of-service-page/terms-of-service-page.tsx
mihow: Once the Terms are finalized, we can customize the markdown file in the repo (or in the database). But I understand that this is the easiest way to keep the terms up-to-date until then.

mihow:
I can see reasons to bring back the modals, but the full-width pages are what we need most now!

PR #551, opened 2024-09-04
Make sure cover images are not overflowing frame

annavik:
In practice, seems most cover images are logos. It looks a bit strange when they overflow the container. Here comes small tweaks to keep all cover images inside the container.

Before:
Skärmavbild 2024-09-04 kl  13 31 31

After:
Skärmavbild 2024-09-04 kl  13 38 36


mihow (approved):
Thank you!!

PR #552, opened 2024-09-04
Updates to README and initial project setup

vanessavmac:
- Automatically redirect the backend API to `api/v2` and remove /home and /about pages.
- Add instructions about installing pre-commit
- Fix UI container in docker compose stack


mihow:
@vanessavmac Thank you! see the failed backend tests. It looks like it's failing the code formatting. Are you using VSCode? The project should suggest the installation of Black & auto-format on save. So you can fix those linting & formatting locally before pushing. Also! There is a pre-commit hook to check the linting & formatting before committing changes. That should be installed as part of the project setup.

Will you add instructions about installing pre-commit and running `pre-commit install`? The pre-commit package is usually installed as part of your system python packages (vs. the local virtualenv). 
https://pre-commit.com/

vanessavmac:
@mihow Thanks for the comments! 

Backend tests now pass and I added instructions to the README about installing the backend packages for IDE support + configuring pre-commit.

config/urls.py
```diff
@@ -21,11 +23,11 @@
...
 urlpatterns += [
     # API base url
-    path("api/v2/", include("config.api_router", namespace="api")),
+    path(API_ROOT, include("config.api_router", namespace="api")),
     # OpenAPI Docs
-    path("api/v2/schema/", SpectacularAPIView.as_view(api_version="api"), name="api-schema"),
+    path(API_ROOT + "schema/", SpectacularAPIView.as_view(api_version="api"), name="api-schema"),
     path(
-        "api/v2/docs/",
+        API_ROOT + "docs/",
```
mihow, line 30: urljoin() is a safer choice for url concatenation.

PR #553, opened 2024-09-04
Require open source license for project creation

annavik:
Fixes https://github.com/RolnickLab/ami-admin/issues/21

In this PR we block project creation client side, unless open source license has been selected. We still show the other option, but disabled.

Skärmavbild 2024-09-04 kl  15 07 32

Left to do on the backend side:
- Store selected license (let me know what format we should use)
- Make a similar blocking of project creation from API, if we want to be in the safe side :)

### DIFF
--- a/ui/src/data-services/hooks/projects/utils.ts
+++ b/ui/src/data-services/hooks/projects/utils.ts
@@ -13,5 +13,9 @@ export const convertToServerFormData = (fieldValues: any) => {
     data.append('image', '')
   }
 
+  if (fieldValues.openSource) {
+    data.append('license', 'open-source')
+  }
+
   return data
 }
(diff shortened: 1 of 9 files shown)


ui/src/data-services/hooks/projects/utils.ts
```diff
@@ -13,5 +13,9 @@ export const convertToServerFormData = (fieldValues: any) => {
     data.append('image', '')
   }
 
+  if (fieldValues.openSource) {
+    data.append('license', 'open-source')
```
annavik, line 17: Let's tweak this, if backend prefers the license to be passed in a different way.

mihow:
Thanks @annavik! I am going to think about the backend part. I think we should check for this acceptance before every image sync or processing job. In-case the option gets unchecked, and also for existing projects.

Issue #549, opened 2024-09-04
Cleanup for demo


annavik:
### Small things to fix:
- [x ] [BE] Job logs are a bit repetitive. Can we fix? Either on BE (best) or try cleanup FE side? #310 
- [x ] [BE] Session detail species count is not in sync (see [example] (https://demo.antenna.insectai.org/projects/3/sessions/104) where species count is still 0). #490 
- [x] [FE] If not logged in, it makes no sense to select occurrences. Skip!
- [x] [FE] Skip average temp field for session details, it's never defined.
- [x] [BE] Pipeline details are missing info.
- [x] [BE] Occurrence image sources are broken.
- [x] [FE] Hide fields "Source images" and "Pipeline" from job details, if not making sense for the job type.
- [x] [FE] If not logged in, hide some disabled controls for session details?
- [x] [FE] Adjust font weight for links in job details view
- [x] [FE] ~~Sorting of collection captures is causing error~~ Cannot reproduce anymore, perhaps something got fixed BE side?

### Screenshots:
<img width="1440" alt="Skärmavbild 2024-09-04 kl  09 33 27" src="https://github.com/user-attachments/assets/6f0ef025-12f2-4852-8a5a-eece1193b35f">

<img width="689" alt="Skärmavbild 2024-09-04 kl  09 25 10" src="https://github.com/user-attachments/assets/681dcf24-205a-4b15-8a46-2c5e5ea0d90a">

<img width="539" alt="Skärmavbild 2024-09-04 kl  09 39 35" src="https://github.com/user-attachments/assets/c503e6a2-d580-4c12-8e95-ed0c9d11d0c6">

<img width="750" alt="Skärmavbild 2024-09-04 kl  09 29 46" src="https://github.com/user-attachments/assets/c66f4b07-ab29-46b6-a540-fda8501be69b">

<img width="411" alt="Skärmavbild 2024-09-04 kl  09 28 10" src="https://github.com/user-attachments/assets/2d92b09a-5179-4b18-8a49-be60762a8252">

<img width="753" alt="Skärmavbild 2024-09-04 kl  09 36 14" src="https://github.com/user-attachments/assets/fc27f130-f791-4402-bd0c-567071c59267">

<img width="198" alt="Skärmavbild 2024-09-04 kl  10 00 31" src="https://github.com/user-attachments/assets/daff2403-c72e-4aea-825c-d67e0fc110b5">

joseepoirier:
[...]

PR #557, opened 2024-09-05
FE cleanup for demo

annavik:
Fixing all FE task mentioned in #549 . Also adding some tweaks for improved a11y and small screen experience. For example considering links, focus states, breadcrumbs, table header and session detail view.


ui/src/pages/sessions/session-columns.tsx
```diff
@@ -146,9 +146,4 @@ export const columns: (projectId: string) => TableColumn<Session>[] = (
       </Link>
     ),
   },
-  {
-    id: 'avg-temp',
```
mihow, line 150: @annavik I added a ticket for this feature so as not to forget it! #562

PR #558, opened 2024-09-06
Add taxonomy search to occurrence page (+ empty state)

annavik:
### Summary
In this PR we add a simple taxon search to the occurrence page, as a small start to the filtering feature. Also, I figured we might want an empty state for both gallery and table view (this could also be nice for new projects that are not yet populated with date). If the current view has some filtering that is active, we tweak the empty state message and present an option to clear filters.

### Screenshots
Skärmavbild 2024-09-06 kl  11 39 46

Skärmavbild 2024-09-06 kl  11 39 07

Skärmavbild 2024-09-06 kl  11 53 36


ui/src/components/empty-state/empty-state.tsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <div className={styles.wrapper}>
+      <span>
+        {translate(
+          filtersActive
+            ? STRING.MESSAGE_NO_RESULTS_FOR_FILTERING
+            : STRING.MESSAGE_NO_RESULTS
+        )}
+      </span>
+      {filtersActive && (
```
annavik, line 18: If we have no active filters, but still no items to show, we might want to suggest creating an new entity and show a button? Kept it simple for now with just the text.

ui/src/pages/occurrences/occurrences.tsx
```diff
@@ -77,6 +78,7 @@ export const Occurrences = () => {
         showAppliedFilters
         tooltip={translate(STRING.TOOLTIP_OCCURRENCE)}
       >
+        <TaxonFilter />
```
annavik, line 81: When possible to also apply filtering by determination id to species view, we can just add the same control with one line of code like this! I tried, but seems that filtering did not had any effect in species view, I skipped for now.

PR #559, opened 2024-09-06
Update terminology

annavik:
Update all copy to use latest terminology. Also make it possible to show tooltip as part of table header (we had one case where this was needed).

Copy is based on [this Google doc](https://docs.google.com/document/d/1e7znB9Chb1TAvPZxmdjxhI9tFginJ6wN3_DrkGmZTqY/edit).


mihow:
I would like to revisit this later with some long and short forms of some names. For example "Monitoring Station" for the page title, but "Stations" for the menu item

PR #567, opened 2024-09-11
Frontend workarounds for demo

annavik:
Including some frontend workarounds while we work on some backend improvements:
- Hide job tab if user is not logged in
- Set default sort to occurrences count in session tab
- Hide wrong species count in session details

### DIFF
--- a/ui/src/data-services/models/session.ts
+++ b/ui/src/data-services/models/session.ts
@@ -71,6 +71,13 @@ export class Session {
   }
 
   get numSpecies(): number | undefined {
+    // Workaround for taxa count being 0 when occurrences count > 0
+    if (this.numOccurrences) {
+      if (this._event.taxa_count === 0) {
+        return undefined
+      }
+    }
+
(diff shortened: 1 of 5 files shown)


mihow:
I like how the menu is more concise without the Jobs tab, but bulk processing is such a primary part of the app. Is it a loss to hide it? I know there is so much to comprehend when you first land on the app, so maybe it's best that way. And the demo video shows it.

ui/src/data-services/models/session.ts
```diff
@@ -71,6 +71,13 @@ export class Session {
   }
 
   get numSpecies(): number | undefined {
+    // Workaround for taxa count being 0 when occurrences count > 0
+    if (this.numOccurrences) {
+      if (this._event.taxa_count === 0) {
```
mihow, line 76: In the future I think the backend should return null for taxa_count if the data has not been processed or taxa_count not yet populated (its a cached value).

PR #568, opened 2024-09-11
Session details improvements

annavik:
**Activity chart improvements:**
- Change hover mode to "x unified"
- Show color indicators on hover
- Skip vertical lines to indicate hours for now (seems not fully clear to users what this line is)

**Session details tweaks:**
- Update plot x axis tick format. This will change [0, 1, 1, 2, 2] -> [0, 0.5, 1, 2, 2.5]. Even if we are dealing with integers for counts, I think the later version looks more correct. I would of course prefer [0,1,2], but setting a fixed tick distance is causing problems for plots with lots of ticks. We could probably try some smarter solution here when there is time.
- Hide plots with 0-1 data points (to avoid empty plots and strange looking plots with just one bar).

### DIFF
--- a/ui/src/data-services/models/timeline-tick.ts
+++ b/ui/src/data-services/models/timeline-tick.ts
@@ -1,5 +1,3 @@
-import { getCompactTimespanString } from 'utils/date/getCompactTimespanString/getCompactTimespanString'
-
 export type ServerTimelineTick = {
   start: string
   end: string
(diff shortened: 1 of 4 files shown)


mihow:
I love the new cursor! So much more clear. @annavik If we add more time labels on the x-axis then the need for the hourly indicators is gone.

PR #575, opened 2024-09-16
Restore search params after login

annavik:
Fixes this comment from @mihow :
"if you log in while on session detail (to process or favorite a photo), it jumps to the beginning of the session afterwards"


mihow (approved):
It works! This is such relief, thank you

PR #576, opened 2024-09-16
Template for an ML backend in the compose stack

mihow:
Create a ML backend in the main or a parallel docker compose config that can be run locally and in CI/CD tests. This will also serve as a template for ML researchers to use to make a new model available in Antenna.

Some todos after discussion on Sept 19th:
- New Django model "ML Backend" this will replace the selectable "Pipeline" model. The ML backend will return the list of available pipelines (which in turn provides a list of the stages and models).
- The "Pipelines" list in the Project overview should still show the list of Pipelines available, but some come from different ML backends. 
- The Pipeline & Algorithm models shouldn't be manually created
- Add a dummy ml backend to the main compose stack so the CI tests can run
- Backends should be available to specific projects, or all projects


ml_backends/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+    ports:
+      - "2000:2000"
+  ngrok:
+    image: ngrok/ngrok:latest
+    ports:
+      - "4040:4040"
+    command:
+      - "http"
+      - "http://host.docker.internal:2000"
+      - "--request-header-add='ngrok-skip-browser-warning': '69420'"
```
mihow, line 16: This request header should be added to the actual API call you are making with `axios`
```json
axios.get('https://your-ngrok-url.ngrok.io/api/endpoint', {
      headers: {
        'ngrok-skip-browser-warning': 'true'
      }
  ```

File: ml_backends/docker-compose.yml [hunk 2]
```diff
@@ -0,0 +1,25 @@
+services:
+  example:
+    build:
+      context: ./example
+    volumes:
+      - ./example/:/app:z
+    ports:
+      - "2000:2000"
+  ngrok:
```
mihow, line 9: This is an interesting approach, to have an ngrok service in the docker compose setup, but its not necessary. Ngrok is a tool you should have on your host machine. Launch the ml backend with `docker compose up`, then launch Ngrok it with `ngrok http 2000` in a separate shell tab. Then when you visit the full ngrok url, it will point to the ml backend.

ui/src/pages/session-details/playback/capture-details/capture-job/process-now.tsx
```diff
@@ -18,7 +20,19 @@ export const ProcessNow = ({
   const { project } = useProjectDetails(projectId as string, true)
   const { createJob, isLoading, isSuccess } = useCreateJob()
   const icon = isSuccess ? IconType.RadixCheck : undefined
-  const disabled = !capture || capture.hasJobInProgress || !pipelineId
+  const disabled = !capture || capture.hasJobInProgress || !pipelineId 
+
+  const getPipelineInfo = async () => {
```
mihow, line 25: We might consider calling the pipeline /info endpoint from the frontend in the future, but let's just call it from the backend for now.

ui/src/pages/session-details/playback/capture-details/capture-job/process-now.tsx
```diff
@@ -44,16 +58,20 @@ export const ProcessNow = ({
...
-            pipeline: pipelineId,
-            projectId: projectId as string,
-            startNow: true,
-          })
-        }}
+        onClick={
+          // @TODO: get the data from the completed processing and print it to console...ngrok
+          getPipelineInfo
+
+          // @TODO: I should create a job which is then processed by ML backend....how does thi createJob work??
```
mihow, line 65: @vanessavmac The createJob function calls the Django / Antenna backend and creates a new processing job for the selected pipeline. Then the Job sends requests to the ML Backend. The Job is a background task in the Antenna backend. 

Here is where the actual request is made. But you shouldn't have to modify this.
https://github.com/RolnickLab/antenna/blob/8e5438e12fe681d4b3edabe23abd4e3dc0439a71/ami/ml/models/pipeline.py#L166

mihow:
@vanessavmac I want to simplify our template for the ML backend a bit. I think we should bring it into the main docker-compose.yml 

We can still provide a separate ML backend compose stack, which will be for larger backends with more dependencies. But the template is so lightweight, that it can go in the main stack. 

The endpoint for the ML backend in the main stack will still be something like `http://ml_backend:2000` so you will still have to change the field type of the Endpoint field on the Pipeline Django model from a URL field to a CharField for either case.

If you want to finish the current approach, that's fine too. But hopefully moving it will reduce some complexity.

compose/local/django/start
```diff
@@ -6,4 +6,5 @@ set -o nounset
 
 
 python manage.py migrate
+python manage.py add_pipeline
```
mihow, line 9: I'm glad you were able to write a management command! those are very useful. However, I have another place for generating the test data when the project starts up:
https://github.com/RolnickLab/antenna/blob/main/ami/tests/fixtures/main.py

You can see there are functions for create_occurrences, create_taxa, etc. So you can move create_pipeline there.

Then you can call that function from `setup_test_project` in that same file.

This is the function that sets up the entire test project and calls a few more of the fixture functions. 
https://github.com/RolnickLab/antenna/blob/fddcd76f7a8231a43c95f1a3d33373301edd7c85/ami/tests/fixtures/signals.py#L20

ml_backends/example/api/test.py
vanessavmac, line 1: @mihow is this file still required? The ML backend test is currently in `ami-platform/ami/ml/tests.py`?

README.md
```diff
@@ -143,10 +142,17 @@ docker compose run --rm django python manage.py graph_models -a -o models.dot --
 dot -Tsvg  models.dot > models.svg
 ```
 
+### ML Backend
+
+Add the minio hostname to /etc/hosts.
```
mihow, line 147: Move this to the Data Storage section, since its relevant to the whole project not just the ML backend

docker-compose.ci.yml
```diff
@@ -1,4 +1,11 @@
 services:
+  ml_backend:
```
mihow, line 2: Will you move the ml_backend service to the bottom of the compose files?

Issue #387, opened 2024-05-03
Add top species chart to overview page


mihow:
Can replace the species accumulation chart. See the session detail page for existing top species chart, but feel free to try another chart type.

<img width="374" alt="image" src="https://github.com/RolnickLab/ami-platform/assets/158175/daf5b521-4199-486e-a744-6091eb3fc504">

![image](https://github.com/RolnickLab/ami-platform/assets/158175/600a57b5-51bb-4b6a-86f1-1a00fd3a7913)

annavik:
Hey @mihow, this can be fully controlled from backend. Checkout field `summary_data` in project details response.

Removing the frontend tag, but just let me know if some tweaks are needed for this! :)

Issue #445, opened 2024-06-28
Small backend issues related to plots


annavik:
Fixed some plot feedback from Lars in #441. There are some small things left which requires backend changes:
- Detections per hour chart is a bit strange 
- Restrict number of species if many in top species chart
- Hover label is sometimes a bit repetitive, for example in top species chart

For the hover label, I have made it possible to specify this from backend since the label might vary depending on what chart we are dealing with. Just pass a `hovertemplate` string: https://plotly.com/javascript/hover-text-and-formatting/#hovertemplate

vanessavmac:
Hi @annavik ! I'm Vanessa, the new backend/MLOps developer intern. Michael assigned this issue to me as a good first issue to get familiar with the code base.

> Detections per hour chart is a bit strange

Would you be able to clarify what issues there are with the detections per hour chart? One thing I noticed is that when I slide the x-axis, sometimes the bars get distorted or turn into weird shapes. Not sure if this is what you were referring to or if there are other changes I should make.

![image](https://github.com/user-attachments/assets/8e7768ce-12dc-4c63-82a4-1ec192eff232)

[445-small-backend-issues-related-to-plots](https://github.com/RolnickLab/ami-platform/tree/445-small-backend-issues-related-to-plots) -- So far, I've addressed the 2 other points you mentioned on this branch.

PR #577, opened 2024-09-16
Update top species chart

vanessavmac:
Addressed the 2 points in #445:
1. Restrict number of species if many in top species chart
2. Hover label is sometimes a bit repetitive, for example in top species chart

The last item "Detections per hour chart is a bit strange" remains unresolved. See this comment @annavik : https://github.com/RolnickLab/ami-platform/issues/445#issuecomment-2330350075


mihow:
Thanks @vanessavmac! This looks good!

I updated the title of this PR, how about we focus it on the species chart and fix #387 here as well? Will you:

1. Add the top species chart to the project overview
2. Flip the chart so the top species count is on top rather than the bottom? I'm not sure when that happened!

image
image

vanessavmac:
Hi @mihow, thanks for the feedback! I've opened PR #580 which includes the 2 changes you requested for the project overview. PR #580  and #577 complement one another.

mihow:
okay thanks, merging this one then!

PR #578, opened 2024-09-17
Add taxon search to species table view

mihow:
Add the hierarchical taxa search to the Species tab. This uses the same component as the Occurrences tab. I updated the url param to be `taxon` for both placements. Even though they mean slightly different things in both contexts.

image

### DIFF
--- a/ui/src/pages/species/species-columns.tsx
+++ b/ui/src/pages/species/species-columns.tsx
@@ -77,7 +77,7 @@ export const columns: (projectId: string) => TableColumn<Species>[] = (
       <Link
         to={getAppRoute({
           to: APP_ROUTES.OCCURRENCES({ projectId }),
-          filters: { determination: item.id },
+          filters: { taxon: item.id },
         })}
       >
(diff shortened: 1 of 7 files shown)


ui/src/pages/species/species-columns.tsx
```diff
@@ -77,7 +77,7 @@ export const columns: (projectId: string) => TableColumn<Species>[] = (
       <Link
         to={getAppRoute({
           to: APP_ROUTES.OCCURRENCES({ projectId }),
-          filters: { determination: item.id },
+          filters: { taxon: item.id },
```
annavik, line 80: Just curious, did you notice this had to be updated thanks to build errors? That would be my hope!

annavik (approved):
Thanks for adding this! Also the new URL param key makes more sense 👌

ui/src/pages/species/species-columns.tsx
mihow: In this case I was searching the codebase, but I'll run the tests first next time to try that!

PR #584, opened 2024-09-18
Handle captures that have no associated Event / Session

mihow:
When captures are first synced, they are not associated with a monitoring session. This fixes the collection detail view when some captures have no eventId. 

After all captures are done syncing, the events should be created in the background. However there are some cases where this takes a long time, or some captures don't get added to an event. This should not be a common case, but I have seen it a few times. Also in the future, some types of projects may want to group images differently than by event, for example in continuous monitoring projects.

image


ami/main/models.py
```diff
@@ -1165,7 +1165,14 @@ class SourceImage(BaseModel):
...
     project = models.ForeignKey(Project, on_delete=models.SET_NULL, null=True, related_name="captures")
     deployment = models.ForeignKey(Deployment, on_delete=models.SET_NULL, null=True, related_name="captures")
-    event = models.ForeignKey(Event, on_delete=models.SET_NULL, null=True, related_name="captures", db_index=True)
+    event = models.ForeignKey(
+        Event,
+        on_delete=models.SET_NULL,
+        null=True,
+        related_name="captures",
+        db_index=True,
+        blank=True,
```
mihow, line 1174: The only BE change here is allowing blank values in the UI for the Django admin CMS. The event can already be null in the database.

PR #595, opened 2024-09-25
Updates to overview tabs

### LINKED ISSUE #594: Add project overview tabs to URL params & auto-refresh
We are using the collections list and other tabs on the project overview page more and more, however when using the back button or sharing the link, the active tab is not preserved. 

Can we add the tab as a URL param and jump to the section title? 

Also confirm if auto-refresh on the collections tab is working like on other table views. 

Right now I click "Populate collection", then am refreshing manually, scrolling down, clicking on Collections, and repeating until it is done.

image

annavik:
Fixes #594 

### Summary of changes:
- Add overview tabs to URL params
- Add polling to collection endpoints
- Update population status logic (show button as loading while the collection is being populated)
- Show correct label in "Created at" column for collections

### DIFF
--- a/ui/src/data-services/hooks/collections/useCollections.ts
+++ b/ui/src/data-services/hooks/collections/useCollections.ts
@@ -6,6 +6,8 @@ import { useMemo } from 'react'
 import { UserPermission } from 'utils/user/types'
 import { useAuthorizedQuery } from '../auth/useAuthorizedQuery'
 
+const REFETCH_INTERVAL = 10000 // Refetch every 10 second
+
 const convertServerRecord = (record: ServerCollection) => new Collection(record)
 
 export const useCollections = (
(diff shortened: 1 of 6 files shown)


ui/src/data-services/hooks/collections/useCollections.ts
```diff
@@ -6,6 +6,8 @@ import { useMemo } from 'react'
 import { UserPermission } from 'utils/user/types'
 import { useAuthorizedQuery } from '../auth/useAuthorizedQuery'
 
+const REFETCH_INTERVAL = 10000 // Refetch every 10 second
```
mihow, line 9: Isn't this constant already somewhere else? For the table views?

mihow (approved):
I left one comment, but merging now because its a great improvement

Issue #585, opened 2024-09-19
Option to skip or hide unprocessed captures


mihow:
In the session detail view, it can be tricky (and confusing) to navigate to frames that have detections ready to view. An option to ignore the unprocessed captures would be helpful. I propose:

- A checkbox titled "Skip unprocessed" next to the prev/next navigation buttons 
 - Checked by default
 - When enabled, it disabled the ability to click on an unprocessed capture in the timeline
 - When enabled, the next/prev & keyboard navigation skip to the next processed captured
 - When the session detail page loads and no capture is specified in the URL, load the first processed capture (The backend likely needs to return this)

![image](https://github.com/user-attachments/assets/8d48d2f1-80bf-4dc8-ba3b-8d49aa04ae50)

@rhine3 @annavik  thoughts?

rhine3:
This looks great! Would definitely make sessions easier for me to navigate. Here are a couple additional suggestions you can take or leave.

- People might not see the automatically-checked box and may wonder why their session appears to only have, say, 40 images. I wonder if somewhere on the session page, you could also add a legend in the same style as the legend that's displayed when you hover over the timeline, but one that represents overall stats for the session. For example, it could show the total number of captures in the session, the total number of processed captures, and the average number of detections per processed capture

- If images in the session have been processed by multiple pipelines: should it display any processed image, or only images processed by a particular pipeline? Is there a way to tell which images were processed by which pipeline, e.g. maybe using additional colors besides the blue to indicate them?

annavik:
Hi @mihow and @rhine3 ! I have started to look into this a bit :) See #596 
- I followed your suggestions Michael, with some small logic adjustments due to some limitations in the backend response. Also I placed the control alongside other view settings.
- Good point Tessa with the session overview! In my suggestion, I kept the graph as is, to keep this overview. Just updated the navigation behaviour, what do you think?
- For images that has been processed by multiple pipelines, I think it requires a bit of more work! Currently we are fully prepared to handle this on the frontend side, as you might have noticed. Also I will need to have a chat with Michael about the BE status for this. I will create a separate issue about this, thanks for bringing it up! :)

mihow:
closed by #596

PR #596, opened 2024-09-25
Snap to tick with detections

annavik:
### Summary
In this PR we suggest a solution to #585 with a new new view setting "Snap to tick with detections". When checked, we will snap to the next tick with detections when using one of the three ways to navigate images:
- When using arrows/keyboard for going to next/previous image
- When clicking the graph
- When using the timeline slider

### Comments and limitations
At the moment, the timeline response from backend does not include processing status for images. From a frontend perspective, we only know if a timeline tick has detections or not. In practice a processed image might not include any detections, this is why I named the view setting this way, without mentioning processing status.

Another detail to mention is that a timeline tick might include multiple images with detections. At the moment, backend only returns the id for the "top image" in the interval. This is why we call the view setting "snap to tick" instead of "snap to image". If the timeline endpoint could provide some more info about images in an interval, we could update this. We have discussed one tick/image before @mihow . If not to much data do handle at once, I think this could reduce some limitations...?

Regarding the first capture that will be visible, currently this ID is specified by the backend session response. Do you think we should change this, to return ID for the first capture with some detections?

### Screenshot
Skärmavbild 2024-09-25 kl  14 48 31


mihow:
- I agree it make sense to snap the cursor to ticks with detections as well
- I agree that the checkbox may be more apparent next to the Prev/Next arrows. 

I am going to merge this so we can use the current version this week.

PR #603, opened 2024-10-21
Setup score threshold filter for list views

annavik:
**Background**
Currently, the score threshold value is saved as a user preference. The value can be changed from the score slider in session detail view. Now, we want to pass this value as a filter for some list views.

**Summary**
In this PR, we pass the score threshold preference as a filter setting for occurrences and species. We also display the applied value like any other filter (see screenshot). Since the score filter can't technically be cleared, we will hide this action.

Just like any other filter, the score filter can also be changed from URL (using key `classification_threshold` in this case). This will override the stored preference, but not change it.

**Thoughts**
- Should we add a control to modify the filter value from the list views? I think it feels a bit strange the only way to change the value is from the session detail view... Perhaps this setting should be presented more like a global preference, with the option to override separately in both list views and detail views?
- With this new filter, the summary and total number of records will not add up. If to treat the score threshold like any other filter, the summary count should probably be the total number of records?

**Screenshot**

Skärmavbild 2024-10-21 kl  13 12 47


mihow:
Thank you for working on this! I ended up setting the threshold to zero for Max's demo last night for a quick fix. But this adjustable one will be very helpful.

I agree that it is strange to set it from the session detail view. Where could we put a global filter?

I think step 1 is making it so the threshold filter can be passed by URL, and displaying the currently active score, as you have in this PR. 

We could move the next interface decisions to another PR! Here are some thoughts on those:

We also need to decide if other components need the filter applied, like the main stats bar & the charts on the project overview, and the occurrence counts in the bubble number links. Currently those all are being filtered by the default threshold.

There are some additional complexities to consider. Currently we are pre-filtering out all non-moths. But if you actually want to include non-moths in counts, then that is another filter based on the model output. So when viewing occurrences we will likely need filters for the output of multiple models. Something like:
```json
{
model_output_filters: [
  {
    'quebec_species_classifier': {'softmax_score__gte': 0.6},
    'moth_nonmoth_classifier': {'labels': ['moth']}
  }
}
```
And as we discussed before, we will need someway to toggle between models. To compare results from one to another. To keep this all under control I think we can use a presets system. So most of the time you don't see so many parameters, but can select some standard defaults that come with every project and pipeline.

annavik:
Thank you for those thoughts! I'll keep them in mind for step 2 of this feature.

I'll merge this!

PR #608, opened 2024-11-06
New filters for Occurrences

mihow:
Added the following date filter query params:

- `date_start` (greater than or equal to the earliest date of a detection within an occurrence)
- `date_end` (less than or equal to the latest date of a detection)

These should be in ISO format (2023-01-01) and are assumed to be in the same local timezone that the occurrence timestamps are in. 

These filters can have multiple instances and will be converted into lists
- `algorithm` filter occurrences that have detections that have classifications that used these algorithm IDs
- `not_algorithm` the same, but exclusive

Include or exclude multiple like:
http://localhost:8000/api/v2/occurrences/?not_algorithm=6&not_algorithm=3
http://localhost:8000/api/v2/occurrences/?algorithm=6&algorithm=3

This may look strange but this is how html multi-select forms will create the query params if you submit them in a GET request. 

Filters for the identification status:

- `identified`, 1 or 0 or true or false, to show occurrences that have or have not been identified by a person
- `identified_by` filter occurrences that have been identified by a certain user id (supports multiple)
- `not_identified_by` filter occurrences that haven't been IDed by a user id (supports multiple)


annavik:
I included some UI controls for for algorithm filters. To start with, the user can set a single value for each filter. We populate the options by calling `/ml/algorithms`.

Skärmavbild 2024-11-12 kl  14 40 19

Let me know what you think. Is setting multiple values important for example?

Next up, date filtering! :)

mihow:
Fantastic, thanks @annavik! I think single selection is great for now!

mihow:
@annavik added filters for the identification status! added details to the PR description

annavik:
Added a ticket for UI controls to "identified by": https://github.com/RolnickLab/antenna/issues/615

We must figure out how to define the list of users to show, but I don't want that to be a blocker for this PR! :)

mihow:
@annavik I changed "identified" -> "verified" as you pointed out, thanks! I also re-ordered a few of the filters. I wonder if we can make it easier to define sections instead of only Filters & Advanced filters? I moved Stations to advanced for Occurrences, but left it under regular Filters for Taxa, since that page doesn't have any other filters. 

Also! I swapped out VerifiedBy for VerifiedByMe! Which is a boolean filter to check what has been verified by the currently logged-in user. We can test that in #615

annavik:
Sorry, I forgot to get back to you on this! I have actually updated the filtering component to make it more flexible. Instead of passing a config object, we now render controls as children. This will make it easier to customize for different views. Small things left would be to add props for `title` and `defaultExpanded`, then I think we are all set for custom filter sections?

Screenshot 2024-12-03 at 12 14 56

PR #620, opened 2024-11-19
Use inverted colors for score sliders (+ some refactoring)

annavik:
### Summary
- Use inverted colors for score threshold everywhere
- "Primary blue" as fill color in occurrence list view
- "Secondary blue" as fill color in session detail view (we needed some more contrast here)
- Occurrence frame colors will now match the ring color in list list view. Logic based on score, see intervals:
  - [0.8, 1]: "success green"
  - [0.6, 0.8): "warning orange"
  - [0, 0.6): "destructive green"
- In session detail view we now use the same slider as in the list view (our new one from Nova UI Kit).
- Remove unused slider code and rework the component structure

### Comments
My plan is to remove more code like this, while we gradually move towards using Nova in more places. My hope is to keep the Antenna FE code more clean this way, keeping it focused on Antenna specific FE logic rather than styling! :)

### Screenshots
After:
Screenshot 2024-11-20 at 12 39 32
Screenshot 2024-11-20 at 12 39 08


annavik:
My suggestion on updates to this PR after discussion on Slack (thank you @mihow !):
- We revert updated score slider color to generic blue
- We update occurrences frame colors in session detail view, to follow same logic as the circles in list view

mihow (approved):

PR #626, opened 2024-11-20
[Draft] Add filters to jobs

annavik:


mihow:
Thanks for jumping on this! It looks great. I will fix the backend filters you pointed out. Also maybe try some links from the Collections view and others.

annavik:
**Update:**
We said adding backend support for the missing filters could be a good first backend task for Anna. To not block merging of this PR, controls for filters that are missing backend support are hidden for now. Description is updated and this one could get merged after review :)

mihow (approved):
Moving right along! This is so helpful.

PR #632, opened 2024-11-25
Enable users to register Processing Services & Pipelines

vanessavmac:
## Summary of Changes
- Users can register an ML Processing Service (a collection of pipelines)
- Register pipelines to the Processing Service using an 'Register Pipelines' button
- To process an image, users select the pipeline. The job checks which services are online before selecting one with lowest latency/response time.

## Example User Flow

Users can register processing services associated with an endpoint URL. The endpoint URL has an `info` endpoint which returns a list of pipelines that should be in this processing service. The "Register Pipelines" button pings the `info` endpoint and adds the pipelines and algorithms to the processing service.

Screenshot 2025-01-19 at 12 13 26 AM

Pipelines are listed in the overview page along with the number of processing services available.

Screenshot 2025-01-19 at 12 14 28 AM

Screenshot 2025-01-19 at 12 19 16 AM

When a processing service is available, the user can select a pipeline to process a capture with. The processing service with the lowest latency is used to process the images with the selected pipeline.

Screenshot 2025-01-19 at 12 16 10 AM

## Follow-Up ToDos (Requires feature spec-ing)

- Revamp the processing page. 
    - Currently, users select a pipeline and the function `choose_processing_service_for_pipeline()` selects a service that has the lowest latency. The UI could display what services are online and prevent users from using pipelines with no available processing services.
    - Recommend better pipelines (i.e. most recently checked, lowest latency, etc)
 - Come up with a better solution than "Stages" for showing the sequential order of algorithms in a pipeline. This processing service feature PR just uses the default values for "Stages."

## Follow-Up Issues

- `ami/ml/models/pipeline.py` Address the TODO (# TODO the algorithms list be retrieved by querying the pipeline endpoint) #681
- `ami/ml/views.py` Rather than select a random image to test the pipeline processing (`test_process`), filter images by projects user has access to. #680
 - Currently we just add and create pipelines. Once pipelines are added, users cannot at the moment modify the pipelines in that processing service. For example, if the user registers pipelines then changes the `info` endpoint on the service, when the user registers pipelines again, it will only add additional pipelines (pipelines won't be deleted or updated). #678
- Use celery beat for periodic task to check when processing services are last seen online (similar to this http://localhost:8000/admin/django_celery_beat/periodictask/1/change/) #676 
- Add section to pipeline details with a json field of pipeline parameters #677


ami/ml/models/pipeline.py
```diff
@@ -431,11 +447,19 @@ def collect_images(
             skip_processed=skip_processed,
         )
 
+    def choose_backend_for_pipeline(self):
+        # @TODO: Create function to backend from the current project and most recently responded OK to a status check
```
mihow, line 451: Raise an error if no backends are online for this pipeline to run

ami/ml/schemas.py
```diff
@@ -62,6 +62,65 @@ class PipelineRequest(pydantic.BaseModel):
...
+class PipelineConfig(pydantic.BaseModel):
+    """A configurable pipeline."""
+
+    name: str
+    slug: str
+    version: int
+    description: str | None = None
+    algorithms: list[AlgorithmConfig] = []
+    stages: list[PipelineStage] = []
+    projects: list[ProjectConfig] = []
```
mihow, line 106: The ML backend will never know about the projects configured in Antenna, so you can remove projects here. These schemas are only for the ML backend requests & responses. The serializers are for the Antenna project & models. Sorry its confusing!

ml_backends/example/api/schemas.py
```diff
@@ -148,10 +148,22 @@ class PipelineStage(pydantic.BaseModel):
...
 class PipelineConfig(pydantic.BaseModel):
     """A configurable pipeline."""
 
     name: str
     slug: str
+    version: int
     description: str | None = None
+    algorithms: list[AlgorithmConfig] = []
     stages: list[PipelineStage] = []
+    projects: list[ProjectConfig] = []
```
mihow, line 169: remove projects from here

ami/ml/schemas.py
```diff
@@ -62,6 +62,65 @@ class PipelineRequest(pydantic.BaseModel):
...
+    timestamp: datetime.datetime
+    request_successful: bool
+    pipeline_configs: list[PipelineConfig] = []
+    error: str | None = None
+    server_live: bool
+    pipelines_online: list[str] | str
+    endpoint_url: str
+
+
+class PipelineRegistrationResponse(pydantic.BaseModel):
```
mihow, line 119: I don't think you need this PipelineRegistrationResponse. Registration is done on the Antenna side, so that's handled by the Django Rest Framework serlializers & viewsets

ami/ml/models/pipeline.py
vanessavmac: Updated. If no backends are online, then I return an empty pipeline response with an error message that 'No backends are online.' Should the Job status be 'FAILURE'? I noticed that the job status isn't always correct (i.e. when no images are processed it still says 'SUCCESS'. Is the job status handled by the frontend? Or the backend?

vanessavmac:
@mihow  I'm getting an error for this test in `ami-platform/ami/jobs/tests.py`: 
```
def test_run_job_unauthenticated(self):
        jobs_run_url = reverse_with_params("api:job-run", args=[self.job.pk])
        self.client.force_authenticate(user=None)
        resp = self.client.post(jobs_run_url)
        self.assertEqual(resp.status_code, 401)
```
The console logs:
```
======================================================================
FAIL: test_run_job_unauthenticated (ami.jobs.tests.TestJobView.test_run_job_unauthenticated)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/ami/jobs/tests.py", line 176, in test_run_job_unauthenticated
    self.assertEqual(resp.status_code, 401)
AssertionError: 403 != 401

----------------------------------------------------------------------
```
I think this error is occurring now after I modified the image processing to use the `choose_backend_for_pipeline()` function in `ami-platform/ami/ml/models/pipeline.py`. However, I'm not sure why exactly it's occurring. I think it might have to do with how the job status is determined?

ami/ml/schemas.py
vanessavmac: I currently use PipelineRegistrationResponse in the viewset `ami-platform/ami/ml/models/backend.py`. This is similar to how the get_status view returns a BackendStatusResponse. Should I remove both of these response schemas?

mihow: Ah yes, okay that makes sense! Thank you keep it

ami/ml/models/pipeline.py
mihow: Raise an Exception instead of returning an empty response with the message.

Also I have made a bunch of improvements to the job status updates in this PR #635, so things might change a little. But still, raising an Exception is the way to go here and should cause the job to fail.

mihow:
> @mihow I'm getting an error for this test in `ami-platform/ami/jobs/tests.py`:
> 
> ```
> def test_run_job_unauthenticated(self):
>         jobs_run_url = reverse_with_params("api:job-run", args=[self.job.pk])
>         self.client.force_authenticate(user=None)
>         resp = self.client.post(jobs_run_url)
>         self.assertEqual(resp.status_code, 401)
> ```
> 
> The console logs:
> 
> ```
> ======================================================================
> FAIL: test_run_job_unauthenticated (ami.jobs.tests.TestJobView.test_run_job_unauthenticated)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/app/ami/jobs/tests.py", line 176, in test_run_job_unauthenticated
>     self.assertEqual(resp.status_code, 401)
> AssertionError: 403 != 401
> 
> ----------------------------------------------------------------------
> ```
> 
> I think this error is occurring now after I modified the image processing to use the `choose_backend_for_pipeline()` function in `ami-platform/ami/ml/models/pipeline.py`. However, I'm not sure why exactly it's occurring. I think it might have to do with how the job status is determined?

Does this only happen when you have SessionAuthentication commented-in in the base.py settings?

ami/ml/models/pipeline.py
vanessavmac: image

I see, I've updated the error handling in the most [recent commit](https://github.com/RolnickLab/antenna/pull/632/commits/8f72fd333d6eeffb34b38e441dfcea7f6eb1698d). I found that I have to raise an Exception and return an empty response. If I raise just an Exception, the error message "No backends are online" shows up in the job log, but the job status is still SUCCESS.

vanessavmac:
> > @mihow I'm getting an error for this test in `ami-platform/ami/jobs/tests.py`:
> > ```
> > def test_run_job_unauthenticated(self):
> >         jobs_run_url = reverse_with_params("api:job-run", args=[self.job.pk])
> >         self.client.force_authenticate(user=None)
> >         resp = self.client.post(jobs_run_url)
> >         self.assertEqual(resp.status_code, 401)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The console logs:
> > ```
> > ======================================================================
> > FAIL: test_run_job_unauthenticated (ami.jobs.tests.TestJobView.test_run_job_unauthenticated)
> > ----------------------------------------------------------------------
> > Traceback (most recent call last):
> >   File "/app/ami/jobs/tests.py", line 176, in test_run_job_unauthenticated
> >     self.assertEqual(resp.status_code, 401)
> > AssertionError: 403 != 401
> > 
> > ----------------------------------------------------------------------
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think this error is occurring now after I modified the image processing to use the `choose_backend_for_pipeline()` function in `ami-platform/ami/ml/models/pipeline.py`. However, I'm not sure why exactly it's occurring. I think it might have to do with how the job status is determined?
> 
> Does this only happen when you have SessionAuthentication commented-in in the base.py settings?

Forgot that that was not commented out -- the tests pass when I comment out `"rest_framework.authentication.SessionAuthentication",`

mihow:
@vanessavmac I merged main into this branch and fixed several conflicts from other updates that have happened. I got into the groove and made several small changes & fixes as well. I think tomorrow we should do one more pass and test on a snapshot of the live database. Then hopefully merge!

mihow (approved):
Fantastic work! I made a few changes in #684 684 while integrating this big feature

PR #634, opened 2024-11-27
[Draft] Framework for exporting data

mihow:
Scalable method for triggering & downloading data from platform


mihow:
Closing this in favor of #725, which has most everything in this PR except for the data formats and the management command. Those should be added in a follow up to #725

Issue #607, opened 2024-11-01
Populate Antenna database with classifier scores from every class for each prediction


rhine3:
Hi all! Making a ticket here at the advice of @mihow -

To evaluate the performance of a classification pipeline on my dataset, I need to be able to download the entire vector of scores (for all classes) output by the ML classifier for each occurrence. In my previous download of my results from the API (in late September), I was only able to download the scores for the top-scoring class for each occurrence. 

Could you please add the ability to download all scores? Will it be accessible from a different API URL?

Thank you for your help!

Issue #310, opened 2023-11-09
Fixes & updates to jobs


mihow:
**Backend:**
- Update status_label from BE if there is an error
- Add nested params {} object on stages with name, key & value so that stage details are complete (when in place update FE)
- Add BE tests for failed jobs states
- Add BE tests for job logs
- Test deleting jobs more thoroughly. It appears to have strange behavior sometimes.
- Add permissions check to only show jobs for projects that the current user is a member of.

Several of these issues are likely related to concurrency (multiple processes updating the same job without refreshing the state first). Some things to try: an "increment" function at the DB level for integer fields, refresh_from_db before save, update specific fields instead of whole object, check timestamps of last_updated before saving.
- Job processing/processed counts go up and down (see #370)
- Job counts sometimes end with a negative count of processed images (see #404) 

These issues seem to happen periodically but are hard to reproduce. Add tests for
- [x] Getting "no Job exists" error again when processing a single image. And job is stuck as pending (consider creating Job model instance from a Task rather than the other way around)
- [ ] Job is reporting FAILURE status even through celery task is successful. But also it stopping immediately https://deploy-preview-379--ami-web.netlify.app/projects/46/jobs/618 - for both AMBER Panama collections tried. 

**Frontend:**
- Add "infinite progress bar" animation for jobs that will not know their progress percentage.
- Add pipeline chooser for the Process Now quick action button on the capture detail page (will also refresh & filter existing detections)

Only semi-related issues, but capturing here because we worked on them in the initial jobs PR #286
- Make score slider update num_detections in timeline items, in addition to actual detections on the current image
- Explore ways to make the score slider more responsive (front end updates first)

mihow:
@annavik I added a `params[]` list to the stages to replace what was in config before. Each param has a name, key and value. There is a `category` as well but I don't think you need to show that one.

<img width="480" alt="image" src="https://github.com/RolnickLab/ami-platform/assets/158175/4d3f3d9c-633b-4893-bc85-3531e22ad37c">

mihow:
Example of negative counts:


![Image](https://github.com/RolnickLab/ami-platform/assets/158175/7fbf866a-c098-4397-9dc1-ee4b0d5a560b)

mihow:
@vanessavmac Here I have documented some of thee issues we are seeing with jobs. These were hard to test before, but should be easier now with the dummy ML backend

PR #635, opened 2024-11-28
Refactor and enhancements to saving of ML pipeline results

mihow:
By addressing #607, we opened a can of bugs and missing features. So this led to a significant amount of refactoring and enhancements that are long overdue. 

Changes:
- Begin storing label lists / category maps that algorithms use. The model is called AlgorithmCategoryMap and maps the class index from the last model layer to the actual categories they represent. The categories have both a simple text label "Species name" as well as an object of metadata with an optional GBIF key, taxon rank, etc. These are used to create taxon entries in the Antenna database as well as show the top 3,5,10, N number of predictions from a model rather than just the top 1 which is saved as the occurrence's determination. 
- Moves job logs to their own field on the job model to reduce DB writes and overwriting the status field when writing logs.
- Add "task_type" field to algorithms so we can determine if its a classification model vs. a detection model and so on.
- Can handle results from multiple algorithms (moth/non-moth only)
- Faster & refactored saving of results (batch saving, split up functions, etc)

Benefits of this change:
- Can apply post-processing filters of species by region (filter and re-weight the scores)
- Can do genus & higher taxon roll-ups
- Can get the top N (top 3, 5, 10) suggestions instead of just the first
- Labels are now entities with GBIF key, synonyms, etc. instead of just a text label - So can better map to Taxon entities in the Antenna database
- Can use confidence score algorithms other than softmax on the Antenna side (e.g. temperature calibrated) 
- Allow seeing and agreeing with results from multiple models
- Uses best score from all (non-intermediate) algorithms to determine the species determination of an occurrence (not just the latest)
- Ready for detection tracking across frames!
- Can better send current results back to backend to determine reprocessing needs

While troubleshooting I should have fixed most of the issues reported in #310 

TODO:

- [x] More testing with live ML backend
- [x] Test migration of existing detections & classifications (they shouldn't have any data in the affected fields)
- [x] Add Algorithm & AlgorithmCategoryMap schemas to ML backends
- [x] Ensure the schemes in the example ML backend and live ML backend are the same
- [x] Update tests
- [x] Add new tests for category maps
- [x] Add tests for creation of algorithms & category maps (document how to update either)
- [x] Fix detection of already processed images (since we now return moth/non-moth classifications, all detections are seen as already processed by every pipeline)
- [x] Fix occurrences using moth/non-moth algorithm as the determination (likely because score is higher)
- [x] Fix algorithms being registered multiple times
- [x] Fix category maps from being registered multiple times
- [x] Raise errors that happen when saving results in subtasks
- [ ] Fix reprocessing - always is processing all images
- [x] Test on fresh export of data from production
- [ ] Verify that full category map + scores map to expected taxa results - Try showing top 3 (logits are empty again?)
- [x] Issues with cache not updating?? - this was due to errors happening in the subtasks for saving results
- [x] Fix tests


mihow:
Closing in favor of #684

Issue #236, opened 2023-09-08
Add general support for image thumbnails & media serving


mihow:
The implementation details for this are TBD. There are common strategies with Django, but we have a unique situation where the source images are stored in external storages that are unique to each Deployment (vs. a global storage backend for Django). One option is to use the Prism API-based service that Fieldguide.ai uses.

loppear:
Per slack discussion, propose that thumbnailing
- have two (for now) thumbnail sizes generated (widths 240px and 1024px), labeled "small" and "medium" - we're less worried about generating every specific size used by frontend than to generally improve performance/transfer, if they are subsequently scaled in browser still that's ok.
- is performed by an image proxy route, so that `captures` API responses aren't delayed by retrieval.
  - `GET /api/v2/captures/thumbnails/<source_image_id>/?label=<size_label>` will
    - check if a thumbnail file needs to generated
      - and retrieve the image from project storage
      - resize using PIL
      - write to media S3
      - save URL to DB
     - and 301 to the thumbnail media URL
- store these thumbnail paths as new fields on `SourceImage` (`thumbnail_url_small` and `thumbnail_url_medium`).
  - Alternatively, if we need to store per-thumbnail width & height, this could be a new table `SourceImageThumbnail ( id, source_image_id, label, url, width, height, size, last_modified )`
  - I assume these will never need presigned_urls as they're in antenna media S3.
- `SourceImageListSerializer` for e.g. `/captures/` to return new sub-field `thumbnails: {small: {url: '...'}, ...}`

annavik:
Thanks for summarizing! @loppear 

To have 2 sizes generated makes sense to me and I think 240px and 1024px will fit the frontend needs well.

Just a thought, I wonder if we have considered to not store the thumbnails permanently, just generate them on demand and serve them with a cache? If so, we could use the endpoint you suggested to access images of different sizes directly. First time would take some time, but at least we would move the load from the client. Sorry if you already discussed this.

loppear:
For size: do we need to specify height & width and possibly crop, or ok to let height vary by source aspect ratio? And does frontend want both dimensions in the API response?

[...]

PR #641, opened 2024-12-02
Update README for local backend setup

### LINKED ISSUE #640: Broken image links during local development
When running the backend locally, I have some problems with images uploaded from the interface. "Data source images" works fine. Probably some setup step I'm missing.

Example with capture from data source (works fine):
Screenshot 2024-12-02 at 17 38 40

Example with uploaded capture (broken):
Screenshot 2024-12-02 at 11 23 51

annavik:
Fixes #640 

I managed to fix issue above on my own @mihow, updated README with some missing information. Let me know if the updates are making sense! :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,11 @@ Antenna uses [Docker](https://docs.docker.com/get-docker/) & [Docker Compose](ht
 
 1) Install Docker for your host operating (Linux, macOS, Windows)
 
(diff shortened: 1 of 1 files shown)


mihow (approved):
Thanks for catching this. Turns out I already have the setting in my local env so I didn't notice! However the manually uploaded images should be using localhost as their hostname. But this can be addressed when we finally tackle #236 which will change how we host manually uploaded images.

PR #642, opened 2024-12-02
Fix user data form bug

annavik:
The save button was not triggering form submit when editing user info. The problem was the submit button was actually rendered outside the HTML `form` tag (not sure when this happened?). In this PR we fix this!

The problematic button:

Screenshot 2024-12-02 at 16 50 43


mihow (approved):
I noticed this but thought is was a fluke. Thanks for fixing! What kind of test do you think can catch this type of issue / regression?

annavik:
I would probably suggest manual regression testing here as a first step (define a regression test plan and make updating user info part of this). When regression testning is part of our release process, I think a good next step could be to look into tools for automation. Cypress, Selenium and Pupeteer are examples of some tools that can be used to simulate browser interactions.

What are your thoughts?

annavik:
I will merge this now, but let's talk more about regression tests if you think it's a good idea! :) I think it could make sense to have this part of our process once we start to do a bit more controlled releases?

PR #643, opened 2024-12-03
Add client side error handling to date filters

### LINKED ISSUE #619: Add some error handling to date filters
If we add a date in the URL that is not valid, some UI components will crash (however in a controlled way, but not very informative). Let's improve how this is handled!

annavik:
Fixes #619 

In this PR we add some client side error handling to date filters. This solution should be easy to extend to other filters as well, in cases we want to add client side validation. If a date filter value is not valid, we will now highlight the problem in the UI, instead of showing a generic error. This should make it easier for users to recover from the problem. Also we will not pass the invalid filter value with the backend request (making it possible to still see some data here).

Small note: I noticed the backend will respond with status code 500 if we include the invalid date filter with the request (in this case when fetching a list of occurrences). I think a 400 error would be more informative, but no biggie! Just a small improvement API users could benefit from.

**Before:**
Screenshot 2024-12-03 at 11 55 28

**After:**
Screenshot 2024-12-03 at 11 47 14


mihow (approved):
@annavik I fixed the issue where the backend was returning a 500 instead of a 400, thanks for catching that! Also the error/validation message is much more helpful now (with the name of the filter field in question). I added #646 to fix all other places where the same is occurring.

annavik:
> @annavik I fixed the issue where the backend was returning a 500 instead of a 400, thanks for catching that! Also the error/validation message is much more helpful now (with the name of the filter field in question). I added #646 to fix all other places where the same is occurring.

Thank you, that's great!

mihow (changes requested):
Can you add a validator for checking if the end date is before the start date? I added one on the backend, but now that returns a 400 error and the frontend doesn't know how to handle it (I know it's annoying because it's a multi-field validator). Perhaps we need both: validate end<start on the frontend, but also handle and display _any_ new 400 validation error coming from the backend. 

If we want to simplify in this PR, we can remove the end<start validator on the backend. Then just no results are returned from the database instead of throwing an error. We could add a general way to handle 400 responses & messages in #646

mihow:
@annavik any thoughts on my comment above? should we disable the backend validation for the end<start check?

annavik:
Hi @mihow ! Sorry for the delay, the changes got a bit bigger bigger than expected. I have updated this PR with some new stuff:
- First, I replaced our generic error state component. The new one is using components from Nova and also we display any backend errors with more details. Also, we will try to keep as much of the content visible, even in case of errors (for example the filtering sidebar). This will make it easier for users to recover from an error (for example by modifying filter settings). I think in future, we should also try to show errors in context (highlight the affected field and show error below it). I added a comment to #646. This PR has already expanded, so I suggest we make those changes as a separate task.
- For the date filtering, I added client side error handling as well. This will make it quicker for users to notice any mistakes. I tried to do this in a way to make it easy to add validation to any filter, by simply define a `validate` function (see `useFilters.ts`). I think a combination of frontend- and backend validation often is the best.

### Screenshots:
Client side validation for invalid date:
Screenshot 2024-12-16 at 10 13 41

Client side validation for invalid date range:
Screenshot 2024-12-16 at 10 13 27

Displaying result of server side validation (still in need if some improvements):
Screenshot 2024-12-16 at 10 12 29

Issue #616, opened 2024-11-13
If multiple labelers are involved, user verification status might be incorrectly updated


annavik:
Comment from Maxim: 
> Two of regional experts vetted the same occurrences on Antenna the first one who had vetted it, noticed they were sent back into their things to vet if another expert approved them with the same ID. [...] Meaning that the ''agree'' button who had turned to ''agreed'' is now back to ''agree''

Issue #615, opened 2024-11-13
Add filters UI to "identified by" filters


annavik:
Filters for "identification by" was added in #608 :
- identified_by filter occurrences that have been identified by a certain user id (supports multiple)
- not_identified_by filter occurrences that haven't been IDed by a user id (supports multiple)

We need some UI controls here. Most important option will be "Verified by me". We also want options for other users. To be figured out is what users to present in this list. In order for this solution to be scaleable, we might need some search option similar to the taxonomy filter.

PR #644, opened 2024-12-03
Add new filter "Verified by" and update logic for agreements

annavik:
Fixes both #615 and #616 

### Summary of changes:
- Make new filter "Verified by" available for logged in users. This filter currently has 2 options:
  - Verified by me
  - Not verified by me
- Update logic for when to show the disabled button "Agreed"
  - Before, we used a simplified logic in the list view. We had no access to all the identifications in this context, only details for the current determination. This ment we only showed "Agreed" if the determining identification was connected to the current user.
  - By exposing all identifications for occurrences in the list view, we could update the logic to better match what was requested in #616. Also we could streamline the code, to make sure we now use same logic for agreement status in the list view and in the detail view.
 - Some cleanup related updated above
   - Replace components `Filtering` and `AdvancedFiltering` with component `FilterSection`. This new component will better support custom filter sections.
   - Remove some unused code
   - Small style tweak for identification cards
   
###  Screenshots:
 Before:
Screenshot 2024-12-03 at 15 31 17
 
 After:
Screenshot 2024-12-03 at 15 44 24
 
Screenshot 2024-12-03 at 15 30 55


mihow (approved):
It works! Thanks so much for adding the FilterSections as well! I like  "More Filters" for now. And first backend change, wohoo! I just made one change to reduce the number of queries in the list view. Django makes it really easy to add data, but also that means more queries, usually.

mihow:
And that's super cool that the "Agreed" state is consistent across all buttons now. We can think more about what agreement means #616

PR #654, opened 2024-12-05
Tweak logout logic to clear token on all types of errors

### LINKED ISSUE #649: If the user intends to Logout, always clear the token in local storage
Currently the frontend waits for a response from the backend.

annavik:
Fixes #649

### DIFF
--- a/ui/src/data-services/hooks/auth/tests/useLogout.test.ts
+++ b/ui/src/data-services/hooks/auth/tests/useLogout.test.ts
@@ -48,7 +48,7 @@ describe('useLogout', () => {
     expect(removeQueriesSpy).toHaveBeenCalledTimes(1)
   })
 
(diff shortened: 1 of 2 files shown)


mihow (approved):

mihow:
The situation in question just happened to me locally! Getting an "Invalid Token" on all requests, and I couldn't click Logout.

But I just merged this new fix and it works!

image
image

annavik:
> The situation in question just happened to me locally! Getting an "Invalid Token" on all requests, and I couldn't click Logout.
> 
> But I just merged this new fix and it works!
> 
> image image

Good think it works now! However I do think our intention should be to logout the user when the token is not valid (instead of users having to logout themselves as a last resort). Can we align the status codes here? Either I can include more response codes in the frontend logic for logout, or if you could return 403 from backend when this happens?

PR #656, opened 2024-12-16
Skip detection counts

annavik:
We decided to focus on occurrences for now, to avoid confusions.

Hiding detection counts from:
- Capture table
- Occurrence table
- Occurrence details
- Session table
- Session details (both from playback and from summary view)
- Taxon details


mihow (commented):
Nice cleanup and simplification for now. Thanks!

mihow (approved):

Issue #650, opened 2024-12-04
Use job status to update UI for collections, captures & deployments


mihow:
- Ensure API returns job details per collection, capture, deployment
- Ensure API returns job ID after creating a job
- Show & link to jobs related to each item (bubble link)
- Use the status of running jobs to indicate done status in the UI context of collections, capture, deployment
"Populate"
"Process"
"Sync"

PR #657, opened 2024-12-16
Present latest job status in collections table and update populate button logic

annavik:
### Summary
In this PR we start look into the task "Use job status to update UI for collections, captures & deployments" (see https://github.com/RolnickLab/antenna/issues/650). To begin with, we include some frontend improvements related to jobs and collections (see screenshots for details). Also @mihow is adding some backend updates related to deployments and jobs as part of this PR. We will include frontend changes for remaining views in a follow up PR.

### Screenshots
Screenshot 2025-01-16 at 15 56 23

Screenshot 2025-01-16 at 15 51 15


annavik:
### Short summary after first session working on this:

✅ We have the backend info we need to link to a filtered job view from collections and captures. Frontend app is up to date here now.

⏳ We need to include a list of jobs in the deployments API response in order to do the same here.

⏳ If we can include job type in the API responses, we can adjust status labels to better reflect what is going on. For example show "Populating" instead of "Running".

🤔 The list of jobs that is included with the collections response has status labels that are not up to date (they are not the same as the ones I see in the full job table). Some cache or rewrite problem? I'm using local backend.

### Screenshot:
Screenshot 2024-12-16 at 15 33 28

mihow:
@annavik I added the jobs list to the Deployments API and added the job_type data to all job status objects. I think I will have fixed the job status caching issue in #635 

image

annavik:
Thanks a lot for adding this! I didn't had time to revisit this PR today, but hope to do it soon.

mihow:
Awesome thanks! Do you think we can merge the work you already did for the collections view first? And then implement the deployments <-> jobs features in a follow-up? @vanessavmac is adding a [new button ](ui/src/data-services/hooks/processing-services/usePopulateProcessingService.ts) similar to the Populate button in her PR #632 but the new button doesn't update based on the response status (if it fails it stays gray forever). I think we can use the code in this PR as an example.

annavik:
> Awesome thanks! Do you think we can merge the work you already did for the collections view first? And then implement the deployments <-> jobs features in a follow-up? @vanessavmac is adding a [new button ](ui/src/data-services/hooks/processing-services/usePopulateProcessingService.ts) similar to the Populate button in her PR #632 but the new button doesn't update based on the response status (if it fails it stays gray forever). I think we can use the code in this PR as an example.

Yes! I have synced this PR with main, pushed some remaining updates related to jobs and collections and also updated the PR description.

Let's merge this and handle remaining views in a follow up PR!

mihow (approved):
It works! This is a great improvement to our collection syncing, and a good template for action buttons in the project entities table, thank you!

PR #659, opened 2024-12-17
Make it easier to go to prev/next occurrence

### LINKED ISSUE #658: Make it possible to navigate to next/prev occurrence
From occurrence detail view:
- Using arrow buttons
- Using keyboard quick actions

Notes:
- If we reach start/end of current page, maybe we want to change page in background?
- It would be nice if we can help users not get lost on the current page. Perhaps change table row focus when navigating?

annavik:
Fixes #658 .

### Summary
In this PR we make it easier to go to prev/next occurrence from the occurrence detail view. Users can navigate using arrow buttons or arrow keys.

### Notes
- To simplify the logic and to avoid users losing context, I started with a solution where we will not change page in background. Instead we disable prev/next buttons when we reach start/end of a page. What do you think?
- When active occurrence is updated, we also update list view scroll (both for table view and gallery view).

### Screenshots
Screenshot 2024-12-17 at 16 50 07

### DIFF
--- a/ui/src/app.module.scss
+++ b/ui/src/app.module.scss
@@ -9,6 +9,7 @@
   height: 100dvh;
   overflow-x: auto;
   overflow-y: auto;
+  scroll-padding: 64px 0;
 }
 
 .main {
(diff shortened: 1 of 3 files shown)


mihow (approved):
I think the auto-scrolling is a great touch! And yes I agree about keeping it simple: not changing the page for the user at this time. But guess what, if you filter by "Not verified", then as you Agree, more occurrences are added to the page you are on! So you can Agree to infinity that way.

PR #661, opened 2024-12-20
Update filter logic to fix capture collection filtering

annavik:
Comment from @mihow :
> I just noticed that when viewing the captures in a collection, they are no longer being filtered by the collection, it's showing all captures in the whole project, no matter which collection you click on.

We recently introduced logic to only pass valid filters with backend calls. However, in this particular view the collections filter is hard coded (not possible to modify from URL or UI). Since hard coded filters are excluded from the validation logic, it wasn't marked as valid and therefore excluded from the backend request. By tweaking the logic to check for errors, rather than "marked as valid", we could fix the problem.

How the hard coded filter is defined in code (FIY I think this is the only hard coded filter situation):

Screenshot 2024-12-20 at 10 11 23

### DIFF
--- a/ui/src/components/filtering/filter-control.tsx
+++ b/ui/src/components/filtering/filter-control.tsx
@@ -68,7 +68,7 @@ export const FilterControl = ({
       </label>
       <div className="flex items-center justify-between gap-2">
         <FilterComponent
-          isValid={!filter.error}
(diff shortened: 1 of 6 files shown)


mihow (approved):
Thank you for the quick fix and explanation! Back in business!

PR #664, opened 2025-01-08
Handle occurrence detail view tab state as a search param

### LINKED ISSUE #662: Make the state of Occurrence detail tabs details sticky
When viewing the identifications tab of an occurrence, if you refresh the page or send a deep link to that view, the user needs to click the intended tab manually again (Details or Identifications).

Can we add that tab state to the URL? It would be helpful for our upcoming verification sprints, especially when sharing links with each other. 

Currently I am trying to fix a caching issue with identifications, and in this case it would also be super helpful if refresh brought me back to the same tab!

annavik:
Fixes #662 

To fix this, we use our already existing hook `useSelectedView`. We use this hook for the project summary tabs and also for the table/gallery state in main views. We update this took to handle a custom search param key (so we can use "tab" instead of "view" to avoid conflicts with the occurrence main view).

### DIFF
--- a/ui/src/pages/occurrence-details/occurrence-details.tsx
+++ b/ui/src/pages/occurrence-details/occurrence-details.tsx
@@ -35,8 +35,12 @@ export const TABS = {
 
 export const OccurrenceDetails = ({
   occurrence,
+  selectedTab,
+  setSelectedTab,
 }: {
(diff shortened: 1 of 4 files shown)


mihow (approved):
It works! Thank you!

PR #665, opened 2025-01-09
Show occurrence raw data

annavik:
In this PR we add a new tab "Raw" to the occurrence dialog. In this tab, users can inspect backend response details or simply open the API endpoint in a new tab:

Screenshot 2025-01-20 at 17 10 11

As part of this PR, we also replace all other code blocks with our new `CodeBlock` component that we just added to Nova.

Screenshot 2025-01-20 at 15 55 12


mihow (approved):
Yes yes yes! Thanks for including a visible URL to the API endpoint, a clickable icon and the embedded raw data. This is just what I needed for troubleshooting and I hope it encourages adoption by the more technical users. Allowing folks to utilize the platform without waiting for new features or data to be available in the UI. Wohoo! And the collapsible code block is perfect for the job logs. I was already experiencing getting stuck after expanding a long history of logs.

Issue #402, opened 2024-06-07
Add project<->user association


mihow:
- Add many2many relationship between projects & users - make a new migration & commit file
- Auto assign a user to a project when they create it
- Expose the projects a user belongs to in the user profile API endpoint (/me)
- Update Django admin for projects so we can manually add users to a project until public UI is ready
- Add tests (current user added to project when it is created via the API)
- See possible UI for "My projects" section on top of home page #518 

Follow-up tasks: 
- Use the upcoming project edit screen for managing project members
- Full permission checks for access to projects and related data #219 #354
- Add tests for permissions and access to different actions #639
- Roles can come later (annotator, data processor, admin, etc)

mihow:
See existing PR/branch #275

mihow:
**Step 1**
<img width="738" alt="Image" src="https://github.com/user-attachments/assets/5e3a5a63-f174-45d3-a29e-826a90c41802" />

**Step 2**
![Image](https://github.com/user-attachments/assets/81ac93ef-4ccd-453d-99cb-8e462acaa7ac)

**Step 3**
see other follow-up tasks

mohamedelabbas1996:
@mihow  may be we can close this issue now, because we have  implemented the tasks in the first part and we already have the follow-up tasks in this issue #624

mihow:
Closed by #275 woohoo!

PR #668, opened 2025-01-14
Renamed `project` filter query parameter to `project_id` for Project Summary entities

### LINKED ISSUE #660: Change param name from "project" to "project_id" for Entities
All of the "Entities" on the Project summary page pass a `project` parameter with the active project's ID. Let's update this to be consistent with other API calls and name the param `project_id` instead. Generally when setting a related object we are using `related_object_id` and then the read response has the object data under the field named `related_object`

So:
`Device.project_id` for setting a device's project
`Device.project` for reading a devices project data (as a nested object `project: {id: 1, name: "test project"}`)

image
https://antenna.insectai.org/projects/20

mohamedelabbas1996:
Closes #660

### DIFF
--- a/ami/main/api/serializers.py
+++ b/ami/main/api/serializers.py
@@ -503,7 +503,7 @@ class TaxonListSerializer(DefaultSerializer):
 
         # request = self.context.get("request")
         # project_id = request.query_params.get("project") if request else None
-        project_id = self.context["request"].query_params["project"]
+        project_id = self.context["request"].query_params["project_id"]
         classification_threshold = get_active_classification_threshold(self.context["request"])
 
         return obj.occurrence_images(
--- a/ami/main/api/views.py
+++ b/ami/main/api/views.py
@@ -10,6 +10,7 @@ from django.db.models.query import QuerySet
 from django.forms import BooleanField, CharField, IntegerField
 from django.utils import timezone
 from django_filters.rest_framework import DjangoFilterBackend
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import exceptions as api_exceptions
 from rest_framework import filters, serializers, status, viewsets
 from rest_framework.decorators import action
@@ -549,7 +550,7 @@ class SourceImageCollectionViewSet(DefaultViewSet):
     )
     serializer_class = SourceImageCollectionSerializer
 
-    filterset_fields = ["project", "method"]
+    filterset_fields = ["method"]
     ordering_fields = [
         "created_at",
         "updated_at",
@@ -562,11 +563,16 @@ class SourceImageCollectionViewSet(DefaultViewSet):
 
     def get_queryset(self) -> QuerySet:
         classification_threshold = get_active_classification_threshold(self.request)
-        queryset = (
-            super()
-            .get_queryset()
-            .with_occurrences_count(classification_threshold=classification_threshold)  # type: ignore
-            .with_taxa_count(classification_threshold=classification_threshold)
+        query_set: QuerySet = super().get_queryset()
+        project_id = self.request.query_params.get("project_id")
+        if project_id is not None:
+            project = Project.objects.filter(id=project_id).first()
+            if project:
+                query_set = query_set.filter(project=project)
+        queryset = query_set.with_occurrences_count(
+            classification_threshold=classification_threshold
+        ).with_taxa_count(  # type: ignore
+            classification_threshold=classification_threshold
         )
         return queryset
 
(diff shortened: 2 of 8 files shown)


ami/main/api/views.py
```diff
@@ -1358,13 +1386,35 @@ class SiteViewSet(DefaultViewSet):
...
+        project_id = self.request.query_params.get("project_id")
+        if project_id is not None:
+            project = Project.objects.filter(id=project_id).first()
+            if project:
+                query_set = query_set.filter(project=project)
+        return query_set
+
+    @extend_schema(
+        parameters=[
+            OpenApiParameter(
```
mihow, line 1407: Can you reuse this `OpenAPIParameter` instance? Or the kwargs that are used to configure it? `project_id_param` perhaps? It could go in the same `ami/utils/requests.py` as the `get_active_project_function`.

mihow (changes requested):
Fantastic! Great job finding the appropriate place for these changes and thank you for all of the tests. 

Do you feel up to making the changes on the frontend as well? Perhaps doing a pair session with Anna? If we merge it before the frontend is updated, then we have to make the change backwards compatible, which is cool but adds some mess to clean up later.

mohamedelabbas1996:
Done with the frontend part. @mihow  Should we identify all the places in the backend code where we should change the `project` parameter name?

ami/main/api/views.py
```diff
@@ -1049,7 +1068,10 @@ def filter_taxa_by_observed(self, queryset: QuerySet) -> tuple[QuerySet, bool]:
         """
 
         occurrence_id = self.request.query_params.get("occurrence")
-        project_id = self.request.query_params.get("project") or self.request.query_params.get("occurrences__project")
+        project_id = self.request.query_params.get("project") or self.request.query_params.get(
```
mihow, line 1071: This is interesting. Let's find any use of `occurrences__project` in the frontend and update it to use `project_id`. Then we can replace lines 1071:1074 with your new `get_active_project` function.

ami/main/tests.py
```diff
@@ -549,7 +549,7 @@ def setUp(self) -> None:
     def test_occurrences_for_project(self):
         # Test that occurrences are specific to each project
         for project in [self.project_one, self.project_two]:
-            response = self.client.get(f"/api/v2/occurrences/?project={project.pk}")
+            response = self.client.get(f"/api/v2/occurrences/?project_id={project.pk}")
```
mihow, line 552: Nice work tracking down all the existing uses!!!

ami/main/api/serializers.py
```diff
@@ -503,7 +503,7 @@ def get_occurrence_images(self, obj):
 
         # request = self.context.get("request")
         # project_id = request.query_params.get("project") if request else None
-        project_id = self.context["request"].query_params["project"]
+        project_id = self.context["request"].query_params["project_id"]
```
mihow, line 506: Thanks for finding this. When we optimize the Taxa view, I think we will be able to remove this function.

mihow (changes requested):
Great work! I like the reusable function. I think we can use it for the next PR when we start adding permission checks! I just requested one change about cleaning up the `occurrence__project` parameter.

ami/main/api/views.py
mohamedelabbas1996: @annavik Could you please help us with finding all uses of `occurrences__project` ?

mihow: @mohamedelabbas1996 I don't see any instances of `occurrences__project` as a URL param. I think it's safe to clean up these lines!

mohamedelabbas1996: @mihow  Done !

ami/main/api/views.py
```diff
@@ -1068,26 +1068,22 @@ def filter_taxa_by_observed(self, queryset: QuerySet) -> tuple[QuerySet, bool]:
...
             "occurrences__deployment"
         )
         event_id = self.request.query_params.get("event") or self.request.query_params.get("occurrences__event")
         collection_id = self.request.query_params.get("collection")
 
-        filter_active = any([occurrence_id, project_id, deployment_id, event_id, collection_id])
+        filter_active = any([occurrence_id, project, deployment_id, event_id, collection_id])
 
-        if not project_id:
+        if not project:
```
mihow, line 1080: @mohamedelabbas1996 in #402 I think we will require a project for all calls. Let's talk about this. We could require the project ID in the URL.

mihow (approved):

annavik:
Almost one year since @mohamedelabbas1996 first PR! 🌟

PR #690, opened 2025-01-22
fix: cannot use storage source for deployments

mihow:
## Summary
We started allowing many endpoints to filter by project. However in this case, if no `project_id` filter is specified then the `StorageSourceViewSet` was returning a 404 for the detail views. 

### List of Changes
Added a check to skip filtering if project_id is not specified, rather than returning only storage sources that have no projects (projects=None)


mohamedelabbas1996:
@mihow Nice catch spotting and fixing this issue.

mihow:
@mohamedelabbas1996 Thanks for noticing! btw the project filter is deployed to the live environment now.

PR #691, opened 2025-01-22
Disable snap behavior for sessions with no detections

### LINKED ISSUE #672: Disable "Snap to detections" if there are no detections in a session
If there are no captures that have detections in the entire session, than set "Snap to images with detections" to false. Otherwise a user cannot click to the next frame in a session after importing their images.

The detections_count or occurrences_count for an entire session can be found in the response of a couple endpoints. One is in the timeline endpoint, which is already being retrieved in the session detail view.
https://antenna.insectai.org/api/v2/events/5268/timeline/

Remember a "session" is the same thing as an "event". In the code we refer to sessions as events (which is also the term in other platforms like GBIF)

image
https://antenna.insectai.org/projects/18/sessions/5268?capture=19234172

image

annavik:
## Summary
For sessions with no detections (for example if not processed yet), the control "Snap to images with detections" is annoying for users since it blocks navigation. Before, we always had it enabled by default. In this PR, we only enable by default if the session has detections. Also we disable this control if the session has no detections, to avoid users ending up in this state.

### Related Issues
Fixes #672 

### How to Test the Changes
1. Go to a session **with no** detections and make sure the images can be navigated **without** the snap behavior.
2. Go to a session **with** detections. Make sure the images can be navigated **with** the snap behavior.

### Screenshots
After changes:
Screenshot 2025-01-22 at 10 20 41

### DIFF
--- a/ui/src/design-system/components/checkbox/checkbox.module.scss
+++ b/ui/src/design-system/components/checkbox/checkbox.module.scss
@@ -64,4 +64,8 @@
   &.neutral {
     color: $color-generic-white;
   }
+
+  &.disabled {
+    opacity: 0.5;
+  }
 }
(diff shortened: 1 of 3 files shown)


mihow (approved):
It works! flawless! Thank you, this is a big help for the new projects being added.

PR #692, opened 2025-01-22
Make it possible to inspect full occurrence details directly from the session details view

### LINKED ISSUE #651: New actions & tooltips for clearer linking between session detail & occurrence details

annavik:
## Summary
In this PR we try make the linking between the occurrence page and the session detail page a bit more clear. Also we make it possible for users to inspect full occurrence details directly from the session detail view, without losing context. This also makes it possible to verify predictions directly from the capture, instead from the occurrence list view.

### Related Issues
Fixes #651 

### Comments
As part of this work, we also bump the Nova version to get access to our new tooltip component. Our new tooltip is a bit more flexible, for example we can show more than just text as part of the content (like the info button).

When the dialog is active in session detail view, I decided to not reflect this as an URL param. This is to avoid name clashes and a bit of complexity. The concept "active occurrences" is already defined as something else in this view (active occurrence frames).

### Screenshots
Screenshot 2025-01-22 at 14 52 53
Screenshot 2025-01-22 at 14 53 02
Screenshot 2025-01-22 at 14 53 24
Screenshot 2025-01-22 at 14 53 33


mihow:
This is so freaking cool! thank you for fitting this in!

mihow (approved):
What a fantastic enhancement! It seems like there wasn't too much refactoring required in order to display the modal in multiple contexts. I hope this unlocks other possibilities as well. In any case, this greatly increases the usability of our existing features. I can't wait to use it.

mihow:
@annavik One follow up change: can you make the modal sticky in the session detail page, for linking to an occurrence detail modal? OR! add a permalink icon in the modal for linking to the occurrence detail view. Either way: when the user wants to share an occurrence detail view (especially with the identifications tab open) they should be able find how to do that, whether they started at the occurrence table or in the session detail view. 

Also btw it's awesome that the back button already works so well for these mixed states!

PR #705, opened 2025-01-28
Fix the list of online processing services per project

### LINKED ISSUE #700: Fix the list of online processing services per-project
## Summary

The count of processing services for a given pipeline includes processing services from other projects. And images can be processed by processing services outside of their own project. Add filters for processing services to restrict to the current project. 

Also "Pipelines registered" includes all pipelines under that processing service, not filtered by current project.

## Detailed Description

TBD

vanessavmac:
## Summary

Fix bugs associated with filtering processing services by project.

### List of Changes
* Fixed the value of "Processing Services Online" on the pipeline overview page. Previous, the count included all processing services of a pipeline across all projects. Filter to only include the activate project.
* Ensure images can only be processed by processing services in the active project. 

### Related Issues
Closes #700


vanessavmac:
@mihow From https://github.com/RolnickLab/antenna/issues/700: 

> Also "Pipelines registered" includes all pipelines under that processing service, not filtered by current project.

My last commit addresses this bug, but just to make sure I understood correctly: When a user adds pipelines by clicking the "Register Pipelines" button, it automatically adds those pipelines to that project. But if the pipeline is removed from the project (i.e. through Django admin) the user should not be able to see this pipeline as "registered" or in the ProcessingServiceDetails for this project? Not sure how in a real-life use case without the admin this would occur

I also updated the ProcessingServiceDetails so that only pipelines added to the project are visible

vanessavmac:
@mihow Backend tests failed but I think it's a cache issue? My most recent commit should fix the tests

ami/ml/views.py
```diff
@@ -75,8 +77,14 @@ class PipelineViewSet(DefaultViewSet):
     def get_queryset(self) -> QuerySet:
         query_set: QuerySet = super().get_queryset()
         project = get_active_project(self.request)
+        # If pipelines are filtered by project, also filter processing services by project
         if project:
-            query_set = query_set.filter(projects=project)
+            query_set = query_set.filter(projects=project).prefetch_related(
```
mihow, line 82: Nice work finding the custom filter for the Prefetch related query

ami/ml/views.py
```diff
@@ -115,8 +127,14 @@ class ProcessingServiceViewSet(DefaultViewSet):
     def get_queryset(self) -> QuerySet:
         query_set: QuerySet = super().get_queryset()
         project = get_active_project(self.request)
+
         if project:
-            query_set = query_set.filter(projects=project)
+            query_set = query_set.filter(projects=project).prefetch_related(
```
mihow, line 132: I don't think the processing service view needs to filter the related pipelines by project, but! it will once you add the enabled flag in #738 so I am okay to leave this here

mihow (approved):
Great work! There is a chance we can make the serializers & views for Pipelines and ProcessingServices more slim by querying less data. But they are fairly small tables used for configuration. If we are working with views & serializers for large tables like Captures, Detections, Occurrences or Taxa then we should profile the queries and limit the data being fetched & passed to the serializer. I'm just noting that here for the future!

PR #712, opened 2025-02-01
Support for order-level taxon ranks & improved search

mihow:
**Changes**
- Adds support for searching name aliases and english common names (currently helpful new top-level entries like "beetle" and "moth", etc)
- Add additional taxon ranks above Order. Arthropod (PHYLUM) is our new root taxon! Instead of Lepidoptera (ORDER)
- Improve speed and add filters in the Django admin

**Deployment notes**
Table of Arthropod taxa that will be imported to live:
https://docs.google.com/spreadsheets/d/1gjwq3hlXXtNMwEyR6rf2jUf3TTHdwJxowjenKL2xxZI/edit?gid=0#gid=0

Command to import them
```
docker compose  run --rm django python manage.py import_taxa --format csv https://docs.google.com/spreadsheets/d/e/2PACX-1vRzHE3kjjIc8iWV1Be4hlTUBU4M1oD7R5h3imEZcsO5C2MWLlk40FolNfkAZiQetyhxm7ya6DDPd9Ye/pub\?gid\=0\&single\=true\&output\=csv --list upper-ranks
```

**Screenshots**
image
image

**Follow-ups**

Standardize the Taxon search / lookup. Add a reusable QuerySet method for finding a taxon. Use this when saving results  from the ML backend. See similar query in ml.models.algorithm.AlgorithmCategoryMap
This may be applicable to add in #719 if we are saving the same results with different Taxon (e.g. creating a new "Non-Moth" Taxon every time)


mihow:
@annavik the search results can contain all parents much more efficiently now, if want to display more than one in the tree. I'm not sure if it's a good idea or will add too much clutter
image

mihow:
Two action buttons that are not yet checking for permissions in the UI

image
image

The Register Pipelines does get blocked by the backend permissions check, but the Star/unstar does not.

PR #715, opened 2025-02-04
Improving status indicators for processing services and pipelines

### LINKED ISSUE #680: Improve test_process pipeline view and status indicators
Improve the processing image experience so that users know what services/pipelines are available.

`ami/ml/views.py` Rather than select a random image to test the pipeline processing (`test_process`), filter images by projects user has access to. 

Also add status indicators to the processing services and pipelines overview.

vanessavmac:
## Summary

Add status indicators to the processing services overview based on value of "last_checked".

Screenshot 2025-02-03 at 9 45 41 PM

Disable a pipeline in the capture details processing page if no processing services are available.

Screenshot 2025-02-03 at 9 47 14 PM

### Related Issues
Closes #680 

## Detailed Description

This PR is affected by #705 (which corrects the project filters on projects/pipelines).


vanessavmac:
Hey @annavik ! This is the PR with the status indicators we discussed adding to the processing services. The drop down on the process images page is not styled very nicely at the moment due to the dark background; I'd appreciate your help in improving the styling. Also let me know if I should make any changes to the UI edits I made. Thanks!

annavik:
Hi @vanessavmac ! Looks like a great start, I'll have a look at this in a day or two!

annavik:
Hey Vanessa! I have tweaked the job UI controls, to better fit this compact sidebar and also the dark background. Let me know what you think!

Screenshot 2025-02-11 at 11 46 53

ui/src/data-services/models/pipeline.ts
```diff
@@ -82,6 +83,22 @@ export class Pipeline {
     })
   }
 
+  get hasOnlineProcessingService(): {
```
annavik, line 86: Nit: since this getter is returning an object (not a boolean value), I would suggest renaming it. Maybe just `currentProcessingService`?

mihow:
Could we make move "last checked" value from it's own column to a tooltip on the status icon? Only if it's easy! But would save some side-scrolling to the actions & edit button. 

Otherwise is this one ready for review?

annavik:
> Could we make move "last checked" value from it's own column to a tooltip on the status icon? Only if it's easy! But would save some side-scrolling to the actions & edit button.
> 
> Otherwise is this one ready for review?

Good idea! Actually, the component `StatusTableCell` has a prop called `details` already. If the defined, the component will render this prop value as a tooltip. @vanessavmac do you want to give this change a try?

vanessavmac:
I will start making the changes requested today! Thanks for improving the UI controls @annavik they look great! Once I make the changes, I'll un-draft this PR.

mihow (approved):
I love this feature, thank you! It makes it much more clear that our ML backends are running on separate servers and may be connected or disconnected from Antenna. It also prevents them from selecting an offline service and being surprised when the image fails.

PR #719, opened 2025-02-06
Fix reprocessing of existing results

### LINKED ISSUE #718: Regression: Fix unnecessary reprocessing of all images
## Summary

When a new job is created, all images are determined to need reprocessing, instead of only those remaining in the collection. This is a problem because if a job fails, it has to start from the beginning. 

This seems to have been introduced in #684 

## How to Reproduce the Bug

- Create a collection with a few images (10 random)
- Process the collection with a pipeline
- Process them again (hit Retry)
- Look at the "filter count" in the job logs, it will be 10 out of 10 again.

mihow:
Fixes #718 

The main culprit was the pipelines in our registry have algorithms registered that are no longer being used by the actual pipeline in the ML Backend. So new images never get classifications from these old algorithms, but the images_filter always thinks they need classifications from those algorithms! And sends them back for processing. 

This PR however adds some cleanup, logging & tests to help clarify what's going on. The main fix is a data entry clean up in the live DB.


mihow:
The remaining image count works when using the same detector and binary classifier but not the species classifier

Issue #298, opened 2023-10-31
Support for exporting


mihow:
- Filter a list view
- Export results in various formats
- Creates job for export
- Notification when ready to download

PR #725, opened 2025-02-17
Framework for exporting data

### LINKED ISSUE #720: Exporting: Support for trigger background exports
Quick description

- User tiggers an export for the Occurrences list view based on current filters https://antenna.insectai.org/projects/18/occurrences (an other list views)
- Choose from various formats (CSV, json, sql)
- New model for tracking export history
- Background job for generating export
- Show user links to exports that are ready
- Start with polling the exports endpoint
- Look into websockets & notifications to user
- Consider emailing the user with link 

When we have an Occurrence Collection/Dataset type, then the user can create a dataset/collection from the Occurrence filters first, then export the Dataset/collection. So exports could actually be a method on Occurrence Datasets.

Examples of good asynchronous exports with notifications
- GBIF export
- Label Studio

mohamedelabbas1996:
## Summary
This PR implements a framework for defining different export formats and exporting data in the background. It also defines the API for creating & retrieving the history of a project's exports. Users can trigger exports via the API or the Admin Page with optional filtering parameters and track job progress asynchronously.

## Related Issues
Closes #720  

## Detailed Description
This PR enables users to export filtered **Occurrence** data in **CSV** or **JSON** format via the API. The export process runs asynchronously in the background.

### Triggering an Export
Users can initiate an export by making a **POST** request to:

```
POST http://localhost:8000/api/v2/exports/`
```

#### Request Body 
The request body should specify the export format and optional filters. Available formats:  
- **project** → Occurrences Project. 
- **occurrences_simple_csv** → Exports simplified tabular data in CSV format.  
- **occurrences_simple_json** → Exports structured JSON matching the API response.  
- **filters** → Currently supports filtering by collection (optional).  

#### **Request Example (CSV Export with Collection Filter)**  
```http
POST http://localhost:8000/api/v2/exports/
Content-Type: application/json


        {   
            "project":105,

            "format": "occurrences_simple_csv",
            "filters":{
                "collection":104
            }
        
             
         }

```

#### Request Example (JSON Export)
```http
POST http://localhost:8000/api/v2/exports/
Content-Type: application/json

        {   
            "project":105,

            "format": "occurrences_simple_json",
            "filters":{
                "collection":104
            }
        
             
         }
```

### Checking Export Status
Once an export job is triggered, users can check the job status using:

```
GET http://localhost:8000/api/v2/exports/<export_id>
```

### Fetching All Exports
To retrieve all exports:

```
GET http://localhost:8000/api/v2/exports/
```

#### Filtering Exports by Project ID
```
GET http://localhost:8000/api/v2/exports/?project_id=<project_id>
```

## Screenshots
image
image

## How to Test the Changes
1. Trigger an export by making a **POST** request to `/api/v2/exports/` with valid `project_id` and request body.  
2. Track export progress using the **GET** request to `/api/v2/exports/<export_id>` with the returned `export_id`.  
3. Once completed, download the exported file from the provided `file_url`.  
4. Test with different values for format and  the collection filter to ensure correct filtering and export behavior.  
5. Verify listing exports using `/api/v2/exports/` and filtering with `project_id`.


annavik:
Woho! Sounds very promising, looking forward testing this out later 🤩 Some early thoughts on the user flow and related backend extensions:
- I think it would make sense if the backend could list all exports tasks for a specific project. I'm thinking maybe users will leave the app to come back later, then it would be helpful to have a way to know about both ongoing export tasks. This would also make it possible to share export results in a team.
- If we should make it possible to list tasks, then we should probably also make it possible to delete and update them. Similar to jobs? I suggest we handle this in a follow up PR though, and keep the first version as readonly.
- Do you think we could get some progress info in the tasks response, not just the status? It would be nice to present some progress bar or similar for users... It doesn't have to be exact, but some approximate progress between 0-1?
- My guess would be users would are often interested in exporting all pages, is this export feature providing that option? Sorry, I couldn't tell from quickly checking the code changes.
- I see the code changes includes functionality for sending an email when a task is done? For me, that is more a "nice to have" feature. I think our focus should be to first integrate this on the web platform! :)

mohamedelabbas1996:
> * Do you think we could get some progress info in the tasks response, not just the status? It would be nice to present some progress bar or similar for users... It doesn't have to be exact, but some approximate progress between 0-1?

I think it's possible but not sure how to implement it may be write the progress value in-memory in redis? @mihow what do you think ?

mohamedelabbas1996:
> * I see the code changes includes functionality for sending emails when a task is done? For me, that is more a "nice to have" feature. I think our focus should be to first integrate this on the web platform!

For the email feature, I think we can add it is just a small change and  we already have email service integration (not working locally for me though)

mohamedelabbas1996:
> * think it would make sense if the backend could list all export tasks for a specific project. I'm thinking maybe users will leave the app to come back later, then it would be helpful if there was a way to know about both ongoing export tasks. This would also make it possible to share export results in a team.
> 
> * If we should make it possible to list tasks, then we should probably also make it possible to delete and update them. Similar to jobs? I suggest we handle this in a follow up PR though and keep the first version readonly.

Yes, i think it would be easy to add export tasks  CRUD since we already have the ExportHistory model in place

mohamedelabbas1996:
> * From my experience, users are often interested in joining results from multiple or all pages, is this export feature providing that option? Sorry, I couldn't tell from quickly checking the code changes.

Thanks for bringing this up. I 've just pushed changes to disable pagination for the `export` action.

mihow:
@Mohamed Elabbas the implementation for non-ML job types isn't super clear right now, but it is flexible! I am curious if you can create a new job type for exports and so we can use that to view status, logs & errors. The download link can even be shown in the logs for the time being. See the Data Sync Job type and Collection Populate job types as examples.

mihow:
The main formats for now should be
1) a JSON export with nested data that mostly matches the current API list views 
2) simplified & flattened CSV of each data type (see #634)
3) a Darwin Core Archive zip of flattened TSVs with data about Occurrences, Sessions and Taxa

This PR can offer the simplest option. It should focus on a scalable background task & API endpoints.

ami/main/api/views.py
```diff
@@ -1040,6 +1040,16 @@ def get_queryset(self) -> QuerySet:
     def list(self, request, *args, **kwargs):
         return super().list(request, *args, **kwargs)
 
+    def paginate_queryset(self, queryset):
```
mihow, line 1043: It may be more scalable to _keep_ pagination, but automatically loop through all the pages. Rather than triggering a single huge database query. Or another way to break it apart? I can give you a large DB snapshot to test on.

mohamedelabbas1996: I would appreciate it very much if I 've access to this DB snapshot.

mihow:
@mohamedelabbas1996 Will you make the first export format be JSON from the API list view for occurrences? That's what people are currently exporting now (@annavik @rhine3). They are using external python scripts to loop over the paginated API views.

ami/main/api/views.py
```diff
@@ -1032,6 +1034,78 @@ def get_queryset(self) -> QuerySet:
...
+        """
+        Override pagination to skip pagination for 'export' action.
+        """
+
+        if self.action == "export":
```
mihow, line 1042: I'm tempted to create a new set of views in the exports app (exports/views.py). Many of the exports will contain multiple models (occurrences & events). It may be harder to reuse filters from the Occurrences view, but they could be moved to QuerySet methods. Also the custom export views can be simpler, since they will all be read-only list views (no CRUD needed). Thoughts?

ami/exports/serializers.py
mihow, line 1: This serializers file is feeling a little redundant now that I see it! It's almost the same as what's in main/api/serializers.py. I think the reason I created the "DefaultExportSerializer" is because when exporting from a management command I didn't have a request object and i couldn't use the existing serializers. But if we are triggering this from the API then we do have a request! Can we try and use the existing serializers for the full json export? I still see a use-case for new serializers when we make dedicated export formats, those will be much simpler.  Sorry I couldn't review this earlier!

ami/main/api/views.py
mohamedelabbas1996: From a code modularity and separation of concerns perspective, I agree that having separate views for exports makes the code cleaner and less cluttered. That approach also simplifies maintenance since export-related logic won't be mixed into the main Occurrence views.

As for filters, I wanted to bring up a related issue: how do we efficiently pass a filtered queryset to a job object? can we consider adding a queryset parameter directly to the job object? Would love to hear your thoughts on this.

ami/exports/serializers.py
mohamedelabbas1996: Yes, that was my first thought as well—I initially tried reusing the Occurrence API serializer, but I ran into a circular import issue. Because of that, I had to create a separate serializer for exports.

I think we might need to restructure the API serializers to make them reusable for exports. The circular dependency happens because the API serializers import the Job model, while the export logic imports the API serializers, and then Job models import the export logic

mihow:
One idea for saving filtered occurrence querysets and passing them to the background jobs:
```
OccurrenceDataset
occurrences = models.ManyToManyField(Occurrence)
1 - 3
2 - 4
2 - 4
occurrences = models.ArrayField(OccurrenceIds) 

OccurrenceDataset.occurrences[]

filtered_occurrences = Occurrence.filter(...).exclude(...)
OccurrenceDataset.occurrences.set(filtered_occurrences)
```

ami/exports/base.py
```diff
@@ -0,0 +1,254 @@
...
+
+    logger.info(f"create_dwc_archive: {archive_path}")
+    return archive_path
+
+
+def export_occurrences_to_json(occurrences: models.QuerySet, job):
+    """
+    Export occurrences to a JSON file
+    """
+
```
mihow, line 169: Try importing the OccurrenceListSerializer within this function

ami/exports/serializers.py
mihow, line 1: Remove the occurrence json export serializers that are almost copies of OccurrenceListSerializer, we will fix the circular import error and pass a request object to the current web API's serializer

ami/jobs/models.py
mihow, line 1: Add a new field to the Job model: `params` JSONField and pass a dict to that with the format, filters, etc

ami/exports/views.py
```diff
@@ -0,0 +1,70 @@
...
+    API endpoint for exporting occurrences.
+    """
+
+    queryset = Occurrence.objects.all()
+
+    def list(self, request, *args, **kwargs):
+        format_type = request.query_params.get("file_format", "json").lower()  # Default to JSON
+
+        # Validate format
+        valid_formats = ["csv", "json", "darwin_core"]
```
mihow, line 21: Let's make the formats named for the use case rather than the file format / extension. Something like [`simple_csv`, `simple_json`, `darwin_core_archive`, `web_api_json`, `species_list_csv`]

For this PR, just `['simple_csv', 'simple_json']`

image

ami/jobs/models.py
```diff
@@ -596,6 +599,77 @@ def run(cls, job: "Job"):
...
+
+        job.logger.info(f"Export completed: {file_path}")
+
+        #  Upload File to MinIO Storage
+        file_name = f"exports/{job.task_id}.{export_format}"
+        with open(file_path, "rb") as f:
+            default_storage.save(file_name, f)
+
+        file_url = default_storage.url(file_name)
+        job.logger.info(f"File uploaded to MinIO: {file_url}")
```
mihow, line 657: change MinIO to "project storage"

requirements/base.txt
```diff
@@ -41,6 +41,10 @@ djoser==2.2.0
 django-sendgrid-v5==1.2.2
 django-anymail[sendgrid]==10.0  # https://github.com/anymail/django-anymail
 
+# Darwinc Core Format
```
mihow, line 44: Let's save these new dependencies for the next PR

ami/exports/views.py
mihow: actually! format options are a flat list for all types of exports (different models, use cases, file formats, etc.)
```
[
'occurrences_simple_csv`,
'occurrences_simple_json`,
'project_species_list',
'darwin_core_archive_zip',
]
```

```
job.params = {
'format': 'occurrences_simple_csv',
'# filters': {'determination__id__in': [1,3,4], 'first_appearance': 2024-01-01}
'filters': {'start_date': 2024-02-01},
'request_obj_data': ...,
}
```

```
run() {
export_format = job.params.get('format')
export_function = f'export_{export_format}'
export_function(filters=job.params.get("filters"))
}

def export_occurrences_simple(filters={}):
    first_appearrance_timestamp = filters.get("start_date")
    occurrences = Occurrence.filter(project=project_id).filter(first_apperance_timestamp=...)

def export_darwin_core_archive():
    events = Event.objects.filter(start_date=)
    event.occurrences
```

ami/exports/views.py
```diff
@@ -0,0 +1,70 @@
+from rest_framework.decorators import action
+from rest_framework.response import Response
+from rest_framework.viewsets import ReadOnlyModelViewSet
+
+from ami.jobs.models import DataExportJob, Job, JobState
+from ami.main.models import Occurrence
+from ami.utils.requests import get_active_project
+
+
+class OccurrenceExportViewSet(ReadOnlyModelViewSet):
```
mihow, line 10: This can likely just become `ExportViewSet` since occurrences will be just one of the "formats" now

ami/exports/views.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class OccurrenceExportViewSet(ReadOnlyModelViewSet, ProjectMixin):
+    """
+    API endpoint for exporting occurrences.
+    """
+
+    queryset = Occurrence.objects.all()
+
+    def list(self, request, *args, **kwargs):
```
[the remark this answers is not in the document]
mohamedelabbas1996: I've updated the `ExportViewSet` to use the `DataExport` model instead of Occurrence. Now, when a POST request is sent, a new `DataExport` entry is created, and an export job is triggered. The `DataExport` Id is then used to track the export progress and retrieve the file URL via the detail view.

ami/exports/registry.py
```diff
@@ -0,0 +1,29 @@
...
+        """Retrieve an exporter class based on format type."""
+        return cls._registry.get(format_type, None)
+
+    @classmethod
+    def get_supported_formats(cls):
+        """Return a list of registered formats."""
+        return list(cls._registry.keys())
+
+
+ExportRegistry.register("occurrences_simple_json")(format_types.JSONExporter)
```
mihow, line 28: Cool approach!

ami/exports/serializers.py
```diff
@@ -0,0 +1,58 @@
...
+        def get_permissions(self, instance_data):
+            return instance_data
+
+        def to_representation(self, instance):
+            return serializers.HyperlinkedModelSerializer.to_representation(self, instance)
+
+    return OccurrenceExportSerializer
+
+
+class OccurrenceTabularSerializer(serializers.ModelSerializer):
```
mihow, line 27: What about a directory / sub module for export formats? Wherein there is a file for each format. Then each format file contains the Exporter, the Serializer, and the registry call for a single type of export, or a group of related formats that have the same query, etc.  `exports/formats/occurrences_simple.py`

ami/exports/views.py
```diff
@@ -0,0 +1,103 @@
...
+from rest_framework.response import Response
+from rest_framework.viewsets import ModelViewSet
+
+from ami.base.views import ProjectMixin
+from ami.exports.registry import ExportRegistry
+from ami.jobs.models import DataExport, DataExportJob, Job, JobState
+from ami.jobs.serializers import DataExportSerializer
+
+
+class ExportViewSet(ModelViewSet, ProjectMixin):
```
mihow, line 11: I like how this ViewSet turned out! clean and easy to follow

ami/exports/views.py
```diff
@@ -0,0 +1,103 @@
...
+    queryset = DataExport.objects.all()
+    serializer_class = DataExportSerializer  # Make sure to create a serializer for DataExport
+
+    def create(self, request, *args, **kwargs):
+        """
+        Create a new DataExport entry and trigger the export job.
+        """
+
+        # Get export format and project from request params
+        format_type = request.data.get("format").lower()
```
mihow, line 25: In the docs you mention to pass the format in the request body and the filter params in the URL. But I believe all of them can be passed as either, correct?

mihow (changes requested):
I did an initial review of just the code. I will test asap. This is looking great.

ami/exports/views.py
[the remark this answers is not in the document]
mihow: On the money!

ami/exports/views.py
```diff
@@ -0,0 +1,103 @@
...
+            user=request.user,
+            job=job,
+            format=format_type,
+            status=JobState.PENDING,
+        )
+
+        # Enqueue the job
+        job.enqueue()
+
+        return Response(
```
mihow, line 59: In the future let's update these custom JSON responses to use a schema defined by a pydantic model or Python dataclass and get them to show up in the API docs. We have only done this for the FastAPI server, not in the Django Rest API views, so I don't have a working example. But I am guessing you may have wanted something less freeform when you wrote this. We probably have half a dozen custom JSON responses that can use the same approach.

https://github.com/RolnickLab/antenna/blob/9d737c2e22a16356395e805b0875bcce75df4fcf/processing_services/example/api/api.py#L97

ami/exports/serializers.py
mihow: @mohamedelabbas1996 I left a number of comments, but this is the only significant one. I am happy to jump on a call about this proposal.

ami/exports/format_types.py
```diff
@@ -0,0 +1,75 @@
...
+    file_format = "csv"
+
+    serializer_class = OccurrenceTabularSerializer
+
+    def get_queryset(self):
+        return self.job.project.occurrences.all().with_timestamps().with_detections_count()
+
+    def export(self):
+        """Exports occurrences to CSV format."""
+        temp_file = tempfile.NamedTemporaryFile(delete=False, suffix=".csv", mode="w", newline="", encoding="utf-8")
```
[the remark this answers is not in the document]
mihow: I see now that you are writing to a tempfile first, and then transferring to the object store. Sorry I missed that!

ami/exports/views.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+        # Extract filters from request
+        filters = request.query_params.dict()
+
+        project = self.get_active_project()
+
+        if not project:
+            return Response({"error": "Project ID not provided or invalid"}, status=400)
+
+        # Create a Job for export
```
mihow, line 40: I suggest moving this to a function on the DataExport model (run_export creates the job)

ami/exports/views.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def retrieve(self, request, *args, **kwargs):
+        """
+        Retrieve a DataExport record and return job status.
+        """
+        data_export = get_object_or_404(DataExport, pk=kwargs["pk"])
+        job = data_export.job  # Get associated job
+
+        response_data = {
+            "export_id": data_export.pk,
```
mihow, line 72: The detail response should be able to use a basic `DataExportSerializer` with a `NestedJobSerializer` for the status, instead of a custom response.

ami/exports/serializers.py
mihow: We decided to wait on this refactor until #298

mihow:
In summary! From our conversation on Mar 12. Here are the most important things that affect how we use the endpoints from the frontend:

- Add project relationship to DataExport model
- Add get_queryset method to DataExportView (filter by active project, select_related('job'))
- Add nested Job to DataExportSerializer
- Remove custom retrieve & list functions

Additionally, this should be an easy modification:
- In the create view, just create the DataExport object first, then call `start_job` from a new method in the DataExport model which actually creates & starts the job. The create method could either wait for the job to be created and started, then return a response, or it can return right away without a job or task ready.

annavik:
### Notes from chat Mar 18:
- We decided to **skip** the current filtering options (date start, date end, taxon)
- We decided to **add** a new filtering option for collection
- We decided to simplify the generated job name, to be more similar to other job names:
  - Current name: `#1 Test Project 51279d37 Export Occurrences (occurrences_simple_json)`
  - Suggestion for new name: `Export occurrences` or `Export occurrences for collection [id]`
- We decided to assign all export jobs a **shared** type, so we can filter jobs based on this type. At the moment, seems like each export format has its own job type.
- We decided to have a look at job stages information, to see if we can return more useful information here. At the moment there is only one stage with repetitive details.
- Labels for job stages should use "Sentence case", not "snake_case" or "Capital Case".
- We decided to return basic information about the export object as part of the job response. In this way we can link from the job to the export, which is not possible at the moment.
- We decided to let API users specify all export settings in the request **body**, to streamline. At the moment some settings are specified as query params.

### Screenshot to clarify some details:
Screenshot 2025-03-19 at 11 45 57

annavik:
I tried to visualize the API data to see where we might have some gaps. I think works well! 🥳

Some comments/questions:
- I will convert the "format" value to a readable string and call it "type", what do you think?
- What do you think about the inline job progress?
- I think export objects should include `created_at` and `updated_at` similar to other entities
- It seems like the list endpoint is not supporting sorting. I think sorting based on `id` and `format` (and any timestamps we might add) would be helpful.
- Do you think we should let users give exports a name and description similar to other entities, or is it not necessary? Maybe if we present timestamp and any related collection, that will be enough for users to identify?
- The job information we embed with the exports list response is quite detailed, however I'm missing the job name. Could be helpful to have if we want to use the real job name for any links (I'm just using the ID for now).

Screenshot 2025-03-19 at 11 15 52

annavik:
When I tried some new export jobs today, some jobs ended up in this state (progress is 25% although job has completed and also I see job stages that are not related to exports). This is for my super slimmed local DB :)

Maybe this work in progress, but just FIY!

Screenshot 2025-03-19 at 12 31 39

mohamedelabbas1996:
> * I will convert the `format` value to a readable string and call it "type", what do you think?

Allowing users to rename exported files is a great idea, but I think we should keep it simple for this PR. Also, converting the format to a readable string is a great improvement!

mohamedelabbas1996:
> I had a closer look at the export formats as well, some comments:
> 
> **JSON export**
> 
>     * Current JSON seems to not be valid, it contains multiple root elements. Here is an [online tool](https://jsonformatter.curiousconcept.com/) you can use to quickly test the output :) However I wonder if we can generate the JSON using some method that will ensure the output is always valid?
> 
>     * The list of occurrences is not flat (we have multiple arrays). If we flatten this, I think the problem above will get solved as well.
> 
>     * I went on and verified some predictions, then the JSON export job failed with message `Job #10 "#1 Test Project 51279d37 Export Occurrences (occurrences_simple_json)" failed: Object of type set is not JSON serializable`. I did not have the same problem for CSV export.
> 
>     * It felt a bit strange to me to see `user_permissions` fields in the export, can we exclude?

Thanks for pointing this out and sharing the json validation tool!  I’ve fixed the serialization error and verified that the JSON format is now valid. As for the user_permissions fields, removing them is a bit challenging at the moment since we are reusing the API serializers.
image

mohamedelabbas1996:
> * Do you think we should include some verification information here, it's currently not possible to see if an occurrence has been verified or not (except we can see the score is 1.0).

I've add the verification information to the CSV format.

mohamedelabbas1996:
> ### Notes from chat Mar 18:
> 
>     * We decided to **skip** the current filtering options (date start, date end, taxon)
> 
>     * We decided to **add** a new filtering option for collection
> 
>     * We decided to simplify the generated job name, to be more similar to other job names:
>       
>       * Current name: `#1 Test Project 51279d37 Export Occurrences (occurrences_simple_json)`
>       * Suggestion for new name: `Export occurrences` or `Export occurrences for collection [id]` (depending on settings)
> 
>     * We decided to assign all export jobs a **shared** type, so we can filter jobs based on this type. At the moment, seems like each export format has its own job type.
> 
>     * We decided to have a look at job stages information, to see if we can return more useful information here. At the moment there is only one stage with a bit repetitive details.
> 
>     * Labels for job stages should use "Sentence case", not "snake_case" or "Capital Case".
> 
>     * We decided to return basic information about the export object as part of the job response. In this way we can link from the job to upcoming export details, which is not possible at the moment.
> 
>     * We decided to let API users specify all export settings in the request **body**, to streamline. At the moment some settings are specified as query params.
> 
>     * We decided if a user will delete an export object, we should delete related files.
> 
>     * We decided if a user will retry a job, we will **not** generate a new export object. Instead we will **update** the current one and override any files.
> 
> 
> ### Screenshot to clarify some details:
> Screenshot 2025-03-19 at 11 45 57

image
image
image

annavik:
> > When I tried some new export jobs today, some jobs ended up in this state (progress is 25% although job has completed and also I see job stages that are not related to exports). This is for my super slimmed local DB :)
> > Maybe this is work in progress, but just FIY! I know Michael has been working on (and fixed?) some related issues, could be more a generic job problem.
> 
> This seems like a generic job issue, or it could be related to the Celery worker restarting after code changes. Could you provide more details on how to reproduce this issue?

I used a newly reset DB (the one with the simple test project) and then I triggered an export. I saw this for both formats. The tricky thing with jobs I guess sometimes hard to reproduce consistently?

The progress problem I have seen before for other jobs, but seeing stages not related to the current job type was new to me.

annavik:
> > * I will convert the `format` value to a readable string and call it "type", what do you think?
> 
> Allowing users to rename exported files is a great idea, but I think we should keep it simple for this PR. Also, converting the format to a readable string is a great improvement!

Sorry, I didn't mean we should let users rename the file, I ment let users assign name and description to export objects, similar to other entities. Just to clarify! Maybe this is not necessary though.

annavik:
Hey @mohamedelabbas1996 ! Thanks for all the fixes, I have tested again now! Overall it's working great.

### Comments:
- Sorting seem to not be working still, or at least it's not working similar to other API endpoints. For example, when we sort by "created_at" in descending order, this is how I would like to call the list endpoint:
http://localhost:4000/api/v2/exports/?project_id=1&ordering=-created_at&limit=20&offset=0
- I tried to delete exports. When I deleted exports of type JSON, the file disappeared as well, nice! However when I deleted exports of type CSV, I could still access the file.
- In the exports list response, I can see the collection ID in the response. Can we include also the name so I can present a link to the collection in the UI? Similar to the job info!

### Questions:
- For other jobs, if they are related to a source image collection, we link the job to it (see screenshot below). I wonder if we should do the same for export jobs? If so, I think this should happen on the backend side in a streamlined way.
- The request body we usually keep pretty flat when creating entities. In this case, I see we are specifying the collection as part of a `filter` object. I don't mind really, but could be nice to streamline the API?
- When creating other entities, we specify `project_id` in the request body and the field is called `project`. If we can do the same for exports, I can avoid some special handling in the frontend code.
- I see the collection field is required, I personally think we could keep it optional. What do you think?

Screenshot 2025-03-21 at 12 11 30

mohamedelabbas1996:
Hey Anna! 😊 Thanks a lot for testing and also for fixing the Sentence case I missed on some fields! 🙌

Just a quick update on your feedback:

1.     I fixed the sorting issue  confirmed it's now working as expected (e.g. using ordering=-created_at).

2.     I’ve added the collection name to the response when the collection filter is applied.

3.     I also included the collection info in the job response if the filter is set..

4.   Regarding the request body, I kept the collection field inside the filters object, since it's a filter on occurrences. I think this makes more sense especially if plan to add more filters in the future.

5.  I added a `project` parameter in the request body (takes the project ID), similar to how we create other entities. You can still use `project_id` in query params for filtering the list view.

6. Exported files are correctly deleted for both JSON and CSV formats when an export is deleted. That works for me locally and I also added unit tests to make sure it behaves as expected.

7.  Finally, I made the collection filter optional .  initially I required it to reduce backend workload, but agreed it makes more sense to keep it flexible.
 
Would be happy to arrange one more testing session if you have time :)

annavik:
Thanks Mohamed, I have tested your updates!
 
**Summary:**
1. Sorting works now!
2. I can now create exports without specifying a collection.
3. I get the collection name in the export response now, thanks! I wonder if we can include in a more streamlined way though (use same field names etc. as we do when collection information is included elsewhere)? Also sorting exports based on collection would be nice!
4. I get the collection name in the job response now, thanks!
5. I see what you mean! I would personally prioritize consistency over this update, but I'm fine keeping it as is as well. I'll let you and @mihow decide how to handle this :)
6. I can now specify project ID in a similar way for all entities, thanks!
7. I can still access CSV files after delete. I'll send you a clip on Slack.

**Left to do:**
- Streamline how collection information is embedded (I suggest we use job endpoints as a reference) 
- Support sorting based on collection
- Look into delete problem
- Make a decision about request body format

mihow:
Changes we propose:

Backend:
- Remove `collection` on top level of Export response
- Make `filters: { collection_id: 1 }`
- Add `filters_display: { collection: { id: 1, name: "My collection" }}` 

when `filters{}` is set (DataExport is created)
populate `filters_display` as well

- Add a field for number of records exported 

In the UI:
- Show a single column with all of the export filters from export_display (similar to collections sampling parameters)
- Add # records exported

mihow:
@mohamedelabbas1996 

- I added one method of building the absolute URL for the final export url.
- I tested two ways of generating the nested representations of the related models. I think #2 is simpler if we only ever need the ID & a name of the related object. But it I think it should still be moved to a cached property, so we don't generate a lot of extra n+1 queries as the export history grows.

```
            "filters_display_one": {
                "collection": {
                    "id": 42,
                    "name": "VCE - Every 10 minutes",
                    "details": "http://django:8000/api/v2/captures/collections/42/",
                    "method": "common_combined",
                    "user_permissions": []
                }
            },
            "filters_display_two": {
                "collection": {
                    "id": "42",
                    "name": "#42 VCE - Every 10 minutes"
                }
            },
```


- More importantly, I tested the exports on some larger datasets! It seems like it works quickly and without failure. However I don't know if all records are included. The number of rows exported doesn't match the number of occurrences in the collection. Do you have a test for this?

mohamedelabbas1996:
Thanks @mihow for pushing these updates. The `filter_display_two` looks great for me I will move forward and try to add caching!

ami/exports/models.py
```diff
@@ -0,0 +1,111 @@
...
+        with open(file_temp_path, "rb") as f:
+            default_storage.save(file_path, f)
+        file_url = f"{settings.MEDIA_URL}{file_path}"
+        return file_url
+
+    def run_export(self):
+        from ami.exports.registry import ExportRegistry
+
+        export_format = self.format
+        export_class = ExportRegistry.get_exporter(export_format)
```
mihow, line 92: I like this convention when getting a class dynamically
```
ExportClass = ExportRegistry.get_exporter(export_format)
exporter = ExportClass()
```

ami/exports/models.py
```diff
@@ -0,0 +1,111 @@
...
+    format = models.CharField(max_length=255, choices=get_export_choices())
+    filters = models.JSONField(null=True, blank=True)
+    file_url = models.URLField(blank=True, null=True)
+    # Number of exported records.
+    record_count = models.PositiveIntegerField(default=0)
+    # Size of the exported file in bytes.
+    file_size = models.PositiveBigIntegerField(default=0)
+
+    @cached_property
+    def filters_display(self):
```
mihow, line 38: It is preferable that we save these display filters in a field on the model it self, so we only calculate them once and avoid all N+1 queries when fetching the list of DataExports.

Change `filters_display` to a field, and rename the function to generate this to `get_filters_display()`. Then call that in the save() method. 

See:
https://github.com/RolnickLab/antenna/blob/d0e0f382d90e6e270002a96c3831593b44c1d67c/ami/main/models.py#L1528C1-L1543C53

ami/exports/base.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, data_export):
+        self.data_export = data_export
+        self.job = data_export.job if hasattr(data_export, "job") else None
+        self.project = data_export.project
+        self.queryset = apply_filters(
+            queryset=self.get_queryset(), filters=data_export.filters, filter_backends=self.get_filter_backends()
+        )
+        if self.job:
+            self.job.progress.add_stage_param(
+                self.job.job_type_key, "Number of records exported", f"{self.queryset.count()}"
```
mihow, line 26: Rename this to "Total records to export"
Add another stage parameter: "Number of records exported" that updates with the progress bar.

ami/exports/models.py
mihow: Hi @mohamedelabbas1996, I talked with Anna about this today. She tried displaying the extra detail in the UI and it ended up looking cluttered. We agreed that it's okay to revisit this later! Let's keep the filters schema with nested attributes `filters_display: {"collection": {"id": 1}"`. But I think it can be generated on the fly as you have it. Just remove the extra queries and only use the ID from the related objects.

mihow (changes requested):
I'm trying to update our existing filters in the preview env so that I can re-save them and generate `filters_display`. However the admin is looking for an old property.
```
AttributeError at /admin/exports/dataexport/
'DataExport' object has no attribute 'status'
```

Since we are making the DataExport model for the first in this PR, I don't mind migrating backwards then forwards, but in other cases you can add a migration step to call `save()` on all existing DataExport instances.

ami/exports/views.py
```diff
@@ -0,0 +1,86 @@
...
+        # Validate format
+        if format_type not in ExportRegistry.get_supported_formats():
+            return Response(
+                {"error": f"Invalid format. Supported formats : {ExportRegistry.get_supported_formats()}"},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+
+        # Check optional collection filter
+        collection = None
+        collection_id = filters.get("collection")
```
mihow, line 52: change this parameter name to `collection_id`

ami/exports/views.py
```diff
@@ -0,0 +1,86 @@
...
+        serializer = self.get_serializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        validated_data = serializer.validated_data
+        format_type = validated_data["format"]
+        filters = validated_data.get("filters", {})
+        project = validated_data["project"]
+
+        # Validate format
+        if format_type not in ExportRegistry.get_supported_formats():
```
mihow, line 44: Can the serializer can know about these choices and validate for us?

mihow:
@mohamedelabbas1996 I recreated all the migrations, we went from 9 migrations to 2!

ami/exports/serializers.py
```diff
@@ -55,6 +56,12 @@ class Meta:
             "updated_at",
         ]
 
+    def validate_format(self, value):
```
mihow, line 59: very clean! nice work

mihow (approved):
I'm closing this epic journey! Well done @mohamedelabbas1996

PR #733, opened 2025-02-19
[Draft] Add reference images from Fieldguide to Taxa views

annavik:
Fixes https://github.com/orgs/RolnickLab/projects/8/ by simplifying the taxa views, both backend and frontend side.


mihow:
@annavik I was able to immensely speed up the backend queries without changing too much. A couple things:
- We should always use the `parents` attribute now instead of the recursive `parent` and `parent.parent`. So this may need some more attention 
- Always need to pass the project_id to the taxa endpoints, even for detail views (I probably should have made project part of the URL path instead of a query param orignally)

mihow:
@annavik the frontend and backend are playing nicely together now. I was able to speed up the backend queries almost 1000x for the taxa view including live occurrence counts. I would love to deploy this optimizations to live asap since the current Taxa page is crashing the site. However we either need to complete the Fieldguide integration, or do some surgery and deploy only the backend changes. If we wait on the Fieldguide integration, I still think we should remove the occurrences list in the Taxon detail modal, and just link to the occurrences page. Thoughts?

annavik:
> @annavik the frontend and backend are playing nicely together now. I was able to speed up the backend queries almost 1000x for the taxa view including live occurrence counts. I would love to deploy this optimizations to live asap since the current Taxa page is crashing the site. However we either need to complete the Fieldguide integration, or do some surgery and deploy only the backend changes. If we wait on the Fieldguide integration, I still think we should remove the occurrences list in the Taxon detail modal, and just link to the occurrences page. Thoughts?

Wow great work, it's so much faster!! I agree we should try deploy this really soon.

Without the Fieldguide integration, we will not have any visual content to show in the detail view, I think not optimal for the current layout 🤔 Maybe the current solution (doing a frontend side search) could work until we are ready to integrate backend side. It's not perfect (extra fetching is required, there is a risk we will not find a match, etc.), but I think better than an empty column?

If we decide to do that, here is my list of remaining items for this PR:
- Have a look at the parent thing you mentioned
- Add an extra check to make sure the scientific name in Fieldguide search response is matching the search term (only accept exact matches) 
- Add a loading state to the image column
- Skip other Fieldguide information (like the common name) for now

What do you think about this plan?

annavik:
We decided to include backend updates as a separate PR and save the Fieldguide integration for later. Also Michael did some more backend optimizations which means some of the frontend simplifications are no longer needed.

I will close this PR, but we can reuse some FE code from it when it's time to rework the taxon detail view :)

PR #736, opened 2025-02-20
Fix connection loop when getting status of processing service

### LINKED ISSUE #701: Fix connection loop when getting status of Processing Service
When checking the status of a Processing Service, the UI keeps connecting, even when the result is successful. This is happening on live, I haven't seen it locally. But try adding the live processing service to the local env to see if you can reproduce.

Image

vanessavmac:
## Summary/List of Changes

Fix the endless connection loop by separating the processing service `last_checked` and `updated_at` fields. When `get_status()` is called for a processing service, `last_checked` and `last_checked_live` should be updated, but `updated_at` should not be changed. `updated_at` should only change when actual data in the processing service is updated.

### Related Issues

Fixes #701 

### How to Test the Changes

Test by creating a processing service with endpoint url https://ml.dev.insectai.org


ami/ml/models/processing_service.py
```diff
@@ -92,16 +92,15 @@ def get_status(self):
...
 
         try:
             resp = requests.get(info_url)
             resp.raise_for_status()
         except requests.exceptions.RequestException as e:
             latency = time.time() - start_time
-            self.last_checked_live = False
-            self.last_checked_latency = latency
-            self.save()
+            ProcessingService.objects.filter(pk=self.pk).update(last_checked_live=False)
```
mihow, line 102: Thanks for finding a method to update without updating the updated_at field. I tested a different method that won't make a new database query every call:
```
self.last_checked_live = False
self.save(update_fields=["last_checked_live"])
```

mihow (changes requested):

ami/ml/models/processing_service.py
mihow: And just save one time if possible

mihow (approved):

ami/ml/models/processing_service.py
```diff
@@ -88,51 +90,52 @@ def create_pipelines(self):
...
     def get_status(self):
-        info_url = urljoin(self.endpoint_url, "info")
+        """
+        Check the status of the processing service.
+        This is a simple health check that pings the /readyz endpoint of the service.
+        """
+        ready_check_url = urljoin(self.endpoint_url, "readyz")
         start_time = time.time()
         error = None
+        pipeline_configs = []
```
mihow, line 100: I think the pipeline_configs could be completely removed from the simple status check. Now its returning an empty list.

PR #738, opened 2025-02-22
Disable a pipeline for a given project

### LINKED ISSUE #695: Ability to disable a pipeline for a given project
When a project registers a Processing Service, all pipelines are added to the project and made available to users. However the project manager will likely only want to make a subset of the pipelines available to the project users. For example, if you add our default Rolnick Lab processing service, all regional models get added to your project as choices for processing images. However you probably don't want users to accidentally select the Costa Rica model in Denmark, or older models. We plan to have a method of auto-selecting regional models in the future based on a geographic shapefile, however there are already a lot of models and a way to limit the choices per project manually seems necessary.

vanessavmac:
## Summary

Add an intermediate model to store the relationship between a project and a pipeline. 

### List of Changes

* Added additional intermediate model `ProjectPipelineConfig`
* Added an additional column to display whether a pipeline is enabled for a project

### Related Issues

- Closes #695 
- Uses similar status indicators as in #715 
- Best to merge #705 first; some changes to `ami/ml/views.py` were taken from #705

## Detailed Description

...

### How to Test the Changes

...

### Screenshots

Screenshot 2025-02-22 at 2 54 47 AM

## Checklist


ami/ml/models/pipeline.py
```diff
@@ -192,9 +192,32 @@ def process_images(
...
     ]
 
+    if job_id:
+        try:
+            config = pipeline.project_pipeline_configs.get(project_id=job.project).config
+            task_logger.info(
+                f"Sending pipeline request using {config} from the project-pipeline config "
+                f"for Pipeline {pipeline} and Project {job.project}."
+            )
+        except Exception as e:
```
mihow, line 202: Will you raise the specific Exception here? Should be `ProjectPipelineConfig.NotFound` or `ObjectNotFound`

ami/ml/models/pipeline.py
```diff
@@ -192,9 +192,32 @@ def process_images(
         if url
     ]
 
+    if job_id:
```
mihow, line 195: We should be able to get the pipeline config even if we are not in the context of a job. Is there another place to get the current project ID? We should add it to the process_images function.

ami/ml/models/pipeline.py
```diff
@@ -876,7 +899,9 @@ class Pipeline(BaseModel):
             "The backend implementation of the pipeline may process data in any way."
         ),
     )
-    projects = models.ManyToManyField("main.Project", related_name="pipelines", blank=True)
+    projects = models.ManyToManyField(
+        "main.Project", related_name="pipelines", blank=True, through="ml.ProjectPipelineConfig"
```
mihow, line 903: Nice work implementing our first through model!

ami/ml/models/processing_service.py
```diff
@@ -59,7 +60,15 @@ def create_pipelines(self):
                 )
                 created = True
 
-            pipeline.projects.add(*self.projects.all())
+            for project in self.projects.all():
+                project_pipeline_config, created = ProjectPipelineConfig.objects.get_or_create(
```
mihow, line 64: I'm curious if you have to create the ProjectPipelineConfig explicitly or if it will be created automatically when you create a pipeline with projects, since it is configured as a through model. Did you test that? You can set the default `enabled=True` in the model definition.

ui/src/pages/overview/pipelines/pipelines-columns.tsx
```diff
@@ -49,4 +49,32 @@ export const columns: (projectId: string) => TableColumn<Pipeline>[] = () => [
     sortField: 'updated_at',
     renderCell: (item: Pipeline) => <BasicTableCell value={item.updatedAt} />,
   },
+  // {
```
mihow, line 52: Can these comments be dropped or do you want to discuss?

mihow:
Great stuff Vanessa! I can't wait to put this one into action. I tested on a recent DB snapshot and updated the migration file. I left a few comments, the most important one is how we get the project-pipeline-config in the `process_images` function. You _can_ get the project_id from one of the images in the collection, but I think we should pass `Project` or `project_id` to the process images function. It's already available in the parent function.

ami/ml/models/processing_service.py
vanessavmac: Just tested, we have to create it explicitly. In the `ami/ml/models/project_pipeline_config.py` I've also set the default values like this:
```
enabled = models.BooleanField(default=True)
config = models.JSONField(default=dict, blank=True, null=True)
```

vanessavmac:
@mihow Thanks for reviewing! I've made the requested changes. Let me know if there's any other changes I should make

mihow (approved):
@vanessavmac  I made one change and helped fix the circular imports by using string references to the Project & Pipeline models in the ProjectPipelineConfig class definition. I tested again on a real DB snapshot and all is well! Great work!

PR #693, opened 2025-01-24
Framework for User Permissions & Roles

mohamedelabbas1996:
## Summary
This PR introduces fine-grained object-level permissions for the Project model and related entities using Django Guardian. It establishes a roles and permissions framework, including Project Manager, Basic Member, and Identifier roles, with appropriate permissions for each.

### List of Changes
* Implemented object-level permissions for Project and related entities.
* Added role-based access control (Project Manager, Basic Member, Identifier).
* Auto-assign Project Manager role to the project owner.
* Auto-assign Basic Member role to project members.
* `project_id` query parameter made optional for all project related entities views.

### Related Issues
Closes #703 #219

### Detailed Description

This PR implements a fine-grained permissions framework for the Project model and its related entities:
Job, Deployment, SourceImageCollection, S3StorageSource, Site, Device, Identification. This ensures that access is properly restricted and prevents unauthorized modifications to project-related entities.

Roles and Permissions:
  * Project Manager: Full permissions on the project and its entities. Auto-assigned to the project owner upon creation.
  * Basic Member: Can access private project data and star captures. Auto-assigned to project members.
  * Identifier: Can create, update, and delete Identifications created by the user.
  

## Deployment Notes

* Requires database migrations to support role assignments.
* Ensure Django Guardian is properly installed and configured.
* Run `manage.py assign_roles --source "Identifiers CSV file URL "` to reset permissions and assign roles for existing project owners and members and  assign identifier, project manager roles to users based on a CSV file.
* Post an update on multiple channels (Slack and perhaps over email). Also confirm which projects users should be apart of.

## Checklist

- [x] I have tested these changes appropriately.
- [x] I have added and/or modified relevant tests.
- [x] I updated relevant documentation or comments.
- [x] I have verified that this PR follows the project's coding standards.
- [x] Any dependent changes have already been merged to main.
- [x] Merge in main branch & fix conflicts
- [x] Test with db snapshot (michael), real projects

Follow-up PRs:
- Read-permissions (filter list views by current user)
- Support for many to many relationship to projects

PR #741, opened 2025-02-26
Update UI for identifications and show multiple suggestions per classification

### LINKED ISSUE #352: Show more than the first prediction in UI
We should find a way to display the top-N (3 to 5) predictions in the occurrence detail view, and the session detail view. The occurrence detail also needs a way for the user to select one of the other predictions as the right choice. Currently only the top-1 (first prediction) is available in the UI. If there is a quick & dirty way in the current UI, we should start with that.

It was discussed that this could be available when suggesting an alternate ID, or in the list of machine learning suggestions. And there should be Genus level & Family level suggestions that have an Agree button as well. It should be an asynchronous call from the API to get additional suggestions. 

Image

annavik:
## Summary

In this PR we implement the new design for identifications. This includes showing multiple suggestions per classificatoins.

### List of Changes

* Use new components to present identifications
* Fetch classification details when a machine prediction is expanded
* Show multiple suggestions per classification
* Update bunch of other related components to use our new building blocks from Nova, to make UI more streamlined

### Related Issues

- Closes #352.
- New UI components were implemented and added to Nova in https://github.com/RolnickLab/nova/pull/9.

## Detailed Description

We have going a bit back and forth around these changes. The initial design (see [Figma](https://www.figma.com/design/nEY85OwvYIOqBt5h13zS5P/Antenna?node-id=2677-2&p=f&t=KqvSGsAZ16ZutC2s-0)) suggested separating "confirming identifications" from "custom identifications" and also a grouped structure. We did like these ideas, however they would mean some significant logic changes, which we decided we are not fully ready for. We have a few stakeholders that needs to be involved in such decisions, some active labeling projects to consider and also some more high prio tasks. Also, our most urgent need for identifications was to present machine predictions in a more detailed way. Therefore, we decided to skip the suggested logic updates on the backend side, focus on our most urgent needs, but still try use as much as possible from the updated design, with some compromises.

### How to Test the Changes
Please test all features related to identifications:
- Confirm machine prediction
- Confirm human suggestion
- Suggest ID
- Expand machine prediction
- Apply classification from an expanded item
- Delete identification

### Screenshots
**Before:**

Screenshot 2025-02-26 at 15 15 27

**After:**

Screenshot 2025-02-26 at 15 12 53
Screenshot 2025-02-26 at 15 12 58

## Deployment Notes

This PR needs a small update after the permission PR (see #693) is merged!


annavik:
@mihow I did notice that if I try to fetch classification details, I get 500-responses from the backend in some cases. This is how we present such a failure in UI:
Screenshot 2025-02-26 at 15 54 25
Screenshot 2025-02-26 at 15 54 19

On localhost, this is the case for the initial occurrences I get, if I spin up a new database. If I process captures, the occurrences that will be generated can be expanded without backend errors.

Is this something to worry about? Maybe it's just the dummy data database that needs to be updated?

mihow:
> @mihow I did notice that if I try to fetch classification details, I get 500-responses from the backend in some cases. This is how we present such a failure in UI: Screenshot 2025-02-26 at 15 54 25 Screenshot 2025-02-26 at 15 54 19
> 
> On localhost, this is the case for the initial occurrences I get, if I spin up a new database. If I process captures, the occurrences that will be generated can be expanded without any problems.
> 
> Is this something to worry about? Maybe it's just the dummy database that needs to be updated, or would this affect all older classifications?

Thanks, yes it shouldn't throw a 500 error if there are not top N choices. I will look into that.

mihow:
I found some cases where there is top-n data but the species values are `null` (even though there are scores). Mainly for the moth/non-moth model. So I need a little more time on this one.

mihow:
@annavik Will you fix the merge conflicts while we wait for me to fix the backend issue?

annavik:
> @annavik Will you fix the merge conflicts while we wait for me to fix the backend issue?

Done! I have synced this with main and included some final tweaks as well. I think this one is ready now!

mihow:
@mihow todo: only return the full list of logits & scores when requests explicitly by the user (query param or another view)

annavik:
> @mihow todo: only return the full list of logits & scores when requests explicitly by the user (query param or another view)

This would be good to fix of course (as we talked about before), however even if we would get the full list, the frontend will now slice it to 5 items max.

annavik:
This one is synced with main now! I have also added a few last tweaks:
- Add tooltips to expand button
- Tweak error messages to be in line with tooltips
- Make sure we are never showing more than 5 top predictions

mihow (approved):
I tested on staging and it's working quite well! This feature is a big deal. Sorry I put the breaks on merging for so long. Thank you for keeping the branch up-to-date. The new styles look great too!

PR #742, opened 2025-02-26
Add flower to .gitignore

annavik:
I think something Docker related, but not fully sure, looks like some binary file? It keeps getting listed as a untracked file for me.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,7 +50,7 @@ coverage.xml
 *.mo
 *.pot
 
-# Django stuff:
+# Django stuff
 staticfiles/
 
 # Sphinx documentation
@@ -273,3 +273,6 @@ ami/media/
 backups/
 data/
 sandbox/
+
+# Other
+flower


mihow:
This has been bugging me too but I am wondering what the root cause is!

annavik:
Looks like som database file? Still not fully sure if the correct solution should be to ignore it or make sure it's not being generated?

A clue:
Screenshot 2025-02-27 at 16 49 47

From ChatGPT:
> Since the file is named "flower", it might be related to [Celery Flower](https://flower.readthedocs.io/en/latest/), a real-time monitoring tool for Celery workers. If you’re using Celery in your project, this could be a cache or database file generated by Flower.

mihow:
@mohamedelabbas1996 can you try and make the flower service save the flower database file to `./data/` instead of the project root? That folder is already ignored in git. I think this started happening when I upgraded flower a few weeks ago. Flower is a web interface for viewing the Celery background tasks. 

Check out the config here:

https://github.com/RolnickLab/antenna/blob/ad9007fcc6906af65ddc502c273932c74f12fb1f/docker-compose.yml#L99-L104

https://github.com/RolnickLab/antenna/blob/ad9007fcc6906af65ddc502c273932c74f12fb1f/compose/local/django/celery/flower/start#L6-L8

mihow (approved):

PR #744, opened 2025-02-27
Streamline project entity columns

annavik:
## Summary
Just some minor cleanup for project entity columns! :)

### List of Changes
* Use "Sentence case" for table and column names (I have no strong opinions about the case, but I think it's looks nice if we use the same)
* Make sure all tables has columns for "Created at" and "Updated at"
* Present "Created at" and "Updated at" after entity specific information


mihow (approved):
Thanks for keeping us consistent!

PR #756, opened 2025-03-04
Handle UI runtime problems reported on Sentry

annavik:
Here are our runtime errors reported on Sentry last 30 days (UI only).
Screenshot 2025-03-04 at 13 41 59

In this PR we take care of issue 1 and 3. The other issues I think are related to request timeouts, they are more difficult to reproduce and I can't fully understand the details (they are not including any references to the frontend code base).

About the fixes:
- First issue is related to some missing data. For human identifications, the frontend app expects them to be associated with a user. However, for a few edge cases, seems like the user object is missing (see https://api-production.dev.insectai.org/api/v2/occurrences/11597/ for example). We can see the problem live here: https://antenna.insectai.org/projects/18/occurrences?ordering=-determination_score. After the fix, we will still present these identifications in UI, but with the user name "Unknown user".
- The other issue was related to some unhandled promise rejections. If we await an async method that might throw errors, we should catch these errors. However the React Query library that we use is already taking care of the state handling for us, so we can actually leave the catch block empty. When using this lib, in most situations there is no need to await these methods. Example: https://tanstack.com/query/v3/docs/framework/react/guides/mutations
- Also, we will include a fix for a problem that was recently discovered about pipelines not being connected to a processing service in the preview environment. After the fix, we will consider such pipelines as offline in the session detail view, instead of having the app crash.


ui/src/data-services/models/occurrence-details.ts
```diff
@@ -53,6 +53,7 @@ export class OccurrenceDetails extends Occurrence {
     }
 
     this._humanIdentifications = this._occurrence.identifications
+      .filter((i: any) => !!i.user) // Workaround to avoid crash when the backend is not returning any user
```
mihow, line 56: What about keeping the identification but showing an unknown user? I wonder if this happens when the user is deleted.

annavik: You are right, that is better! History will make more sense. I have updated the code, such cases will now be presented like this:

Screenshot 2025-03-05 at 10 31 32

mihow (approved):

PR #764, opened 2025-03-10
Adjust page size based on page width to avoid gallery gaps

annavik:
## Summary

This is a small fix to a address an issue that was brought up on Slack. For the project overview, we will now adjust the page size depending on window width to avoid gaps in the grid (making it more clear when a gallery page has reached its end).

If page width is > 1024 px (meaning we will show 3 columns), we will use page size 21. If not, we will use default page size 20.

We will update the page size if window is resized and refetch projects if needed.

## Screenshot

Screenshot 2025-03-10 at 22 08 10


mihow (approved):
Much better!!

PR #768, opened 2025-03-11
Fix processing service error handling

vanessavmac:


mihow (approved):
Thanks for catching this

PR #772, opened 2025-03-12
Add page "Code of conduct" to UI

annavik:
## Summary

In this PR we include our new page "Code of conduct" in Antenna. The page is setup in a similar way as the page "Terms of service". We use same style and structure.

The content for this new page can be edited from a markdown file, see `ui/src/components/info-page/code-of-conduct-page/code-of-conduct.md`.

Since I modified our routing a bit as part of this PR, I also included a simple 404 page in this PR. We were missing that! For now, we use it if the current route can't be matched to a page.

## Screenshots
Screenshot 2025-03-12 at 09 50 15

Screenshot 2025-03-12 at 10 27 43

### DIFF
--- a/ui/src/app.tsx
+++ b/ui/src/app.tsx
@@ -11,6 +11,8 @@ import { TermsOfServicePage } from 'components/info-page/terms-of-service-page/t
 import { Menu } from 'components/menu/menu'
 import { TermsOfServiceInfo } from 'components/terms-of-service-info/terms-of-service-info'
 import { useProjectDetails } from 'data-services/hooks/projects/useProjectDetails'
+import { AlertCircleIcon, ChevronRightIcon } from 'lucide-react'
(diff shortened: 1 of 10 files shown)


ui/src/app.tsx
```diff
@@ -90,7 +91,22 @@ export const App = () => (
           <Route path="collections/:id" element={<CollectionDetails />} />
           <Route path="*" element={<UnderConstruction />} />
         </Route>
-        <Route path="/:slug" element={<InfoPageContainer />} />
```
annavik, line 93: We will no longer try to render an info page for generic slugs. This is a left over from when the backend returned dynamic content here.

Now we require exact matches!

mihow (approved):
Awesome this is a milestone! Perhaps there will be a reason to bring back dynamic pages in the future, but the addition of the 404 page and the fast response for bad paths is much better!

Issue #745, opened 2025-02-27
Add filtering options for Taxa Lists


mihow:
## Summary

Add a select box for Taxa Lists that belong to the project in the Taxa & Occurrences views. This will require adding some data to the API serializer as well. 

## Detailed Description

We have a Taxa List model with a many2many relationship with the Taxon model however it is not being utilized by or exposed to users. This is will be used for "species of interest" feature.

## Implementation Details

...

## Additional Context

...

annavik:
On the frontend side, this is what we would would need to know before we can start add UI controls:
- How do we fetch filter options for taxa lists?
- How do we pass a selected taxa list as a filtering option when fetching data?
- Should it be possible for a user to select one or may options here?

Happy to discuss this when we are ready! :)

mihow:
- Single select is fine!
- The endpoints & viewsets should be CRUD sets that are about same as the other data types (Deployment, Taxon, etc) https://api-production.dev.insectai.org/api/v2/deployments/
- For fetching choices to populate the form field, I'm guessing we should follow the pattern we are using to fetch choices in the Occurrence list (fetch from the data type's API list view) https://antenna.insectai.org/projects/18/occurrences
- A new `taxa_list_id` filter will need to be added to the Taxon viewset

Also let's start thinking about ways to add a Taxon to a list. From the Taxon pages, perhaps you can add taxa to lists using similar components we use for identifying Occurrences?? At least from the Taxon detail page to start. For this PR we can start by using the Django admin to manage lists.

annavik:
Sounds great, thank you M! I agree about handling list management from UI as a separate task.

@mohamedelabbas1996 I think the frontend updates will be straight forward here, I suggest we start when the backend endpoints are in place to use our time more efficiently! :)

PR #777, opened 2025-03-13
Species of Interest TaxaList Filter

mohamedelabbas1996:
## Summary

This PR introduces a Taxa List filter for occurrences and taxa views, enabling filtering based on taxons in a  Taxa List . The filter ensures that Occurrences and Taxa can be filtered by their presence in a Taxa List or if their ancestors belong to a specified Taxa List. Additionally, Taxa Lists can  be managed from the admin dashboard, allowing users to Create new Taxa Lists, Add taxa to a Taxa List and Link Taxa Lists to projects.

### List of Changes

* Added `OccurrenceTaxaListFilter`: Enables filtering Occurrence objects where determination is in a TaxaList or has an ancestor in the list.
* Added `TaxonTaxaListFilter`: Enables filtering `Taxon` objects where the taxon is in a `TaxaList` or has an ancestor in the list.
* Added Taxa List filter UI for Occurrences and Taxa Pages:
   Users can now filter occurrences and taxa by selecting a Taxa List from a dropdown.

### Related Issues

#745 

### How to Test the Changes
1. Verify Taxa List Filtering in Occurrences and Taxa Pages
2. Navigate to the Occurrences Page.
3. Locate the Taxa List filter dropdown.
4. Select a TaxaList.
5. Verify that:
        Only Occurrences with a determination in the selected TaxaList or its ancestors are displayed.
        Changing the TaxaList filter updates the results accordingly.
6. Repeat the same steps on the Taxa Page.
## Screenshots 
image
image
image

## Deployment Notes
No database migrations required.

## Checklist


ami/main/api/views.py
```diff
@@ -967,6 +969,40 @@ def filter_queryset(self, request, queryset, view):
...
+            taxa_list = TaxaList.objects.filter(id=taxalist_id).first()
+
+            if taxa_list:
+                taxa = taxa_list.taxa.all()  # Get taxalist taxon objects
+
+                logger.debug(f"Filtering by taxalist {taxalist_id} with {len(taxa)} taxa")
+                logger.debug(f"Filtering by taxalist {taxalist_id} taxa_ids {taxa}")
+
+                # filter by the exact determination
+                query_filter = Q(determination__in=taxa)
```
mihow, line 993: Can you use a sub query instead? If you generate the list of taxa in python first and then pass it back, it may be too long. Some regional taxa lists will have over a few thousand items.

config/api_router.py
```diff
@@ -27,6 +27,7 @@
 router.register(r"detections", views.DetectionViewSet)
 router.register(r"occurrences", views.OccurrenceViewSet)
 router.register(r"taxa", views.TaxonViewSet)
+router.register(r"taxalists", views.TaxaListViewSet)
```
mihow, line 30: Will you try mounting under /taxa? `/taxa/lists`? Like the ml prefix. We will eventually move all taxonomy related things into a taxa app.  If it causes a headache then no worries.

ami/main/api/views.py
```diff
@@ -967,6 +969,40 @@ def filter_queryset(self, request, queryset, view):
...
 
+class OccurrenceTaxaListFilter(filters.BaseFilterBackend):
+    """
+    Filter occurrences by TaxaList.
+    """
+
+    query_param = "taxalist"
+
+    def filter_queryset(self, request, queryset, view):
+        taxalist_id = request.query_params.get(self.query_param)
```
mihow, line 980: I'm glad you are using the IntegerField here for validating the query param! Here is how we are doing it in other filters:
https://github.com/RolnickLab/antenna/blob/c18ff50b4f2d5207b8dfa8198089714c200d249e/ami/main/api/views.py#L864-L868

ami/main/api/views.py
mihow: Example:
https://github.com/RolnickLab/antenna/blob/1adb99da52166b4a18e353abf32935af135d8778/ami/main/models.py#L2343-L2346

mihow: Actually I don't think a subquery is possible with our parents_json field. We can come back to that later. Also the first filter can occur on the Taxa view. Instead of the Occurrences view. Ideally both, but Taxa is the most important for the first demo.

ami/main/api/serializers.py
```diff
@@ -532,6 +533,25 @@ def get_occurrences(self, obj):
...
+
+    class Meta:
+        model = TaxaList
+        fields = ["id", "name", "description", "taxa", "projects"]
+
+    def get_taxa(self, obj):
+        """
+        Return URL to the taxa endpoint filtered by this taxalist.
+        """
+        return reverse_with_params(
```
mihow, line 548: Thanks for finding this!

mihow:
It works! 
image

mihow:
@annavik will you review the frontend changes on this one?

We will likely have a Taxonomy section in the new Project sub-menus as well!

mihow:
@annavik I pushed one simple method of hiding the Taxa List filter if there are no lists configured for the current project. I couldn't see how to add a `showFilter()` method to the TaxaList type because we don't have the current project in that context.

If we do more of these, we could add a `hideIfDisabled` attribute to the FormControl component? Because the Taxa List filter component has already checked the data and disabled the form field. 

Anyway, let us know what you think of the other changes. We can improve the lists feature over time.

ui/src/data-services/hooks/taxalists/useTaxalists.ts
```diff
@@ -0,0 +1,49 @@
...
+import { FetchParams } from 'data-services/types'
+import { getFetchUrl } from 'data-services/utils'
+import { useMemo } from 'react'
+import { UserPermission } from 'utils/user/types'
+import { useAuthorizedQuery } from '../auth/useAuthorizedQuery'
+
+// Convert server response to a TaxaList instance
+const convertServerRecord = (record: ServerTaxaList) => new TaxaList(record)
+
+export const useTaxaLists = (
```
annavik, line 12: I think this whole file can be removed, seems to be a repetition of `ui/src/data-services/hooks/taxa-lists/useTaxaLists.ts`.

ui/src/data-services/models/taxalist.ts
```diff
@@ -0,0 +1,64 @@
...
+  id: number
+  name: string
+  description?: string
+  taxa: string // URL to taxa API endpoint (filtered by this TaxaList)
+  projects: number[] // Array of project IDs
+  created_at?: string
+  user_permissions?: UserPermission[]
+}
+
+export class TaxaList {
```
annavik, line 14: If we extend class `Entity` we can simply this class a lot! I think `taxaUrl` is the only custom field.

ui/src/utils/useFilters.ts
```diff
@@ -87,6 +87,10 @@ export const AVAILABLE_FILTERS: {
     label: 'Taxon',
     field: 'taxon',
   },
+  {
+    label: 'Taxa List',
```
annavik, line 91: Nit: Taxa List -> Taxa list

ui/src/data-services/models/taxalist.ts
```diff
@@ -0,0 +1,64 @@
+import { getFormatedDateString } from 'utils/date/getFormatedDateString/getFormatedDateString'
```
annavik, line 1: Nit: Rename file from `ui/src/data-services/models/taxalist.ts` to `ui/src/data-services/models/taxa-list.ts`

ui/src/data-services/constants.ts
```diff
@@ -21,6 +21,7 @@ export const API_ROUTES = {
   SESSIONS: 'events',
   SITES: 'deployments/sites',
   SPECIES: 'taxa',
+  TAXALISTS: 'taxa/lists',
```
annavik, line 24: Nit: `TAXALISTS ` -> `TAXA_LISTS`

ui/src/pages/occurrences/occurrences.tsx
```diff
@@ -72,6 +73,7 @@ export const Occurrences = () => {
     occurrences?.some((occurrence) => occurrence.id === id)
   )
   const { selectedView, setSelectedView } = useSelectedView('table')
+  const { taxaLists } = useTaxaLists({ projectId: projectId as string })
```
annavik, line 76: We are fetching taxa lists from both this view and from the control itself. I see why, but let's think about how we can avoid this... I will get back to you! :)

annavik (changes requested):
So glad you could include some FE updates @mohamedelabbas1996 ! 👏

Just let me know if you want to have a chat about anything. The multi fetch problem I want to think about a bit more, I think no show stopper though!

ui/src/pages/occurrences/occurrences.tsx
mohamedelabbas1996: I would love to hear more from you about this multi-fetch problem!

annavik:
Thanks for the fixes @mohamedelabbas1996, this looks good to me now! :) I pushed an update to make it possible to pass custom data to filter controls. Thanks to that we can now skip the multi fetching. See commit https://github.com/RolnickLab/antenna/pull/777/commits/7cd7738f0fc8d0037521979f04565a44fb36efe0 if you are curious about the solution.

I will approve this now!

annavik (approved):

annavik:
> > One detail: do you think we should exclude "id" from the filter key everywhere, to follow same pattern we use for other filters? No blocker though.
> 
> Thanks for pointing that out, Anna! I was actually following the pattern discussed in [issue #660](https://github.com/RolnickLab/antenna/issues/660), where we're aiming to use object_id to refer to the actual ID and object for the object itself (e.g., project_id vs. project). Not sure if we’re still aligning with that convention, happy to adjust if needed!

I see, I understand it's a bit confusing since we have been a bit inconsistent here! I do like having `id` as part of the key, I think it's more clear :) I also like to keep the filter feature consistent though...

Let's keep this filter key as is, if the plan is to update other filter keys in a near future? :) @mihow any thoughts?

mihow:
@annavik I'm glad you noticed this detail! And I am glad you like the `_id` version.  This will help distinguish which properties are just IDs (or specific attributes) and which ones are nested json representations of related objects. For example:

`project: { deployment_id: 2}`
`project: { deployment: {name: "Mothra", id: 2} }`

I'm sorry we haven't updated them all at once! `project` => `project_id` was the first full-project change. And now `taxa_list_id` is the first new attribute it's being applied to.

mihow:
> Thanks for the fixes @mohamedelabbas1996, this looks good to me now! :) I pushed an update to make it possible to pass custom data to filter controls. Thanks to that we can now skip the multi fetching. See commit [7cd7738](https://github.com/RolnickLab/antenna/commit/7cd7738f0fc8d0037521979f04565a44fb36efe0) if you are curious about the solution.
> 
> I will approve this now!

Thanks for sharing the fix. It seems like that approach opens the door for a lot of flexibility.

PR #784, opened 2025-03-19
Support for a default Pipeline config

mihow:
Follow up to #738


mihow (commented):
Tested on staging server

PR #786, opened 2025-03-24
Make it possible to export data from UI

### LINKED ISSUE #737: UI for creating & downloading data exports
## Summary

#720 introduces the ability to trigger exports, see their status and download a file via the backend API. We will need an interface to allow users to interact with it. 

We have a request for this to be fully ready by Mid-march. The priority features only.

## Detailed Description

They will likely be tied to a new type of job, but also likely need a dedicated list view with the history, status and download links. Something like the Collections list in Project detail page. 

**Priority features**
- Start an export from the Occurrences view with the current filters
- View a list of exports in progress or completed 
- Provide download link 

**Secondary features**
- Allow exports of Sessions
- Allow exports of Captures (collection detail view?)
- Allow export format to be configured (Darwincore zip for occurrences and sessions, COCO for detections, etc)
- Allow exports of Deployment metadata
- Email notifications working

**Future features**
- Notification center!

## Implementation Details

See API endpoints added in #720
Hopefully no design is needed for this version, I suggest using the Project detail view tabs. Something like the the Collections list that shows the history and configuration for each export triggered, where each export is tied to a job & the job status.

The main formats for now will be a 
1) a JSON export with nested data that mostly matches the current API list views
2) simplified & flattened CSV of each data type
3) a Darwin Core Archive zip of flattened TSVs with data about Occurrences, Sessions and Taxa

## Related Issues

#298 
#720 

## Additional Context

Image

annavik:
## Summary

In this PR we implement views to make export features available to users from UI.

As part if this work, we add following views:
- List exports
- Show export details
- Register new export
- Delete export
- List export jobs
- Show details for export job

### Related Issues

Closes #737.

### How to Test the Changes

The changes has to be tested with a backend that has export support. The easiest way to do that is to checkout this branch locally.

### Screenshots
List exports:
Screenshot 2025-04-01 at 16 45 28

Export details:
Screenshot 2025-04-01 at 16 47 40

Register new export:
Screenshot 2025-04-01 at 16 49 10

Delete export:
Screenshot 2025-04-01 at 16 49 29

List export jobs:
Screenshot 2025-04-01 at 16 48 38

Export job details:
Screenshot 2025-04-01 at 16 48 46


mohamedelabbas1996:
Nice work, Anna!

mihow (changes requested):
@mohamedelabbas1996 @annavik There's a change that `filters_display` is null if the backend fails to generate it (or right now, because we had existing filters that need to be migrated). We should fallback to the `filters` attribute if this is the case. It would be nice to do that on the backend in serializers. However we need to make sure the filter is `collection_id: 123` and not `collection: 123` so the schema doesn't break the frontend. The other option is to update the UI to fall back to `filters` if `filters_display` is null. That may be safest! 

image

mihow:
> @mohamedelabbas1996 @annavik There's a change that `filters_display` is null if the backend fails to generate it (or right now, because we had existing filters that need to be migrated). We should fallback to the `filters` attribute if this is the case. It would be nice to do that on the backend in serializers. However we need to make sure the filter is `collection_id: 123` and not `collection: 123` so the schema doesn't break the frontend. The other option is to update the UI to fall back to `filters` if `filters_display` is null. That may be safest!
> 
> image

We fixed this issue and pushed the changes to this branch. The frontend falls back to plain `filters` now

mihow:
@annavik We updated the total records count as soon as the export is created. But I wanted to clarify that the value on the DataExport model is not intended to be a progress value. The DataExport is meant to store the input variables for an export (format & filters) and the final output (similar to the GBIF export). Only the Job should handle the logs & "in-between" states. Does that align with your thinking?

But we did update the counts in the Job progress data to be real integers, so now they are formatted nicely in the UI!

annavik:
Thanks for pushing those changes! 🙏 

It makes sense, I don't expect the data export model to handle progress, I just want to present progress information in UI.

I think our current solution works well though! We show total records and a progress bar next to it (based on job progress). If it was a bit easier to access number of records exported, I could present that as well, but no biggie! I would currently have to search the dynamic jobs stages after this specific key. I think a bit unnecessary complexity in code for a small win.

Let's keep things as is!

mihow (approved):

mihow (approved):

Issue #791, opened 2025-04-01
Allow UI to handle Occurrences with no event/session


mihow:
## Summary

Update the occurrence list & details views to handle occurrences where `occurrence.event` is null. Such as:
https://antenna.insectai.org/projects/84/occurrences/125214

## Detailed Description

Occurrences _almost_ always are associated with a Session (Event) but there are cases where the events change and we don't want to loose an Occurrence, so the `event` attribute is set to null. Currently we filter out all Occurrences that have no event in the API response, but this query is not implemented consistently in all parts of the app (the export method exports all occurrences with or without an event). Let's let the user decide what to do with Occurrences that no longer belong to an event. But currently they cannot see them at all.

mihow:
Thanks @annavik! After reviewing the Kenya project, you can see how the events might sometimes change! After a data sync, we automatically create sessions based on a 2-hour gap. However if the sessions need to be recalculated after images have been processed, then occurrences may be shifted to a different session. In the future, the user should be able to configure what is a "survey event" for their research project. For some, they may want to monitor continuously -- but then break every day into a 24 survey event / session. Currently users _could_ configure this if they install their own instance and modify the constant variables in the python code.

PR #794, opened 2025-04-02
Make session info optional for occurrences

annavik:
## Summary

In this PR we make session info optional for occurrences. See #791 for background.

## Screenshots

After the fix, we will display occurrences without a session like this. Also any links to session details are disabled. See https://deploy-preview-794--antenna-preview.netlify.app/projects/84/occurrences/125214 for live example.

Screenshot 2025-04-02 at 13 13 34

### DIFF
--- a/ami/main/models.py
+++ b/ami/main/models.py
@@ -2170,11 +2170,14 @@ class Detection(BaseModel):
         return f"#{self.pk} from SourceImage #{self.source_image_id} with Algorithm #{self.detection_algorithm_id}"
 
 
-class OccurrenceQuerySet(models.QuerySet):
-    def with_detections_count(self) -> models.QuerySet:
+class OccurrenceQuerySet(models.QuerySet["Occurrence"]):
+    def valid(self):
+        return self.exclude(detections__isnull=True)
+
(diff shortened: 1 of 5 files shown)


mihow:
Thanks for jumping on this! Now I was able to update all Occurrence queries to be consistent and make our counts match!

image

ami/main/models.py
```diff
@@ -2170,11 +2170,14 @@ def __str__(self) -> str:
         return f"#{self.pk} from SourceImage #{self.source_image_id} with Algorithm #{self.detection_algorithm_id}"
 
 
-class OccurrenceQuerySet(models.QuerySet):
-    def with_detections_count(self) -> models.QuerySet:
+class OccurrenceQuerySet(models.QuerySet["Occurrence"]):
+    def valid(self):
```
mihow, line 2174: @mohamedelabbas1996 I introduced a new queryset method called `valid()`. Let's begin adding this method to all models and use it in place of `all()`. We could override `all()` however this makes is almost impossible to debug or troubleshoot strange instances since you will never see them, even in the Django admin.

mihow (approved):

PR #779, opened 2025-03-18
Add affiliation field to user profile, update admin, UI, and migration

Esmat434:
### Summary
- Added an `affiliation` field to the `User` model to store users' organization or affiliation.
- Created and applied migrations for the new field.
- Updated Django admin to display and filter users based on affiliation.
- Updated the public UI to show the affiliation in user profiles.

### Why?
This change helps track which users are working on which projects and improves user management.

PR #799, opened 2025-04-05
Add affiliation field to user profile and update admin & UI

Esmat434:
### Summary
- Added an `affiliation` field to the `User` model to store users' organization or affiliation.
- Created and applied migrations for the new field.
- Updated Django admin to display and filter users based on affiliation.
- Updated the public UI to show the affiliation in user profiles.

### Why?
This change helps track which users are working on which projects and improves user management.


f-PLT:
Hi there.

Thank you for your enthusiasm, but please do not create duplicate pull requests with the same content.

See current comments and change requests in #779

PR #800, opened 2025-04-08
Improve collection form

### LINKED ISSUE #472: Choose sampling from dropdown (+ improve UI in general for registering a collection)
The way sampling method is configured at the moment is not very user friendly since it requires knowledge about keys available. I suggest we render options in a dropdown.

Skärmavbild 2024-08-01 kl  13 42 27

annavik:
## Summary

In this PR we try to make the form for creating collections a bit more user friendly.

### List of Changes

* Split form up into sections to bring some structure
* Use new date picker for date values
* Add descriptions and error handling to number inputs
* Skip fields "Earliest date" and "Earliest month" (our current assumption is that start date and end date will be enough)
* Replace text input for "Sampling method" with a select
* Conditionally show fields related to sampling method, depending on what type is selected

### Related Issues

Closes #472.

## Detailed Description

For the sampling method, we will still always pass "common_combined" to the backend. The new field "Type" is only there to make things easier to understand for users. I'm curious if we could try align this more with the sampling methods available on the backend side though.

As part of the changes, we are also trying to further align form components used in app, for example by introducing a shared `SaveButton` component to streamline how loading state and success state is presented.

### Screenshots

**Before:**
Screenshot 2025-04-08 at 16 44 12

**After:**
Screenshot 2025-04-08 at 16 41 51
Screenshot 2025-04-08 at 16 41 58
Screenshot 2025-04-08 at 16 42 06

## Deployment Notes

No backend changes are introduced in this PR, deploying should not be complicated.


ui/src/data-services/models/collection.ts
```diff
@@ -87,4 +77,30 @@ export class Collection extends Entity {
...
+    return snakeCaseToSentenceCase(this.method)
+  }
+
+  get settingsDetailsDisplay(): string[] {
+    return Object.entries(this._data.kwargs).map(
+      ([key, value]) => `${snakeCaseToSentenceCase(key)}: ${value}`
+    )
+  }
+
+  get type(): string {
```
annavik, line 95: @mihow this FE specific logic is a bit strange, I would prefer to rely on backend here if possible! Maybe we can have a chat about it if tricky to understand what I mean :)

ui/src/pages/project/entities/details-form/collection-details-form.tsx
```diff
@@ -138,7 +148,7 @@ export const CollectionDetailsForm = ({
           name: values.name,
           description: values.description,
           customFields: {
-            method: values.method,
+            method: 'common_combined',
```
annavik, line 151: Can we avoid passing this hard coded string?

ui/src/pages/project/collections/constants.tsx
```diff
@@ -1,2 +1,9 @@
-// Only some sampling methods are editable from the UI
-export const editableSamplingMethods = ['common_combined']
+// Server sampling method is not editable from the UI
```
annavik, line 1: Can we extend the list of server sampling methods here?

ui/src/data-services/models/collection.ts
mihow: Can you use `method` instead of type? I agree that we may need to work this out in a chat! perhaps we can try some API approaches together.

ui/src/pages/project/collections/constants.tsx
mihow: Yes I'd like to switch to the choices instead of the hard-coded values. But it may take some time to figure out which methods and how to expose the field choices for each of them. I wonder if we should simply this PR and continue to use a single sampling method for a little longer.

ui/src/data-services/models/collection.ts
annavik: Updates related to this are pushed now, thanks for the help!!

mihow:
@annavik can we add the selector for Stations to this PR? Or at least allow a list of comma separated IDs for now?

annavik:
> @annavik can we add the selector for Stations to this PR? Or at least allow a list of comma separated IDs for now?

Yes, on it! :)

annavik:
> @annavik can we add the selector for Stations to this PR? Or at least allow a list of comma separated IDs for now?

Actually can we handle this as a follow up task? I realize we need new UI components for this and it would be helpful to have this merged 🙏

annavik:
Can we merge this @mihow?

PR #801, opened 2025-04-08
Role Management API

mohamedelabbas1996:
## Summary

This PR introduces a complete set of role management APIs for project managers, enabling them to manage project members and their roles

### DIFF
--- a/config/api_router.py
+++ b/config/api_router.py
@@ -3,6 +3,7 @@ from django.urls import path
 from django.urls.conf import include
 from djoser.views import UserViewSet
 from rest_framework.routers import DefaultRouter, SimpleRouter
+from rest_framework_nested.routers import NestedDefaultRouter
 
 from ami.jobs import views as job_views
 from ami.labelstudio import views as labelstudio_views
@@ -17,6 +18,10 @@ else:
 router.register(r"users", UserViewSet)
 router.register(r"storage", views.StorageSourceViewSet)
 router.register(r"projects", views.ProjectViewSet)
+# Nested router for project-specific endpoints like /projects/{project_id}/members
+projects_router = NestedDefaultRouter(router, r"projects", lookup="project")
+projects_router.register(r"members", views.ProjectMemberViewSet, basename="project-members")
+
 router.register(r"deployments/devices", views.DeviceViewSet)
 router.register(r"deployments/sites", views.SiteViewSet)
 router.register(r"deployments", views.DeploymentViewSet)
@@ -61,3 +66,4 @@ urlpatterns = [
 
 
 urlpatterns += router.urls
+urlpatterns += projects_router.urls
--- a/ami/base/permissions.py
+++ b/ami/base/permissions.py
@@ -33,6 +33,16 @@ class IsActiveStaffOrReadOnly(permissions.BasePermission):
         )
 
 
+class IsProjectManager(permissions.BasePermission):
+    """
+    Allows access only to project managers for a specific project.
+    """
+
+    def has_object_permission(self, request, view, obj):
+        # `obj` is the Project instance
+        return ProjectManager.has_role(request.user, obj)
+
+
 def filter_permissions(permissions, model_name):
     """Filter and extract only the action part of `action_modelname`"""
 
--- a/ami/main/api/serializers.py
+++ b/ami/main/api/serializers.py
@@ -1517,3 +1517,8 @@ class StorageSourceSerializer(DefaultSerializer):
             "total_size",
             "last_checked",
         ]
+
+
+class AddMemberSerializer(serializers.Serializer):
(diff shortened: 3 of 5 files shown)


ami/users/api/serializers.py
```diff
@@ -10,7 +10,7 @@
 class UserListSerializer(DefaultSerializer):
     class Meta:
         model = User
-        fields = ["id", "name", "details", "image"]
+        fields = ["id", "name", "details", "image", "email"]
```
annavik: Note: I added email here because I needed it for the teams table, however I think this might also effect other API endpoints where we don't want to expose this in the response

annavik:
Thank you so much @mohamedelabbas1996, this looks great! I will collect notes in this comment while hooking this up with FE.

### Notes
- Superusers cannot list or update project members
- Can we include role display name in the project members response?
- Do you think the project members response should have support for sorting and pagination (similar to other list endpoints)?

mihow:
Perhaps this is a good moment to add an affiliation field to the User model?

ami/users/api/serializers.py
```diff
@@ -49,3 +50,135 @@ class GroupSerializer(DefaultSerializer):
...
+        if user:
+            attrs["user"] = user
+        attrs["project"] = project
+        return attrs
+
+
+class UserProjectMembershipListSerializer(UserProjectMembershipSerializer):
+    user = UserListSerializer(read_only=True)
+    role = serializers.SerializerMethodField()
+    role_display_name = serializers.SerializerMethodField()
```
mihow: I don't believe you need to list these fields as SerializerMethodFields if you are not writing a method or customizing it. If they are just fields or methods on the model.

config/api_router.py
```diff
@@ -2,18 +2,34 @@
...
 router.register(r"projects", views.ProjectViewSet)
+# NESTED: /projects/{project_id}/members/
+projects_router = routers.NestedDefaultRouter(router, r"projects", lookup="project")
+projects_router.register(
+    r"members",
+    UserProjectMembershipViewSet,
+    basename="project-members",
+)
+
+# NESTED: /projects/{project_id}/roles/
```
mihow: If this returns a fixed list of Roles for all of Antenna, independent of a project, than perhaps it should go somewhere else. But! if this will ever be configurable per project, then leave it here and consider our current list the "default roles".

perhaps `/users/roles/` is a good home for the global static roles

ami/users/api/views.py
```diff
@@ -0,0 +1,90 @@
...
+    ProjectRoleSerializer,
+    UserProjectMembershipListSerializer,
+    UserProjectMembershipSerializer,
+)
+from ami.users.roles import Role
+
+logger = logging.getLogger(__name__)
+
+
+class ProjectRolesViewSet(viewsets.ViewSet, ProjectMixin):
```
mihow: Is there a generic View class that only has what is needed here? perhaps just APIView?

mihow:
@annavik @mohamedelabbas1996 I brought this branch up-to-date with main and did some final testing! It seems to be working well. Do we need a migration or method to assign existing users to a role? For example, all of these users are members of the project, but don't have a role in the UI until I set one.

image

ami/users/api/serializers.py
```diff
@@ -3,20 +3,35 @@
...
+            "image",
+        ]
 
         extra_kwargs = {
             "details": {"view_name": "api:user-detail", "lookup_field": "pk", "lookup_url_kwarg": "id"},
         }
 
 
+class MemberUserSerializer(UserListSerializer):
+    """User serializer for membership context - includes email for management purposes."""
```
mihow: @annavik @mohamedelabbas1996 new serializer that includes email. the other one doesn't now.

mihow:
Hmm, I'm still seeing no roles populated in the UI after I run the migration 
image

Also, I do prefer the pattern of nesting object attributes (`{role: {id, display_name, description}}`, rather than flattening them out (`role_display_name`). But perhaps it adds complexity in the serializer and for updates from the UI. I think we already have a mix, so not a big deal, but something to consider going forward!
image

mohamedelabbas1996:
> Hmm, I'm still seeing no roles populated in the UI after I run the migration image
> 
> Also, I do prefer the pattern of nesting object attributes (`{role: {id, display_name, description}}`, rather than flattening them out (`role_display_name`). But perhaps it adds complexity in the serializer and for updates from the UI. I think we already have a mix, so not a big deal, but something to consider going forward! image

Hey @mihow , thanks a lot for testing this out!

It’s working for me on the Mount Totumas project using the Panama DB snapshot. Could I try testing it on the specific database snapshot you're using? I'd like to see if I can reproduce the empty roles there!
@annavik, are you able to reproduce this on your end as well?

PR #817, opened 2025-04-29
Adjust default font seen on slow connections

annavik:
Comment from Michael:

> One tiny quirk, I noticed on slow connections that the default font is a serif font, then switches to the Nova font, which is very different. I wonder if the default font can be something sans-serif and a little more similar? No big deal but something I know that you will notice too :slightly_smiling_face:


mihow (approved):
This is definitely an improvement!

PR #829, opened 2025-05-05
Modify docker compose configuration for local development

f-PLT:
## Summary

This PR modifies the default docker compose file and the UI Dockerfile to use a pre-built version of the frontend and improve startup time of the dev compose stack.

### List of Changes

* Adds `docker-compose-frontend-dev.override.yml` file that can be used to override the UI entrypoint command and enable hot reloading of the UI
* Modify the UI local development dockerfile to use a pre-built version of the frontend by default
* Add information documenting these changes in the `README.md` file

### Related Issues

N/A

## Detailed Description

This PR changes the default behavior when initiating a local dev environment. Default configuration now uses a pre-built version of the frontend. 

While this disables hot reloading of the UI, it makes startup time faster when restarting the stack and/or working on the backend components.

An override is provided and described below for those who might prefer doing frontend development using docker instead starting the UI separately from the compose stack. 

### How to Test the Changes

**Running the override for Frontend development:** 

Run the docker compose with the override:
```sh
docker compose -f docker-compose.yml -f docker-compose-backend-dev.override.yml up -d
```

**Standard development**

Running the base docker compose by itself will now run the Frontend from a pre-build version.

```sh
# Build the frontend

(cd ui && yarn install && yarn build)

# Start the whole compose stack
docker compose up -d
```

### Screenshots

N/A

## Deployment Notes

This doesn't affect deployment, as it targets only the local dev environment.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,23 +16,33 @@ Antenna uses [Docker](https://docs.docker.com/get-docker/) & [Docker Compose](ht
     127.0.0.1 minio
     127.0.0.1 django
 ```
+3) The following commands will build all services, run them in the background, and then stream the logs.
+   1) Standard development
+      ```sh
+      docker compose up -d
+      docker compose logs -f django celeryworker ui
+      # Ctrl+c to close the logs
+      ```
+   2) Backend only: will use a pre-built version of the frontend that will not have hot-reloading enabled, but makes startup time faster when restarting the stack.
+      ```sh
+      # Build the frontend (only needed the first time starting the stack, or if the frontend changed since your last build)
+      (cd ui && yarn build)
+      ```
+      ```sh
+      # Run docker compose with the override config
+      docker compose -f docker-compose.yml -f docker-compose-backend-dev.override.yml up -d
+      ```
+      If there's a need to update the frontend while using this override, simply re-build the frontend with the `(cd ui && yarn build)` after modifying the frontend code.
 
-2) The following commands will build all services, run them in the background, and then stream the logs.
 
-```sh
-    docker compose up -d
-    docker compose logs -f django celeryworker ui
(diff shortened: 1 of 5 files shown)


mihow (commented):
This is very helpful, thank you Francis! I am leaning toward making this the default behavior, and then use the override for working on the frontend with Docker. So to work on the frontend, you can either user npm on the host (which is Anna's preference) or Docker. 

Also I suggest we remove the `docs` container from the stack for now. We are not using it yet and it adds overhead & build time.

f-PLT:
Ok, great! I'll make the adjustments tomorrow.

mihow (approved):

mihow:
@f-PLT I noticed a couple things after I merged, sorry I didn't test more thoroughly:

- The README says you need to install Node.js locally & build the frontend before running `docker compose up`. It should be one or the other (use Docker and skip Node locally, or skip Docker and use Node)
- The frontend is still auto-building when I change a file
- The startup time is still takes a few minutes for me, even on a fast machine. Maybe we should discus the purpose and benefits of this change more first!

f-PLT:
Yes, these are all normal:

* It's necessary to install node/yarn because of the base structure of docker file. I can convert it to a full build docker style, but this will probably need we'll need to manually rebuild the whole image everytime there's a change because we don't use versions yet and I'm not sure docker compose will pickup on file system changes and trigger a rebuild by itself. I'm fine with that, but it does start to change the paradigm away from how docker is currently use.
* As to the frontend, you probably need to run `docker rmi` on your current frontend image. I think I had the same problem at one point: Since only the `CMD` was changed, it somehow did properly create the new version of the docker image until I remove my initial one.

PR #836, opened 2025-05-07
Streamline developer experience by making UI load faster by default

f-PLT:
## Summary

Fixes comments from previous PR #829 .

### List of Changes

* Makes the base UI Docker self-sufficient
* Add a try/catch to `vite.config` to identify clearly when a local docker image is used for the frontend

## Detailed Description

As suggested, the UI docker approach is now completely self-sufficient and separated from a developer's local development setup for the UI.

A developer wanting to work on the UI with docker now doesn't need to have Node installed locally at all (although at some point it still is worth it to do so). A developer working locally will have node/yarn installed locally should see their workflow stays the same as before.

The change proposed to `ui/vite.config.ts` is because currently, building the UI requires access to the commit history to tag each build. While copying the `.git` folder at build time could work, it would also have required to change the whole build context of the UI service, remove the `ui/` folder from the main `.dockerignore` and possibly refactor all the other docker files in the project to make sure no adverse effects came of such a change.

Since we are considering reworking the docker files soon anyway, I figured it wasn't worth the trouble right now. It does also have the benefit of clearly signaling that a pre-build docker image is being used and reminds the user to make sure to rebuild the image when necessary.

image

Using the optional compose override still displays the appropriate build number, like before.

### How to Test the Changes

1) Standard development: will use a pre-built version of the frontend that will not have hot-reloading enabled, but will make startup time faster when restarting the stack.

```sh
# Start the whole compose stack
docker compose up -d

# To stream the logs
docker compose logs -f django celeryworker ui
# Ctrl+c to close the logs
```

If there's a need to update the frontend while the compose stack is running, use the following command to rebuild the frontend and load the new changes
(and remember to refresh your browser after!).

```sh
docker compose build ui &&  docker compose up ui -d
```

   2) With Hot Reload UI: Hot reload is enabled for frontend development, but the primary web interface will be slow to load at startup and later restarts.

```sh
# Run docker compose with the override config
docker compose --profile ui-dev up -d --scale ui=0
```

### Screenshots

N/A

## Deployment Notes

Potentially worth to make sure that the changes to `vite.config.ts` do not cause unforeseen problems with the Netlify deployment structure

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Expanded development setup with multi-step build/run instructions, UI hot-reload guidance, log streaming, cleanup steps, and additional access URLs (admin, API docs, object storage) plus port-check notes.

* **Chores**
  * Reworked local services: removed docs service, added a dedicated UI development service and updated UI runtime to serve a preview on port 4000.
  * Excluded build artifacts from UI build context and made startup behavior more resilient when source metadata is unavailable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Antenna uses [Docker](https://docs.docker.com/get-docker/) & [Docker Compose](ht
       docker compose build ui && docker compose up ui -d
       ```
 
-   2) With Hot Reload UI**: Hot reload is enabled for frontend development, but the primary web interface will be slow to load when it first starts or restarts.
+   2) **With hot reloading UI**: Hot reload is enabled for frontend development, but the primary web interface will be slow to load when it first starts or restarts.
       ```sh
       # Stop the production ui first, then start with ui-dev profile
       docker compose stop ui


mihow (approved):
@f-PLT I am finally merging this! However we just introduced a default override file to enable the python debugger, and the command is quite cumbersome to use both, if that works at all. So I decided to try using the newer `--profile` flag here instead to enable the hot reloading UI. 

`docker compose --profile ui-dev up -d`

Perhaps we can give this a try. 

Then later on I still want to rework the compose files so that it's easy to run all of Antenna locally with prebuilt containers, and opt-in for dev containers. It looks like you can set the override files and profiles with env vars too. So if the CLI commands are getting cumbersome, we can suggest a single .env var to toggle the type of environment.  or a Makefile! :) 

For the meantime, this greatly speeds up local dev, thank you so much. I'm happy to drop that docs container for a while as well.

PR #841, opened 2025-05-10
Improve bulk identification workflow

### LINKED ISSUE #831: Show more recently used Taxa during bulk ID
Description from user on Slack:

>  when validating on Antenna, I would like to see more than one species in the 'recent' (bottom left in the square).  Most of the time, it's always the same 4-5 species mixed together.  Would be much faster to go through.

Image

This may be a backend only task! If it's just about increasing the num items returned. But needs investigation of where the recent items are coming from.

mihow:
## Summary

Improvements to the bulk ID workflow. Changes by @annavik 

### List of Changes

- Extend list of recently used taxa from 1 to 5 
- Wait with invalidating query for occurrence data until all id requests has completed. I think refetching data just once might help with some confusions. It could also be an optimization that will avoid requests queuing up a bit?

### Related Issues

- Closes #831 
- Addresses issue that 2 identifiers experience on a regular basis: IDing more than 2-3 occurrences at a time does not always update the list view.

## Detailed Description

.

### How to Test the Changes

.

### Screenshots

.

## Deployment Notes

.

## Checklist

### DIFF
--- a/ui/src/data-services/hooks/identifications/useCreateIdentification.ts
+++ b/ui/src/data-services/hooks/identifications/useCreateIdentification.ts
@@ -15,7 +15,10 @@ const convertToServerFieldValues = (
   comment: fieldValues.comment,
 })
 
-export const useCreateIdentification = (onSuccess?: () => void) => {
+export const useCreateIdentification = (
+  onSuccess?: () => void,
+  invalidate: boolean = true
+) => {
   const { user } = useUser()
   const queryClient = useQueryClient()
 
@@ -29,8 +32,10 @@ export const useCreateIdentification = (onSuccess?: () => void) => {
         }
       ),
     onSuccess: () => {
-      queryClient.invalidateQueries([API_ROUTES.IDENTIFICATIONS])
-      queryClient.invalidateQueries([API_ROUTES.OCCURRENCES])
+      if (invalidate) {
+        queryClient.invalidateQueries([API_ROUTES.IDENTIFICATIONS])
+        queryClient.invalidateQueries([API_ROUTES.OCCURRENCES])
+      }
       onSuccess?.()
     },
   })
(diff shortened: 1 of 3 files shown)


mihow:
@annavik I think there is an issue with the front end cache invalidation. When I bulk ID I don't see any updates in the list view or detail view. Even though the requests were successful. 

https://github.com/user-attachments/assets/10fc00ae-06d8-4d60-bf17-4d47bda63abe

annavik:
> @annavik I think there is an issue with the front end cache invalidation. When I bulk ID I don't see any updates in the list view or detail view. Even though the requests were successful.
> 
>  Bulk.ID.cache.invalidation.Screen.Recording.2025-05-11.at.9.55.12.AM.mov

Hey @mihow , thanks for testing!

In the clip it looks like you have sorted your list of occurrences based on score. At the moment, the backend will set the determination score to 1 when a human has been involved, which means when the list is updated you might see different occurrences with this sorting applied. This is also why nothing is selected after the update in your clip.

I understand how this could be a bit confusing, I think it could be interesting to discuss the backend behavior here. Would it make sense to use the score field for machine predictions only, not update it based on human identifications? We could still update verification status, but without changing the score? Probably a bigger change to discuss with Max and others...

To summarize, I think what you saw when testing is not a cache invalidation problem and the behavior is actually expected given the current backend logic. Please test again though with different sort settings and let me know if I got something wrong! :)

mihow:
Oh my gosh you are so right! I can't believe that got me again. In the ood branch we can sort by `updated_at` now, which is helpful to see what's going on in this case. I will merge this into main and the ood branch.

mihow:
Merged this into both deployment!
https://ood.antenna.insectai.org/projects/2/occurrences?ordering=-updated_at

I did notice that if you click Confirm, those species do not show up in the Recent menu. Only if you search for an ID or use a quick action button, those will show up in the Recent menu. Is that data coming from the backend?

PR #847, opened 2025-05-14
Make it possible to suggest ID in bulk

### LINKED ISSUE #824: Suggest ID in bulk
## Summary

For the OOD project, it would be helpful being able to suggest ID for occurrences in bulk. In this way, we can move occurrences between clusters.

This feature could also be helpful for Antenna users in general and we suggest we include it as part of the main app later.

## Implementation Details

The idea is to extend the current quick action feature with a "Suggest ID" option.

Image

annavik:
## Summary

In this PR we make it possible to suggest ID in bulk. We also replace the old taxonomy search component with the new one, to streamline and simplify development.

### List of Changes

* Replace old taxa search with new component
* Make it possible to suggest ID from inline popover
* Make it possible to suggest ID in bulk

### Related Issues

Closes #824 

### How to Test the Changes

Select multiple occurrences from list view. Select a taxon from the popover. Add a comment (optional) and click save. See how labels are updated.

Also, make sure the label that was just suggested will show up under the "recent" section for ID quick actions.

### Screenshots
Screenshot 2025-05-14 at 19 50 13
Screenshot 2025-05-14 at 19 50 51


mihow (approved):
Tested and works well! I think this will be a very welcome change for current identifiers. I am comfortable merging right into main. I also added the `updated_at` column to occurrences to this PR and made it the default sort order. I think this makes it much easier to see what you have IDed or changed.

PR #848, opened 2025-05-15
Handle undefined OOD scores in UI

annavik:
If the OOD score is undefined or null, we want to present it as "n/a" in UI, not 0.

Defined:
Screenshot 2025-05-15 at 08 55 03

Not defined:
Screenshot 2025-05-15 at 08 55 51


mihow (approved):
It works! This is a nice reason to keep both deploy previews active. One to test on the `ood` environment and one on the preview environment / `main`.

PR #850, opened 2025-05-15
Add unknown species filter control

annavik:
In this PR we add a dedicated filter control for unknown species. We also extend the backend filtering with this new query param.

This is how I suggest we present it, let me know if you have feedback on copy!
Screenshot 2025-05-15 at 14 53 56


mihow:
What do you think about using this as a catch-all filter component for fixed value queries? the current ones I'd like to use are:
- include unobserved
- missing reference image
- unknown species
- known species (optional)
- has disagreement (for the occurrences table)

mihow (approved):

ami/main/api/views.py
```diff
@@ -1187,6 +1187,24 @@ def filter_queryset(self, request, queryset, view):
         return queryset
 
 
+class TaxonUnknownSpeciesFilter(filters.BaseFilterBackend):
```
mihow, line 1190: Great work implementing this filter! This follows all of my favorite conventions.

PR #854, opened 2025-05-19
Update pagination logic

annavik:
## Summary

@mihow noticed that if a user is inspecting records from a different page than then first page, the page settings will remain when the filtering is updated. This can be a bit confusing, especially if data is no longer visible for the current page.

In this PR, we reset the current page when filters are modified from the UI. We also update the empty state component to consider page settings. To see no data because the current page is not the first page should be an edge case now, but it can still happen if page is set from URL.

### Screenshots
No data and page is not first page (see URL):
Screenshot 2025-05-19 at 11 23 59

No data and filters are set:
Screenshot 2025-05-19 at 11 27 41

No data:
Screenshot 2025-05-19 at 11 28 23


mihow (approved):
Tested and works!

PR #856, opened 2025-05-23
Add best score filtering and column settings support

annavik:
## Summary

Add support for:
- Present best determination score for taxa in UI
- Add best score filter for taxa
- Add column settings support in taxa view

## Screenshots

Screenshot 2025-05-23 at 00 03 10

### DIFF
--- a/ami/main/api/serializers.py
+++ b/ami/main/api/serializers.py
@@ -786,6 +786,7 @@ class TaxonSerializer(DefaultSerializer):
             "cover_image_url",
             "cover_image_credit",
             "unknown_species",
+            "best_determination_score",
         ]
 
     def get_cover_image_url(self, obj):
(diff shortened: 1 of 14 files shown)


mihow (changes requested):
Tested and most everything works! The only thing I am noticing is that the filter isn't applied until you move the slider, even if there is a value set in the preferences or from another view. The slider is updated with the current preferences, but the API request is not.

image

annavik:
> Tested and most everything works! The only thing I am noticing is that the filter isn't applied until you move the slider, even if there is a value set in the preferences or from another view. The slider is updated with the current preferences, but the API request is not.
> 
> image

Oh good catch! It was a key typo by me for the default filter settings, fixed here: https://github.com/RolnickLab/antenna/pull/820/commits/d928ba3ba554f69d67ff4d39bb599c151d0c943e

PR #858, opened 2025-05-25
[Draft] Filter images before clustering

mihow:
Experimental work by @Yuyan-C for the Panama research field trip in May 2025.

See comments below about the suggested plan for this work. Some parts can be integrated into the Antenna `main` branch and deployed to live, some parts depend on the OOD features which will stay in the experimental branch `deployments/ood.antenna.insectai.org`.


mihow:
I suggest we open a new branch from main and pull in several of these filters for the regular classification task (and keep clustering it's own experimental branch)

Filters:
- Filter crops that are too small
- Filter crops that are too blurry
- Filter crops on the edge

These can be first implemented in a post-processing job, which can be run on existing predictions. It will add a prediction to existing occurrences and update their determination. If the occurrence does not pass the filters, a new prediction should be added "Not identifiable".

Then we can consider when & how it is run in the existing pipeline. I suggest this pattern for all post-processing steps (regional class masking, rank predictions, etc)

mihow:
I suggest we remove the functions that depend on OpenCV & the transformers library (they can be moved to an experimental branch) and then merging this into the `deployments/ood.antenna.insectai.org` branch. Then we can close this PR and continue cherry picking features from the `deployments/ood.antenna.insectai.org` into main.

PR #830, opened 2025-05-06
Add support for Taxa Tags

mohamedelabbas1996:
## Summary
This PR introduces backend support for managing and assigning tags to taxa. Tags are project-scoped and can now be created, assigned, and filtered by project.

### List of Changes
-  Added  `Tag` model to keep track of tags.
 - Added Tag admin model to allow tag creation and management through the admin interface.

- Added a custom `assign_tags` action in TaxonViewSet to assign tags to taxa via API.

- Implemented TagViewSet with full CRUD functionality (create, read, update, delete).

- Enabled filtering of tags by `project_id` in both TagViewSet and TaxonViewSet.

- Updated TaxonSerializer to expose `tags`, showing tags assign to the taxon across all projects.

- Updated TaxonSerializer to expose `project_tags`, showing only tags relevant to the active project.

- Integrated project-specific tag prefetching in TaxonViewSet.

### Related Issues

#825 

## Detailed Description

This feature enables users to create descriptive, project-scoped tags  and assign them to taxa via the API. The assignment is handled through an `assign_tags` action on the `TaxonViewSet`, allowing users to attach tags to a specific taxon within the context of a project.

- The `tags` field in the taxon detail response  always includes all tags assigned to that taxon, regardless of project.
- When the taxon detail endpoint is filtered with a `?project=<id>` query parameter, the `project_tags` field  is populated with a list of tags belonging to the specified project.

- **Filter by tag**:  
  `GET /api/v2/tags/<tag_id>/` — the `taxa` field in the tag detail response shows all taxa currently assigned to that tag.

- **Remove a tag from a taxon**:  
  Use the `assign_tags` action with a new list of tag IDs that excludes the tag you want to remove.  
  Example:
`POST /api/taxa/<taxon_id>/assign_tags/`
```json
  
  {
    "tag_ids": [1, 2, 5]    
}
  ```
`POST /api/taxa/<taxon_id>/assign_tags/`
  ```json
  
  {
    "tag_ids": [2, 5]        // previously had [1, 2, 5]; tag 1 is now removed
}
  ```
### Request/Response Format

**Endpoint**  
`POST /api/v2/taxa/<taxon_id>/assign_tags/`

---

**Request Body**
```json
{
  "tag_ids": [1, 2, 5]
}
```
**Response**
```json
{
  "taxon_id": 42,
  "assigned_tag_ids": [1, 2, 5]
}
```

### How to Test the Changes

1. Create a new tag from the admin page.

[...]

Issue #825, opened 2025-05-02
Make it possible to add tags to taxa


annavik:
## Summary

On the OOD workflow demo, one feature request that got mention was being able to review clusters. In this process, it would be helpful if we could tag clusters.

## Detailed Description
Example of tags:
- Reviewed
- Most wanted
- Also wanted
- Found
- Collected

Sub features needed:
- Being able to see tags
- Being able to filter by tag
- Being able to add a tag
- Being able to remove a tag

## Implementation Details
We think we can implement this in a similar way as taxa lists (see https://github.com/RolnickLab/antenna/pull/777), except we also need to make it possible to assign this from UI.

annavik:
Fixed with https://github.com/RolnickLab/antenna/pull/830 and https://github.com/RolnickLab/antenna/pull/828.

PR #892, opened 2025-07-04
Make it possible to add tags to taxa

annavik:
### Summary

In this PR, we cherry pick the tag feature from the [OOD branch](https://github.com/RolnickLab/antenna/pull/820). The changes includes both frontend and backend updates.

### Related Issues

Original issue: https://github.com/RolnickLab/antenna/issues/825 
Original PR:s: https://github.com/RolnickLab/antenna/pull/828 and https://github.com/RolnickLab/antenna/pull/830

### Comments

- For the original feature, we added a set of global tags to the database as a migration step. I think for the prod app, we can skip this? I excluded this step from code (see commit c6106f6ec76a0e83a8976b09c04cbb8a01f97d35).
- We want to start by introducing tags for selected projects only. To make this possible, we will put tags behind a feature flag. To begin with, feature flags will be hard coded on the backend side and tags will be disabled for all projects. This means we can merge the code, but introduce tags to users at a later point.

### DIFF
--- a/ui/src/components/taxon-tags/tags-form.tsx
+++ b/ui/src/components/taxon-tags/tags-form.tsx
@@ -107,8 +107,7 @@ const FormRow = ({
       htmlFor={`tag-${tag.id}`}
       className="flex items-center gap-2 body-small"
     >
-      <div className="w-2 h-2 rounded-full bg-primary" />
-      {tag.name.toLowerCase()}
+      {tag.name}
     </label>
   </div>
 )


ami/main/models.py
```diff
@@ -125,6 +125,7 @@ class Project(BaseModel):
     image = models.ImageField(upload_to="projects", blank=True, null=True)
     owner = models.ForeignKey(User, on_delete=models.SET_NULL, null=True, related_name="projects")
     members = models.ManyToManyField(User, related_name="user_projects", blank=True)
+    feature_flags: dict[str, bool] = {"tags": False}  # @TODO return stored feature flags for project
```
annavik, line 128: @mihow here is where we can tweak the hard coded feature flags

mihow (changes requested):
I added a true feature_flags field to the model, and I see the `tags: true` in the API details for my test project, but I can't get any tag features to render in the UI. I added some tags for the project as well. I'm not sure where the project details are being fetched.

Issue #873, opened 2025-06-17
Make score threshold a project setting


annavik:
More info to be added. Similar to settings for #868 ?

annavik:
Closing in favor of #927

Issue #868, opened 2025-06-09
Exclude "Not a Moth" by default


annavik:
In some cases, data of type "Not a Moth" is bloating our views, making the actual moth data more difficult to inspect. We should probably keep the "Not a Moth" data possible to inspect on demand (in some scenarios this is also interesting data), however by default it should probably be filtered out?

**Implementation ideas:**
- Project setting
- Taxa to exclude/hide by default (from Occurrences list, from charts)
- Consider using a Taxa List
- Should hide recursively - all children under Not a Moth
- Need a way to _show_ the hidden taxa on demand (filter or checkbox?)

Examples of when the "Not a Moth" data is taking over a bit:
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/57c5267e-7979-42db-bc07-87341fc2ea24" />
 
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/15ed4388-56e2-4be8-bea5-919c4d4a5fc5" />

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/bbffde7b-2f6e-42c1-81a4-4923021ac9ae" />

annavik:
I'm thinking something like this could be a start? A view for more detailed project configuration.
<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/d164f63d-4575-4099-8c54-f2b4c6a67ac3" />

For the "root taxon", my idea is to use this for default filtering everywhere. We still want to make it possible to show hidden taxa on demand in a few selected views, but instead of all taxa being the default, we could fallback to the root taxon specified here. Maybe something similar for score?

When we introduce project settings like this, could also be a good time to include related configuration, for example make it possible to specify reject options for a project. Let's talk more in person!

Issue #237, opened 2023-09-08
Some sessions are incorrectly grouped


mihow:
For example, in Event # 648, there is a long gap between SourceImage 2595550 and 2595551, but they were not broken into different days. Check the timestamps of those images and investigate the `group_images_into_events` function.

https://app.preview.insectai.org/projects/4/sessions/648?capture=2595452
https://app.preview.insectai.org/projects/4/sessions/750?capture=2283677
https://app.preview.insectai.org/projects/4/sessions/654?capture=2597301

<img width="513" alt="image" src="https://github.com/RolnickLab/ami-platform/assets/158175/f740734d-cc88-4f5e-8805-d2996433e78a">

mihow:
Another one here:

https://deploy-preview-295--ami-web.netlify.app/projects/1/sessions/60?capture=141321
![image](https://github.com/RolnickLab/ami-platform/assets/158175/854ff4c8-f39a-48a8-9c9e-655b1bef55a2)

PR #893, opened 2025-07-04
Setup default filters for project

annavik:
## Summary

In this PR we setup default filters for projects. This will help us present more relevant data for users and simplify view settings. The PR is in draft since we need to include BE changes and also a few remaining FE tweaks.

### List of Changes

* Setup new settings view "Default filters"
* Setup a form to configure following filters for a project
  * Score threshold
  * Include taxa
  * Exclude taxa
* Exclude view controls for score filtering (for sessions, occurrences and taxa)

### Sub Tasks
* [FE] Setup settings page for default filters ✅
* [BE] Update the project model to handle default filters ✅
* [BE] Update the project API to expose default filters ✅
* [BE] Apply the default filters backend side when data is returned ⏳
  * Stats bars and charts
  * List views
  * Capture view
  * Numbers in columns
  * Taxa search
* [FE] Connect settings form with backend ✅
* [FE] Update quick actions to use exclude taxa values  ✅
* [FE] Make it possible to disable default project filters when needed ⏳

### Related Issues
#873
#868 
#237 
#906 (the time gap between monitoring sessions should be configurable by users)

### Screenshots

Screenshot 2025-07-04 at 16 13 51


ui/src/data-services/hooks/projects/utils.ts
```diff
@@ -13,5 +13,8 @@ export const convertToServerFormData = (fieldValues: any) => {
     data.append('image', '')
   }
 
+  // TODO: Remove console log and include default filters as form data
```
annavik, line 16: The form is working, we just need to pass the values here when the backend is ready.

ui/src/pages/project-details/default-filters-form.tsx
```diff
@@ -0,0 +1,216 @@
...
+  isSuccess?: boolean
+  onSubmit: (data: DefaultFiltersFormValues) => void
+  project: Project
+}) => {
+  const {
+    control,
+    handleSubmit,
+    setError: setFieldError,
+  } = useForm<DefaultFiltersFormValues>({
+    // TODO: Replace default values with stored values for project
```
annavik, line 59: When default settings are returned with the project we should replace the placeholder values. For returned taxa, the type `{ id: number, name: string}[]` is helpful (rather than a flat list of ids).

mihow:
Note: Consider adding a size filter here as well. Start with pixel percent, but if we add a size calibration for each device type (physical dimensions of the size of the target) then the filter can be physical units (centimeters). The goal being to automatically hide any moths/detections that are under X square centimeters.

annavik:
Closing in favor of #929 and #930

PR #895, opened 2025-07-04
Advanced collection filters 

mihow:
## Summary

Enable several needed filter methods for source image collections to the UI, using raw IDs.

image

### List of Changes

- Exposes `deployment_ids` filter to the UI - very helpful, but currently requires Django admin to use
- New filter for `station_ids` - requested by partner
- New filter for `event_ids` - required for testing tracking
- Method for specifying a custom serialization & deserialization for form fields in the UI
- New "Advanced Filter" section in the source image collection form

### Implementation Details

This is a simple way to allow filtering by related data using existing UI components. However a custom validator was added for comma separated integers, as well as a way to serialize & deserialize JSON lists to & from the API.


ui/src/pages/project/entities/details-form/collection-details-form.tsx
```diff
@@ -107,7 +107,7 @@ const config: FormConfig = {
   },
   'kwargs.deployment_ids': {
     label: 'Station IDs',
-    description: 'Enter comma-separated integers (e.g., 1, 2, 3).',
+    description: 'Enter comma-separated numbers (e.g., 1, 2, 3).',
```
annavik, line 110: We use "numbers" for other description fields, that is why I changed this. Also because I think it sounds a bit more simple! :)

mihow: Thank you!

mihow:
> Thanks Michael, this is great! I just tweaked the labels a bit.
> 
> Some thoughts:
> 
> * I think we should first introduce the station filters and see if that is enough, before we add site filtering. I want to keep the form as simple as possible and I suspect filtering by station might solve the problem for the partner :)
> * I think the advanced filter section is making the form a bit too complicated. I like the idea and I think we should split up filters this way moving forward, but I think advanced filters should only be visible on demand. Since that requires some more work, I wonder if I good start could be we just add a row to the current filter section? As the next step, I would like to collapse this section a bit and add a "Show more" button?
> 
> Screenshot 2025-07-05 at 15 07 11

Thanks @annavik! I agree with you and I look forward to a Show More button here. The Site filter will be really convenient when we add it because you only need to specify one Site instead of looking up all the Station IDs. But I agree that just Station IDs & Session IDs keeps it more flexible and just one row of new fields. 

Another option (for a follow-up PR) is to add a general config with a JSON editor. Then we don't have to restrict any functionality and encourage users to switch away from Antenna back to their own scripts, etc. That would be my preference across the site (hidden by default). I think this is a common pattern, where the UI provides a visual config builder for basic things, but then lets you switch to the advanced/code mode to edit some syntax directly (QGIS, NewRelic, LabelStudio, Overleaf, Weights & Biases, any WYSIWYG editor, are some examples that come to mind).

PR #896, opened 2025-07-05
Tweak collection views

annavik:
## Summary

Just adding some small tweaks to the collection views.

### List of Changes
* Align form labels ("images" -> "captures")
* Make it possible to sort collections by ID (and use this as the default sort setting)
* Make it possible to specify column settings for collections (I noticed the action buttons was getting hard to see)
* Use Nova button component for "Register new" actions in other views, to streamline

### Screenshots

Screenshot 2025-07-05 at 14 46 03


ami/main/api/views.py
```diff
@@ -644,6 +644,7 @@ class SourceImageCollectionViewSet(DefaultViewSet, ProjectMixin):
     ]
     filterset_fields = ["method"]
     ordering_fields = [
+        "id",
```
annavik, line 647: Sort by ID requires backend deploy at some point, but nothing urgent

mihow: Will do! Thanks!

PR #898, opened 2025-07-11
[Draft] Fix deployment sessions creation &  counts 

### LINKED ISSUE #872: Sessions not auto updated when captures are synced
Current solution is we sometimes manually trigger session updates from admin (in practice means users will run into problems when they try inspect data and have to reach out to us). Creating a placeholder ticket for this. Maybe we have an exciting one?

mohamedelabbas1996:
## Summary

This PR ensures sessions are created and updated when captures are synced.

### List of Changes

TBD

### Related Issues
Closes #872  

## Detailed Description

TBD

### How to Test the Changes
TBD

### Screenshots
TBD

## Deployment Notes
TBD
## Checklist

### DIFF
--- a/ami/main/models.py
+++ b/ami/main/models.py
@@ -610,6 +610,8 @@ class Deployment(BaseModel):
             job.progress.add_stage("Update deployment cache")
             job.update_progress()
 
+        # Group images into events
+        group_images_into_events(deployment)
         self.save()
         self.update_calculated_fields(save=True)
 


ami/main/models.py
```diff
@@ -1091,6 +1094,48 @@ def group_images_into_events(
...
+        logger.debug(f"No source images found for deployment {deployment}.")
+        return False
+
+    if latest_event_time is None:
+        # No events exist yet — need to group
+        logger.debug(f"No events found for deployment {deployment}.")
+        return True
+
+    # Compare timestamps
+    needs_update = latest_image_time > latest_event_time
```
mihow, line 1128: I think this will always be true. Since the timestamps are not updated. There will always be source images created after the event was created.

mihow: Two other options
- Run the grouping function without actually saving sessions and use that output to compare the images in the current sessions to the would-be sessions.
- Add a `last_checked` field to the Event (session) model and update that after regrouping.

I'm okay to postpone those checks and just run the session regrouping at the end of the sync for now.

mihow: The regrouping "audit" function would need to be efficient - just fetch the timestamp field from all source images from the deployment and determine the would-be sessions. 

See
https://github.com/RolnickLab/antenna/blob/147183b25b5db0752183bd82505240be623f6bfa/ami/utils/dates.py#L98-L101

mihow:
Remember the create time for an event is the first time the object was
created. Same for captures.

Perhaps here is what you meant!

SourceImage.timestamp > Event.end_datetime

Then you’re comparing the time that a capture was captured with the current
date range of the session.

Another idea, What if you check and see if the source image does not belong
to an event yet?

SourceImage.event = None

SourceImage.filter(deployment=1, event=None)

Or event__isnull=True

annavik:
Thank you so much for this update, it will be such a relief to have this working! 🙏

Issue #220, opened 2023-09-02
Label Studio integration


mihow:
- Ability to create sample set in the AMI platform
- Ability to generate individual task JSON files and write them to the tasks bucket

- Create new object detection tasks in Label Studio based on Source Images in the AMI platform
- Listen for annotated Source Images from Label Studio (manually labeled bounding boxes)
- Create Detection objects in the AMI platform based on human annotations
- Create binary classification tasks in Label Studio
- Listen for complete moth/non-moth annotations
- Update detections with binary labels in AMI Platform
- Create species classification tasks in Label Studio
- Listen to complete species tasks
- Update detections with species labels in AMI Platform

PR #904, opened 2025-07-22
Fix incorrect session grouping 

### LINKED ISSUE #237: Some sessions are incorrectly grouped

mohamedelabbas1996:
## Summary

This PR fixes the issue of incorrectly grouped sessions (events) by refactoring and improving the image grouping logic when syncing a deployment. 

### List of Changes

- Refactored `group_images_into_events` to support a new `use_existing` flag:
  
  - If `use_existing=True`:
    - Only new images (images not assigned to an event)  are grouped.
    - Groups are merged into existing events based on time proximity and overlap.
  
  - If `use_existing=False`:
    - All images in the deployment are re-grouped.
    - If a new group exactly matches an existing event (same start and end time), it reuses the event.
    - Otherwise, new events are created.

- Removed the `group_by` field from the `Event` model.
- Added two admin actions:
    - Fix Sessions: allows to fix incorrectly grouped sessions. 
    - Remove from Event: allows admins to manually remove selected source images from selected events.

### Related Issues

Closes #237

## Detailed Description

Previously, session grouping relied on the `group_by` field, which reused an existing event if a group had the same start date. This caused issues when images taken on the same day—but far apart in time—were incorrectly grouped into a single session, even though timestamp-based grouping split them into multiple groups. Since all groups shared the same start date, they got assigned to the same event due to `group_by`. 
This PR fixes this  issue  by improving the `group_images_into_events` function. It introduces a `use_existing` flag to control the behavior: when `use_existing=False`, all deployment images are regrouped; when `True`, only new images (those not yet assigned to an event) are processed. Images are grouped based on their timestamps using a `max_time_gap` threshold, and then each group is either merged into an existing event (if overlapping or close enough and `use_existing=True`) or assigned to a new event (or an existing one if its start and end time exactly match the group). The `group_by` field is removed from the `Event` model, and an admin action is added to help fix  incorrectly grouped sessions. Additionally, cached fields in event related models (e.g. `Occurrence`) are updated accordingly.

### Screenshots

N/A

## Deployment Notes

N/A

## Checklist

## Related Issues

- #1157 - Expose session regrouping as a job stage when syncing images
- #906

## Summary by CodeRabbit

* **New Features**
  * Admin actions to dissociate related objects and to regroup/fix event sessions.
  * New management command to audit and repair event clustering with previews and detailed reports.

* **Improvements**
  * More robust event grouping with merging of overlapping/nearby events and better reuse/refresh of event data.
  * Admin search extended to include event start dates.

* **Tests**
  * Expanded tests covering grouping, merging, regrouping, and syncing scenarios.

* **Migrations**
  * Schema change replacing group-based uniqueness with start/end uniqueness and rollback safety.


ami/main/models.py
```diff
@@ -1016,60 +1003,82 @@ def group_images_into_events(
             f"Found {len(values)} images with the same timestamp in deployment '{deployment}'. "
             f"Only one image will be used for each timestamp for each event."
         )
+    # Get all images
+    image_qs = SourceImage.objects.filter(deployment=deployment).exclude(timestamp=None)
+    if use_existing:
+        # Get only newly added images (images without an event)
+        image_qs = image_qs.filter(event__isnull=True)
+
+    images = list(image_qs.order_by("timestamp"))
```
mihow, line 1012: You likely don't have to evaluate the queryset yet with `list(image_qs)`. You can check if images are found with `images_qs.exists()`, which is efficient for large datasets.

ami/main/models.py
```diff
@@ -1016,60 +1003,82 @@ def group_images_into_events(
...
-            f"From {start_date.strftime('%c')} to {end_date.strftime('%c')}."
-        )
+        group_start, group_end = group[0], group[-1]
+        group_set = set(group)
+        group_image_ids = [img.pk for img in images if img.timestamp in group_set]
+
+        event = None
+        if use_existing:
+            # Look for overlap or proximity
+            for existing_event in existing_events:
```
mihow, line 1037: If you are looping over a queryset, you can do `for existing_event in events_qs`, which supposedly avoids loading the whole queryset result into memory. Sometimes you need to convert to a list so you can index the list like events[3], but often you never need to convert to a list.

ami/tests/fixtures/main.py
```diff
@@ -416,3 +461,5 @@ def create_local_admin_user():
     email = os.environ.get("DJANGO_SUPERUSER_EMAIL", "Unknown")
     password = os.environ.get("DJANGO_SUPERUSER_PASSWORD", "Unknown")
     logger.info(f"Test user credentials: {email} / {password}")
+    password = os.environ.get("DJANGO_SUPERUSER_PASSWORD", "Unknown")
```
mihow, line 464: Was this intentional?

ami/tests/fixtures/main.py
```diff
@@ -162,6 +162,51 @@ def create_captures(
     return created
 
 
+def create_captures_in_range(
```
mihow, line 165: This looks helpful, thanks!

ami/main/models.py
```diff
@@ -1016,60 +1003,82 @@ def group_images_into_events(
...
+            # Look for exact match
+            event = Event.objects.filter(
+                deployment=deployment,
+                start=group_start,
+                end=group_end,
+            ).first()
+
+        if event:
+            if use_existing:
+                # Adjust times if necessary (merge)
```
mihow, line 1058: Is this check necessary? I think you just checked if and existing event has the exact start & end time. Perhaps you meant to do an OR query? If an existing event has _either_ the same start or end time as the group.

If there is an existing event with exactly the same start AND end time (for same deployment), then I don't think we should check for `use_existing`. Just re-use those without question.

mihow:
I tested on the sessions in the Mothra deployment in project 18 and was able to repair the existing sessions:

before:
image

after:
image
image

you can see it broke the two long sessions into smaller chunks

mihow:
I pushed a change with my suggested action for removing source images from existing events. I also noticed that Occurrences have a cached field that keeps track of the event as well, so this needs to be updated in our grouping methods. There are other ways to keep occurrences in sync, but they will likely be per-occurrence or per-image update, whereas this can update more at once.

mihow (changes requested):
@mohamedelabbas1996 I just brought this up-to-date with main and rebased the migration and fixed one test. I noticed there are type errors for `event.end` since the end time can be `None` if the event is on-going. We should handle these None cases, or if that adds too much complexity, we can start requiring an end time on the model, and set it to the last capture's timestamp (and perhaps use another method to detect ongoing events, like if the end timestamp is within the time gap of the current real-world time).


image

mihow:
Hi @rhine3!

We have this fix for events/sessions that are incorrectly grouped (sessions that span multiple days, or a single session that contains multiple short sessions). I am testing the fix on a recent db snapshot and seeing how it will affect existing deployments & sessions. Below is the output for one deployment in your project. It seems that most changes are about splitting an event into multiple events - [one main 7hr event and then some random 1 minute test sessions where the camera was turned on briefly](https://antenna.insectai.org/projects/84/sessions/4400). This will not be run automatically when we deploy it, but it will be run next time your deployment images are synced. Would this change throw off your existing analysis, or do you welcome it? Thanks in advance for your feedback.


**Summary for a single deployment** ([#220 LEPS-033_Box1](https://antenna.insectai.org/projects/84/sessions?deployment=220&ordering=-duration))
```
INFO 2025-09-01 19:57:36,790 models 1 136098806425408 

Finished grouping 7636 images into 5 events for deployment '#220 LEPS-033_Box1'
  Events: 3 -> 5
  Captures: 7636 -> 7636

  BEFORE Statistics:
  Count: 3
  Duration: avg=8.8h, std=3.1h, range=7.0h-12.3h
  Captures: avg=2545, std=47, range=2518-2600

  AFTER Statistics:
  Count: 5
  Duration: avg=4.2h, std=3.8h, range=0.1h-7.0h
  Captures: avg=1527, std=1357, range=38-2518
  New events (3):
    - Event 7037: 2024-09-02 15:11:49 to 2024-09-02 15:18:00 (0.1 hours, 38 captures)
    - Event 7038: 2024-09-02 17:44:50 to 2024-09-02 17:52:00 (0.1 hours, 44 captures)
    - Event 7039: 2024-09-02 20:30:30 to 2024-09-03 03:30:00 (7.0 hours, 2518 captures)
  Modified events: None
  Unchanged events: 2
  Deleted events (0 captures): 1
    - Event 4400: 2024-09-02 15:11:49 to 2024-09-03 03:30:00 (2600 captures) [DELETED]

  Day-to-day analysis:
    Days with multiple events AFTER regrouping: 1
      2024-09-02: 3 events
        - 15:11 to 15:18 (0.1h, 38 captures)
        - 17:44 to 17:52 (0.1h, 44 captures)
        - 20:30 to 03:30 (7.0h, 2518 captures)

  Multi-day event analysis:
    (Multi-day = spanning MORE than 2 calendar days)
    (Normal overnight monitoring spans exactly 2 days)
```

**All deployments in project 84**
```
Summary:
  Total events: 158 -> 215
  Total captures: 384507 -> 389334
  Events created: +57
```

rhine3:
Realized I replied to you directly but never posted it here - yes, I HUGELY welcome this change! I do want to exclude the 1min camera turn-on events from my analysis.

mihow:
# Notes from Claude's review of #904

Claude ran a structural review while bringing the branch up to date with main. A couple of real bugs surfaced (already fixed in the merge commit), plus a few larger items worth a quick discussion before merge.

## Fixed in the merge commit

These were quietly broken and the diff alone wouldn't have caught them. The merge commit (`5e56805a`) handles all of them — calling them out so the reasoning is visible.

- **24h event-duration cap from #1268 wasn't being honored.** That cap was added to stop continuous-monitoring deployments from coalescing months of captures into one mega-event (prod deployment 489 had one event spanning Apr→Jul 2023, 355k images). Without threading `max_event_duration` through `group_images_into_events` and `merge_with_following_events`, the new merge logic would have re-introduced that bug. Now passed through both the `merge_with_following_events` helper and the inline existing-event extension path.
- **Occurrence.event_id realignment from #1268 was being undone.** PR #904 had a per-event `Occurrence.objects.filter(detections__source_image__event=event).update(event=event)` inside the loop. That correctly updates occurrences that *gain* a new event but doesn't re-derive occurrences whose source images *left* an event — so they'd silently keep pointing at stale events. Replaced with main's deployment-wide Subquery realignment.
- **N+1 query inside the inner loop.** `for existing_event in existing_events_qs: existing_event.refresh_from_db(...)` ran once per existing event per group → O(groups × events) DB hits. On a deployment with 200+ existing events that's a lot of queries. Snapshotted `existing_events` once outside the loop and pass it as a plain list.
- **`assert event.end is not None` as a production guard.** `assert` gets stripped under `python -O`, so these silently disappear in prod builds. Replaced with explicit `None` checks.
- **Duplicate `audit_event_lengths(deployment)` call** at the end of `group_images_into_events`. Pure dupe — removed.
- **`dissociate_related_objects` had a no-op `.filter()`** (`events = self.filter().distinct()`). The `.filter()` does nothing without args; only `.distinct()` matters. Cleaned up. Also tightened the docstring — it claimed "clear all related objects" but only touches `SourceImage.event` and `Occurrence.event`, which is the right scope but worth saying so.

## Worth discussing before merge

These aren't bugs, more about scope and shippability. Open to your call:

1. **Scope: this PR is doing four things in one branch** — (a) the algorithm rewrite, (b) `group_by` field + migration removal, (c) two new admin actions, (d) a 710-line `recalculate_events` mgmt command not mentioned in the PR description. (a)+(b) is the actual fix from the title; (c)+(d) is operator tooling for retroactive cleanup. The mgmt command alone is bigger than the algorithm change and has no test coverage. Splitting would let the algorithm change land + bake in production while the ops tooling iterates separately.
2. **`use_existing=True` is a load-bearing invariant that isn't documented in the function.** Per Mohamed's reply on 2025-08-29: when `use_existing=True`, existing sessions can only merge, never split. That's the right behavior for production safety on every sync, but it's not stated in the function docstring or enforced anywhere. Worth either adding to the docstring or surfacing as a named flag, since a future reader would have a hard time inferring this from code alone.
3. **Migration rollback works, but is lossy.** `populate_group_by_on_rollback` synthesizes `2024-01-01-2`-style suffixed strings to satisfy the old unique constraint. The rollback succeeds structurally but the original `group_by` values are gone — meaning rollback isn't really a "go back and investigate" path. Worth mentioning in the migration docstring so anyone considering rollback knows what they're getting.

## Smaller cleanups (low priority)

- `create_local_admin_user` fixture in `ami/tests/fixtures/main.py:467` has duplicate `password = os.environ.get(...)` and `logger.info(...)` lines — looks accidental.
- The two "another approach to keep occurrences up-to-date" comments in `SourceImage.update_calculated_fields` (~line 1813) — main solved this via the Subquery realignment, so those comments can be deleted.
- The `# noqa` after `remaining_events = events[current_index + 1 :]` — unclear what it's suppressing. Either drop the comment or explain it.
- `recalculate_events` mgmt command (710 lines) duplicates stats logic that probably belongs alongside `group_images_into_events` or in `audit_event_lengths`. A shared helper would be cleaner.
- `models.py` is +250 lines on top of an already 4000+ line module. The `EventQuerySet`/`merge_with_following_events`/`group_images_into_events` cluster is a natural extract candidate per the existing CLAUDE.md tech-debt note — could happen as a follow-up.

---

Tests after merge: 14/14 grouping, 182/182 main, 112/112 jobs+tasks. Branch is in good shape — just wanted to surface the discussion items before they got buried.

Issue #260, opened 2023-09-29
Split ami.main into multiple modules


mihow:
- Split up ami/main/models.py into ami/main/models/*.py
- Same for serializers
- Same for views
- Same for tests (ami/main/tests.py)
- Move ami/main/api/* to ami/main/

Consider multiple apps vs. single app with multiple model files. Moving model classes to new apps requires making data migrations because you are moving tables. I suggest first splitting into separate model files within the main/ app as suggested above, then moving into new apps. If moving into new apps, the table names can be manually so no data migrations are needed during development. 

Step 1: make a list of models in main that _should_ be split into their own files or apps and share that here.

See notes and the beginning of this in #918

kra:
This will be a long-running task and people who start working on it will probably be diverted to other things for tit to be picked up again over it's lifetime. And the new layout will probably change over time as code or opinions change. So we should expect to see a hybrid of the old and new layout for a while - in fact, we already have it now.

The current loose strategy probably works:
- create models_future that follows the new layout, including a docstring in init
- put new main.models code in models_future when feasable
- move main.models code to models_future on updates when feasable
- when we've gotten to some subjective far enough point, rename models_future to models and move models.main to models.legacy or models.misc or something

PR #909, opened 2025-07-29
Auto-process manually uploaded images (if enabled)

mihow:
This pull request introduces a workflow for automatically processing manually uploaded images, if enabled (via feature flag). This was implemented as a prototype to allow users to test data more quickly (https://github.com/RolnickLab/antenna/milestone/29). It will be disabled by default however the prototype introduced several features and fixes that are helpful in general (and will be available by default).

- Add support for reading image timestamps from EXIF data for manual uploads
- Created new directory structure for ML modules under `orchestration/` where methods like `process_single_image and `get_default_pipeline` can live. Many methods from `ml/models.py` and `main/models.py` should be moved here, as well as some of the upcoming post_processing functions such as `tracking.py` See #260, #863, #915, #857
- Fix the "Source images" header when adding images to Stations (it was sharing a translation string with Collections)

### DIFF
--- a/ami/ml/orchestration/pipelines.py
+++ b/ami/ml/orchestration/pipelines.py
@@ -13,7 +13,7 @@ def get_default_pipeline(project: Project) -> Pipeline | None:
 
     @TODO use project settings to determine the default pipeline
     """
-    return (
+    default_pipeline = project.default_processing_pipeline or (
         Pipeline.objects.all()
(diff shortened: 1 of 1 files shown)


mihow:
@mihow put this behind a feature flag in the project settings and then I think we can merge it

The EXIF reader can always happen by default - consider falling back to todays date

PR #914, opened 2025-08-06
Make project description optional

annavik:
## Summary

We currently have some projects with a dot as the description. Even though we think it's nice when projects have descriptions, I think the dots are a sign we should make it optional. This tweak will also simplify the upcoming quick start feature.

### List of Changes

* Make project description optional FE side
* Make project description optional BE side
* Tweak project gallery cards to handle missing description

### Screenshots
Some projects that use dot as description:

Screenshot 2025-08-06 at 10 16 02


mihow (approved):
Thanks for this! It's been bugging me too :) And thanks for making sure the project cards still look nice without a description. 

All that was missing was a database migration since you modified a Model field, so I added that.

PR #929, opened 2025-08-28
Setup default filters form

annavik:
### Summary

In this PR we setup a new view "Default filters" where users can control:
- Default score threshold for project
- Taxa to include for project
- Taxa to exclude for project

This will help us present more relevant data for users and simplify view settings. The form is put behind a feature flag "default_filters" since values are not yet applied backend side.

### How To Test
- Make sure the default filters form is **not visible** by default
- Set feature flag "default_filters" to true from admin for the current project
- Make sure the default filters form is **visible**
- Change some values and save
- Check that values are updated in admin for the current project

### Screenshots
Screenshot 2025-08-28 at 11 12 05

### Future plans
For future settings, one idea could be we setup a dedicated settings view and reuse the sidebar design. In this way, we could allow settings to expand with time without bloating the main view too much.

Screenshot 2025-08-28 at 11 23 11


mihow (approved):
Tested and all works! I'm excited about this multiple entity select component too!

image

Issue #927, opened 2025-08-26
Implement the Taxa filters in the backend 


annavik:
In https://github.com/RolnickLab/antenna/pull/929 we setup a project settings form, where default taxa filters that can now be controlled on a project level. This task is about applying these filters backend side when data is returned.

Views to consider:
- Stats bars and charts
- List views
- Capture view
- Numbers in columns
- Taxa search

Issue #926, opened 2025-08-26
Implement the Score threshold filter in the backend


annavik:
In https://github.com/RolnickLab/antenna/pull/929 we setup a project settings form, where score threshold is one of filters that can now be controlled on a project level. This task is about applying this filter backend side when data is returned.

Views to consider:
- Stats bars and charts
- List views
- Capture view
- Numbers in columns
- Taxa search

PR #930, opened 2025-08-28
[Draft] Introduce default filters feature

annavik:
### Summary

In #929, we setup a view to make it possible to adjust default filters for a project. In this PR, we apply the filters. The changes includes both frontend and backend updates.

### List of Changes

* [FE] Update reject options to use default filters ✅
* [FE] Exclude score threshold control from occurrence view ✅
* [FE] Exclude score threshold control from species view ✅
* [FE] Exclude score threshold control from playback view ✅
* [FE] Inform and link to default filters from list views ✅
* [BE] Apply score threshold filters backend side ⏳
* [BE] Apply score taxa filters backend side ⏳
* [FE and BE] Make it possible to ignore default filters on demand ⏳
* [BE] Consider setting default exclude & include taxa for existing projects

### What has been tested

- [x] Reject options work when identifying, use the new "exclude Taxa" choices
- [ ] Threshold filter applies to all list views
- [ ] Threshold filter applies to all detail views
- [ ] Threshold filter applies to nested stats & lists (the detections in the session detail view)

### Related Issues
- #926
- #927 

### Screenshots

Informing about default filters and linking to settings:
Screenshot 2025-08-28 at 11 13 32

To make it possible to exclude filters on demand, one idea is to include a toggle:
Screenshot 2025-08-28 at 11 13 58

## Notes for development

This requires an update to the Nova package (our UI kit)

```
yarn install  # if running the ui on host machine
docker compose build ui  # if running the UI using docker stack
```

### DIFF
--- a/ami/main/api/views.py
+++ b/ami/main/api/views.py
@@ -1109,7 +1109,6 @@ class OccurrenceViewSet(DefaultViewSet, ProjectMixin):
         qs = qs.with_detections_count().with_timestamps()  # type: ignore
         qs = qs.with_identifications()  # type: ignore
         qs = qs.filter_by_score_threshold(project, self.request)  # type: ignore
-
         if self.action != "list":
             qs = qs.prefetch_related(
                 Prefetch(
--- a/ami/main/models.py
+++ b/ami/main/models.py
@@ -2461,6 +2461,8 @@ class OccurrenceQuerySet(models.QuerySet["Occurrence"]):
         return qs
 
     def filter_by_score_threshold(self, project: Project | None = None, request: Request | None = None):
+        if project is None:
+            return self
         score_threshold = get_default_classification_threshold(project, request)
         return self.filter(determination_score__gte=score_threshold)
 


mohamedelabbas1996:
The following API views now apply the default classification threshold filter automatically when returning data. If needed, this behavior can be overridden by including `apply_defaults=false` in the query parameters, along with a custom `classification_threshold`.

GET /api/v2/occurrences/
Returns filtered occurrences based on the classification score threshold.

GET /api/v2/taxa/
Returns taxa with example occurrences that are filtered using the threshold.

GET /api/v2/captures/
Returns source images with annotated counts (occurrences_count, taxa_count) filtered by the threshold.

GET /api/v2/events/
Returns taxa_count filtered by the threshold.

GET /api/v2/status/summary/
Summary view that applies the threshold filter internally when retrieving occurrence-based statistics.

annavik:
> @annavik @mohamedelabbas1996 should this PR be closed now that we have #944 and #945?

Yes, however I think we should clean up #945 a bit. With the separated approach, I think we should rebase 945 on 944.

Closing this though! :)

PR #944, opened 2025-09-12
Apply score threshold filter

annavik:
### Summary
In https://github.com/RolnickLab/antenna/pull/929, we setup a view to make it possible to adjust default filters for a project. In this PR, we apply the filter for score threshold. The changes includes both frontend and backend updates.

### Related Issues
Fixes https://github.com/RolnickLab/antenna/issues/926

### List of Changes
- Added `get_default_classification_threshold(project, request)` to read threshold from project settings (falling back to request params if `apply_defaults=false`).

- Added `filter_by_score_threshold()` filtering queryset method to `OccurrenceQuerySet`.

- Applied default threshold filtering in:

    `OccurrenceViewSet`
    
    `TaxonViewSet`
    
    `SourceImageViewSet`
    
    `SourceImageCollectionViewSet`
    
    `EventViewSet`
    
    `SummaryView`

- Removed `OccurrenceDeterminationScoreFilter`.

## Detailed Description

This PR is the first step toward using default project-level settings filters across the platform.
It introduces support for applying the project’s configured default classification score threshold to multiple API endpoints. When a threshold is defined in the project settings, all relevant endpoints (Occurrences, Taxa, Events, Source Images, Collections, and Summary) automatically filter out records below that threshold.

At the same time, flexibility is preserved: users can bypass the defaults by passing `apply_defaults=false ` in their request, which reverts filtering behavior back to the per-request classification_threshold parameter.

### How to Test the Changes

1. Go to the Project Details admin page and set a default classification score threshold.

2. Check that the Occurrence, Taxon, Event, and Summary endpoints all respect the threshold.

3. Add ?apply_defaults=false to API requests and confirm that results are no longer filtered by the project defaults.

4. Remove the flag (or set apply_defaults=true) and confirm that defaults are applied again.

### Screenshots

TBD

## Deployment Notes
TBD

## Checklist

- [x] Any dependent changes have already been merged to main."

### DIFF
--- a/ami/main/models.py
+++ b/ami/main/models.py
@@ -299,14 +299,10 @@ class Project(ProjectSettingsMixin, BaseModel):
         for deployment in self.deployments.all():
             deployment.update_calculated_fields(save=True)
 
-    def save(self, *args, update_related_calculated_fields: bool = True, **kwargs):
+    def save(self, *args, **kwargs):
         super().save(*args, **kwargs)
         # Add owner to members
         self.ensure_owner_membership()
-        # Update calculated fields including filtered occurrence counts
-        # and taxa counts for related deployments and events
-        if update_related_calculated_fields:
-            self.update_related_calculated_fields()
 
     class Permissions:
         """CRUD Permission names follow the convention: `create_<model>`, `update_<model>`,


annavik:
> @annavik should the bounding box colors change in the session detail view when the default score is changed?
> 
> image

I'm thinking not. But I think detections should be filtered backend side, depending on the default score, similar to the occurrence view.

ui/src/utils/constants.ts
```diff
@@ -107,3 +107,59 @@ export const SCORE_THRESHOLDS = {
...
+ * @param projectScoreThreshold - The project's configured score threshold (e.g., 0.6)
+ * @returns Object with ALERT and WARNING thresholds for use in styling logic
+ * 
+ * @example
+ * // With project threshold of 0.6:
+ * const thresholds = calculateDynamicScoreThresholds(0.6)
+ * // Returns: { ALERT: 0.64, WARNING: 0.84 }
+ * // Red: < 0.64, Orange: 0.64-0.84, Green: >= 0.84
+ */
+export const calculateDynamicScoreThresholds = (
```
annavik, line 128: I gave the dynamic scores some thought. It's a cool experiment, but I think I suggest we keep the thresholds constant, for explainability reasons 🤔 However in future, it would be nice to make them configurable for the project, like the filter threshold.

mihow:
@annavik I reverted the dynamic color coding, and they gray/neutral coloring, and I am now returning occurrence data for all detections that have it. Use the new `occurrence_meets_criteria` property on each detection with a capture response to determine whether to render that detection when the Apply Default Filters toggle is on. Let me know if that works!


http://localhost:4000/api/v2/captures/331210/?project_id=24
image

PR #954, opened 2025-09-18
Introduce generic post-processing framework

### LINKED ISSUE #957: Implement a reusable post-processing framework
## Summary

We need a reusable framework in Antenna to handle post-processing steps on processing service pipelines outputs. The framework should allow us to define post-processing tasks (locally or remotely) that can be triggered in a consistent way, with outputs saved back into the system using the same interfaces as existing pipeline results.

## Detailed Description

TBD
## Implementation Details

TBD
## Related Issues

TBD

## Additional Context

TBD

mohamedelabbas1996:
## Summary

This PR introduces a reusable and unified framework for post-processing in Antenna, providing a consistent pattern to implement and manage post-processing tasks.

### List of Changes

- **Introduced Post-Processing Framework**
  - Added a new base class `BasePostProcessingTask` to define a common structure for all post-processing tasks.

- **Added Two basic Post-Processing Tasks**
  - **Small Size Filter**  
    Marks detections with a small relative bounding-box area (compared to the full image) as *Not Identifiable*,  making it easy for  noisy or low-information detections to be  filtered out.

- **New Job Type: `PostProcessingJob`**
  - Introduced a new `JobType` that executes post-processing tasks .
 
- **Trigger tasks from admin page**
  - Both the **Small Size Filter** task can now be triggered directly from the **SourceImageCollection** admin page.

### Related Issues

#957

## Detailed Description

This PR lays the foundation for Antenna’s post-processing framework.  
It provides a modular, extensible framework for running data cleanup and refinement tasks after the main classification pipeline to improve the processing pipeline results.

The post-processing framework allows to:
- Implement new post-processing logic by simply subclassing `BasePostProcessingTask`.
- Execute them as jobs through the existing job infrastructure.
- Access logging, progress tracking, and error handling.

Initial tasks include:
- **Small Size Filter**: Flags small detections as non-identifiable.

### How to Test the Changes

1. Open the Django admin interface.
2. Go to **Source Image Collections**.
3. Select one or more collections.
4. Trigger:
   - **Run Small Size Filter** admin actiona and  verify detections below the size threshold are relabeled as *Not Identifiable*.
5. Observe logs and job progress under **Jobs** to confirm successful execution and completion.

### Screenshots

image

image

image

## Deployment Notes

Includes several migrations

## Checklist


mihow (commented):
Thanks for thinking about the framework abstractly, @mohamedelabbas1996! This looks like a good start.

Another aspect to consider: How do we want to show the output of the post-processing functions in the UI and track what was applied an occurrence in the DB? 

Right now we show the classification model that was used, the model type, and the date that the prediction was applied. I think we should add a new field for tracking the post-processing step that was applied as well. 

Classification
- model
- date 
- filter name / filter class or list of post_processing steps

**Alternatively**!

We could register each post processing step as an Algorithm. This may fit into our current structure more naturally with less effort (a Pipeline is already a series of algorithms applied).

image

It may work for some filters and not others. But most of them are types of algorithms. (rank rollups, tracking, etc). 

In the AMI Data Companion we consider the tracking stage as the last algorithm applied.

mihow:
This is getting super close! All of the co-pilot comments look valid, but note that they are about the specific filters, rather than the overall framework.

Also I am noticing the determination is not updating automatically. After I run a filter it still looks like this:
image

But should look like this:
image

After the new classifications are created in batch, You have to loop through every occurrence that was modified and run `update_determination()`. There is no batch method for that.

ami/main/models.py
```diff
@@ -2225,7 +2227,17 @@ class Classification(BaseModel):
         related_name="classifications",
     )
     # job = models.CharField(max_length=255, null=True)
-
+    applied_to = models.ForeignKey(
```
mihow: Looks good, thank you

Issue #923, opened 2025-08-25
Implement new UI views for project quick start


annavik:
## Summary

In #902, we designed new UI views for the project quick start. This task is about implementing them.

## Background
It’s currently too complicated for users to test out Antenna. To setup a project, multiple steps are involved and we are also missing documentation about these steps. We want an easy way for users to upload a few trap images and quickly see what Antenna can do with them.

To address this problem, a project quick start was designed in #902. Antenna is a data management platform and we want to highlight also these features, not just the algorithm output. We also think a “project quick start” can benefit both current and new users.

## Implementation Details

See design sketches and more comments here: https://docs.google.com/presentation/d/1U_trejvDxVZVyG9GnTjn3XJvlvqRSlLzLRLvI8py_J4/edit?usp=sharing

For the frontend parts, Anna estimates she needs about 1 week of work to implement the interface, the logic and connect this to the backend. In practice this means 2 weeks since Anna is working 50% on Antenna.

## Related Issues
The implementation of this depends on a few backend tasks:
 - Support for private drafts (#899)
 - Extended support for automation (to make it possible to do a setup in one step as the design suggests)

PR #979, opened 2025-10-06
Pipeline picker updates

annavik:
## Summary

- Show pipeline descriptions in default pipeline picker to provide more guidance
- Limit number of pipeline options available for new projects
- Tweak form field description

### Related Issues

Follow up to #923 

### Screenshots

Before:
Screenshot 2025-10-06 at 10 53 11

After:
Screenshot 2025-10-06 at 10 56 15

### DIFF
--- a/ui/src/pages/project-details/pipelines-select.tsx
+++ b/ui/src/pages/project-details/pipelines-select.tsx
@@ -1,6 +1,15 @@
 import { usePipelines } from 'data-services/hooks/pipelines/usePipelines'
 import { Select } from 'nova-ui-kit'
+import { useMemo } from 'react'
 import { useParams } from 'react-router-dom'
+
+// TODO: Define these options backend side?
+const DEFAULT_PIPELINE_IDS = [
+  '3', // Quebec & Vermont moths
+  '2', // UK & Denmark moths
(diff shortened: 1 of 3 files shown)


ui/src/pages/project-details/pipelines-select.tsx
```diff
@@ -1,7 +1,16 @@
 import { usePipelines } from 'data-services/hooks/pipelines/usePipelines'
 import { Select } from 'nova-ui-kit'
+import { useMemo } from 'react'
 import { useParams } from 'react-router-dom'
 
+// TODO: Define these options backend side?
+const DEFAULT_PIPELINE_IDS = [
```
annavik, line 7: Perhaps not optimal, but when project ID is not specified (which is the case for a new projects), the backend is returning the full list of pipelines.

annavik:
In this case, I think the comments from Copilot would make the code a bit too verbose, I ignored.

I will go ahead and merge this to make the quick start demo more complete, but I created a ticket about defining recommended pipelines backend side: https://github.com/RolnickLab/antenna/issues/982

mihow:
Can you use the pipeline keys/slugs instead of IDs? That should work across
environments

PR #1007, opened 2025-10-16
Update deployment models

annavik:
Fixing a type issue.


mihow (commented):
Thanks! that's great we can reuse site & device in both the list & detail views now

PR #1020, opened 2025-10-29
Add taxa from UI

### LINKED ISSUE #975: Make it possible for users to add missing species
## Summary

When users are labelling crops, if a species is missing they currently have to reach out to the Antenna team and ask us to add it. We then check if we can find the species on GBIF, if there are any synonym species to consider, any alternative spellings, etc. before we add it. This process is time consuming for both the Antenna team and for users. We want to find a way for users to add species without support, but still keep the Antenna taxonomy clean and free from duplicates.

## Detailed Description

Thoughts:
- The backend is already prepared to let users add or modify species. It's the process around it that needs some more thinking, to make sure we will not end up with taxonomy chaos...
- One idea could be we use the GBIF or iNaturalist API and let users search for species? If a species is present there, it should be pretty safe to add.
- A lot of species however, especially in tropical regions, are not present on GBIF. When labeling data, users might still want to refer to such species. Can we find a way to still support this? Maybe make it possible to add project specific species, that are not connected to any external resource.

annavik:
## Summary

A small first step towards more flexible taxa management. Still more work and discussions to have related to this, but at least this will make it possible for users to add missing taxa that is present on GBIF. See notes below for some specific questions to discuss before merge.

### List of Changes

* Setup a new GBIF select component (similar to our taxa select component, except this will call the GBIF API in the background)
* Setup a form where users with taxa create permissions can add global taxa

### Related Issues

Closes #975 

### Notes
- Add globally or to project? I couldn't notice a difference when testing.
- What happens if a user adds the wrong parent? Do we need more checks here?
- How to limit the GBIF search? Currently allowing add all taxa below [Insecta](https://www.gbif.org/species/216). Is this too flexible?

### Screenshots and videos

Screenshot 2025-10-29 at 11 42 46

https://github.com/user-attachments/assets/639a2b0c-28bc-4a26-9ba1-1850c7392a16


ui/src/components/gbif/useGBIFSearch.ts
```diff
@@ -0,0 +1,63 @@
+import axios from 'axios'
+import { useEffect, useState } from 'react'
+import { GBIFTaxon } from './types'
+
+const BASE_URL = 'https://api.gbif.org/v1/species' // See docs at https://techdocs.gbif.org/en/openapi/v1/species
```
annavik: More details about the GBIF API call here!

mihow (commented):
This is a awesome! The search works well.

- I like the rank filter. I'm not sure everyone will, but as long as it defaults to SPECIES I think it's an effective way to limit the search results and find what you are looking for.
- I think the Name field should be presented more as a search field, or at least say "Search" instead of "Select a value".
- I think it's important that we add parents automatically. That seems quite cumbersome for users. The GBIF API should return a parent ID. But we will likely need a backend function to connect the parent to existing parents in the Antenna DB. 
- **OR** you could reverse the form fields a bit.  Ask: "Where do you want to place the new taxa?" Then you search our existing taxa for the proper parent. Then we will know the rank of the parent, and we could even show existing children. Then when you search for a taxa, we already know the rank to filter on. 

I'm guessing the global vs. project taxa filter was added to the `ood` branch but not in main. 

I think we can allow users to add global taxa if they are from GBIF. But I know Yves and Chris will want to add custom taxa not in GBIF, which should be just for their project.

annavik:
Regarding this PR, it has been on pause because I got a bit hesitant to the whole thing. I wonder if this could be a step in the wrong direction? All entomologists I have talked to seem skeptical to the GBIF taxonomy in general and I worry they will not like this approach to extending the Antenna taxonomy.

In recent discussions, it sounds like we might move away from global taxonomy. In future, projects on Antenna might manage their own taxonomy in more isolation, but more freely. If this is the new approach, I think we should rethink both the backend and UI and perhaps close this PR.

Let me know what you think! @mihow @f-PLT

annavik:
Closing this! Feel free to revisit the code if this becomes relevant in future :)

PR #1024, opened 2025-10-30
Docker healthcheck and auto-restart for Celery workers

mihow:
## Summary

Adds Docker healthchecks and automatic restart for the Celery worker and beat services. Addresses the failure mode observed on 2026-04-16, where the celerybeat container showed `Up 10 hours` / `RestartCount=0` in `docker ps`, but its scheduler thread was hung — the `jobs_health_check` beat task hadn't fired in 12+ hours, so stuck job 2421 was never reaped. Docker's default `restart: unless-stopped` only catches process death, not scheduler/worker hangs, so we need an explicit liveness signal plus a mechanism to act on it.

## What this changes

**Worker healthcheck** — `celery inspect ping` over the broker, via `compose/{local,production}/django/celery/healthcheck.sh`. Catches stuck, deadlocked, and disconnected workers that a PID check would miss.

**Beat healthcheck** — file-mtime heartbeat. Beat doesn't respond to `inspect ping` (it's a worker control message), and `DatabaseScheduler` has no on-disk schedule file. So `ami/celery_schedulers.py` adds `HeartbeatDatabaseScheduler`, a subclass of `django_celery_beat`'s `DatabaseScheduler` that touches `/tmp/beat-heartbeat` at the top of every `tick()`. The touch runs inside the beat process, so the file lives in the beat container's filesystem. The healthcheck script (`healthcheck-beat.sh`) fails if the file is older than 2 minutes. If the scheduler hangs anywhere (Redis pool lock, DB call, sync deadlock), the file goes stale and Docker flips the container to `unhealthy` — which would have caught the 2026-04-16 incident.

**Autoheal sidecar** — `willfarrell/autoheal:1.2.0` added to production, staging, and worker compose files. Docker Compose still has no native restart-on-unhealthy (that's swarm/k8s territory — [compose issue #4826](https://github.com/docker/compose/issues/4826) has been open since 2017). Autoheal watches the docker socket and restarts any container labeled `autoheal=true` that Docker marks unhealthy.

**Staging compose** (`docker-compose.staging.yml`) now has the same healthchecks + autoheal. Previously it had zero celery healthchecks. This is what the demo host runs.

**Worker start scripts**
- `compose/production/django/celery/worker/start`: multi-line format; comment explains how healthcheck, `max-tasks-per-child`, autoheal, and `restart: always` layer together
- `compose/local/django/celery/worker/start`: renames `DEBUGGER=1` → `CELERY_DEBUG=1`, adds `CELERY_NO_RELOAD=1` to disable watchfiles
- Worker protections (`--max-tasks-per-child=100`, `--max-memory-per-child`) kept from earlier draft

**Production compose** — `celeryworker` removed; workers run on dedicated hosts via `docker-compose.worker.yml`. This was already the direction.

## Files

New:
- `ami/celery_schedulers.py` (scheduler subclass with docstring explaining the 2026-04-16 context)
- `compose/{local,production}/django/celery/healthcheck.sh` (worker)
- `compose/{local,production}/django/celery/healthcheck-beat.sh` (beat)

Modified:
- `compose/{local,production}/django/Dockerfile` — add `procps`, COPY healthcheck scripts
- `compose/{local,production}/django/celery/worker/start`
- `config/settings/base.py` — switch `CELERY_BEAT_SCHEDULER` to `HeartbeatDatabaseScheduler`
- `docker-compose.yml`, `docker-compose.production.yml`, `docker-compose.worker.yml`, `docker-compose.staging.yml`, `docker-compose.override-example.yml`

## Verification

Healthy path:
```bash
docker compose -f docker-compose.staging.yml build
docker compose -f docker-compose.staging.yml up -d
# After ~90s start_period:
docker ps --filter name=celery --format '{{.Names}} {{.Status}}'
# celeryworker and celerybeat should both report (healthy)
```

Simulated hang (either service):
```bash
docker exec <celerybeat-container> kill -STOP 1
# Within ~2 min, healthcheck flips to unhealthy, autoheal restarts the container
docker inspect <celerybeat-container> --format '{{.RestartCount}}'
```

## Notes

- Supersedes an earlier draft that used `pgrep -f 'celery.*beat'` for the beat healthcheck. That would not have caught the 2026-04-16 incident — the process was alive; the scheduler thread was not.
- `start_period: 90s` on both healthchecks covers the slow production boot (postgres wait + New Relic agent init + DatabaseScheduler's initial sync).
- Autoheal container names are unique per environment (`ami_production_autoheal`, `ami_worker_autoheal`); the staging compose omits `container_name` so multiple staging instances on the same host don't collide.

### DIFF
--- a/compose/local/django/celery/healthcheck.sh
+++ b/compose/local/django/celery/healthcheck.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Celery worker healthcheck: verify the worker is responsive via the broker.
+# Catches stuck, deadlocked, and crashed workers — not just process existence.
+set -e
+exec celery -A config.celery_app inspect ping \
+    --destination "celery@$(hostname)" \
+    --timeout 10 > /dev/null 2>&1
--- a/compose/production/django/celery/healthcheck.sh
+++ b/compose/production/django/celery/healthcheck.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+# Celery worker healthcheck: verify the worker is responsive via the broker.
+# Catches stuck, deadlocked, and crashed workers — not just process existence.
+set -e
+exec celery -A config.celery_app inspect ping \
+    --destination "celery@$(hostname)" \
+    --timeout 10 > /dev/null 2>&1
--- a/compose/production/django/celery/worker/start
+++ b/compose/production/django/celery/worker/start
@@ -12,8 +12,18 @@ set -o nounset
 #
 # With prefork pool (default), each CPU gets a worker process.
 # Example: 8 CPUs × 2 GiB = 16 GiB max total worker memory
+#
+# These options work in conjunction with the Docker healthcheck:
+# - Healthcheck (celery inspect ping) detects STUCK workers (not responding)
+# - These options prevent RESOURCE LEAKS (memory/task buildup over time)
+# - Autoheal restarts UNHEALTHY containers
+# - restart:always brings containers back after any exit
 
 MAX_TASKS_PER_CHILD=100
 MAX_MEMORY_PER_CHILD=2097152  # 2 GiB in KB
 
-exec newrelic-admin run-program celery -A config.celery_app worker --queues=antenna -l INFO --max-tasks-per-child=$MAX_TASKS_PER_CHILD --max-memory-per-child=$MAX_MEMORY_PER_CHILD
+exec newrelic-admin run-program celery -A config.celery_app worker \
+    --queues=antenna \
+    -l INFO \
+    --max-tasks-per-child=$MAX_TASKS_PER_CHILD \
+    --max-memory-per-child=$MAX_MEMORY_PER_CHILD
--- a/ami/celery_schedulers.py
+++ b/ami/celery_schedulers.py
@@ -1,0 +1,74 @@
+"""Celery beat scheduler with a Docker-healthcheck-friendly heartbeat file.
+
+Why this module exists
+----------------------
+Docker's default ``restart: unless-stopped`` only catches process death, not a
+frozen scheduler thread. On 2026-04-16 the celerybeat container on ami-live
+showed "Up 10 hours" in ``docker ps`` with four live PIDs and
+``RestartCount=0``, yet its last log line was "Sending due task
+celery.check_processing_services_online" twelve hours earlier — a Redis
+connection blip had deadlocked the connection-pool lock and the scheduler
+thread never recovered. The 15-minute ``jobs_health_check`` beat task stopped
+firing, and stuck job 2421 was never reaped.
+
+To let Docker flip the container to ``unhealthy`` on that failure mode, we
+need a heartbeat signal that proves the scheduler's main loop is progressing.
+Constraints:
+
+- Beat does not answer ``celery inspect ping`` (that's a worker control
+  message), so we can't reuse the worker healthcheck.
+- We use ``DatabaseScheduler`` from ``django_celery_beat``, which keeps the
+  schedule in Postgres, so there is no on-disk schedule file whose mtime
+  would update naturally.
+- A plain Celery task written from a worker would touch a file in the
+  **worker's** filesystem, not beat's — Docker healthchecks read files
+  inside the checked container.
+
+So: override ``DatabaseScheduler.tick()`` to touch ``/tmp/beat-heartbeat``
+on every iteration. ``tick()`` runs inside the beat process itself, so the
+file lives in the beat container. If the scheduler loop hangs anywhere
+(Redis pool lock, DB query, sync deadlock), ``tick()`` stops returning and
+the file goes stale within ~60 s. The healthcheck
+(``compose/*/django/celery/healthcheck-beat.sh``) fails, Docker marks the
+container ``unhealthy``, and autoheal restarts it.
+
+Activation
+----------
+Wired in via ``CELERY_BEAT_SCHEDULER`` in ``config/settings/base.py``.
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+
+from django_celery_beat.schedulers import DatabaseScheduler
+
+logger = logging.getLogger(__name__)
+
+HEARTBEAT_PATH = Path("/tmp/beat-heartbeat")
+
+
+class HeartbeatDatabaseScheduler(DatabaseScheduler):
+    """DatabaseScheduler that touches a heartbeat file on every tick.
+
+    Each call to ``tick()`` represents one cycle of the scheduler's main loop:
+    evaluate due tasks, enqueue them, return the seconds until the next tick.
+    If any step in that cycle hangs (e.g. a Redis or DB call blocks forever),
+    ``tick()`` stops returning, the file mtime stops advancing, and the Docker
+    healthcheck flips the container to ``unhealthy`` within ~2 minutes.
+
+    We touch the file *before* delegating to ``super().tick()`` so a successful
+    iteration of the loop itself is what proves liveness; if the heartbeat
+    write ever fails (disk full, permission error), we log at warning level
+    but don't re-raise — an I/O problem writing ``/tmp`` shouldn't take down
+    the scheduler. Docker will eventually mark the container unhealthy on the
+    stale file, which is the right outcome.
+    """
+
+    def tick(self, *args, **kwargs):
+        try:
+            HEARTBEAT_PATH.touch()
+        except OSError as exc:
+            logger.warning("beat heartbeat touch failed: %s", exc)
+        return super().tick(*args, **kwargs)
--- a/compose/local/django/Dockerfile
+++ b/compose/local/django/Dockerfile
@@ -41,6 +41,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
   libpq-dev \
   # Translations dependencies
   gettext \
+  # healthcheck dependencies (pgrep, findmnt, etc.)
+  procps \
   # cleaning up unused files
   && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false \
   && rm -rf /var/lib/apt/lists/*
@@ -73,6 +75,11 @@ RUN chmod +x /start-celerybeat
 COPY ./compose/local/django/celery/flower/start /start-flower
 RUN sed -i 's/\r$//g' /start-flower
 RUN chmod +x /start-flower
+
+# Healthcheck scripts for celery worker (inspect ping) and beat (heartbeat mtime)
+COPY ./compose/local/django/celery/healthcheck.sh /celery/healthcheck.sh
+COPY ./compose/local/django/celery/healthcheck-beat.sh /celery/healthcheck-beat.sh
+RUN chmod +x /celery/healthcheck.sh /celery/healthcheck-beat.sh
 
 
 # copy application code to WORKDIR
--- a/compose/local/django/celery/healthcheck-beat.sh
+++ b/compose/local/django/celery/healthcheck-beat.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+# Celerybeat healthcheck: verify the scheduler is alive by checking heartbeat file age.
+#
+# Beat doesn't respond to `celery inspect ping` (that's a worker control message),
+# and with DatabaseScheduler (django_celery_beat) there's no schedule file whose
+# mtime we can watch. So we rely on a dedicated `ami.tasks.beat_heartbeat` task
+# that runs every 60s via CELERY_BEAT_SCHEDULE and touches /tmp/beat-heartbeat.
+#
+# If beat hangs (e.g. scheduler thread deadlocked on a Redis connection blip —
+# the 2026-04-16 incident), the task stops firing and the file goes stale.
+# Docker flips the container to `unhealthy`, autoheal restarts it.
+#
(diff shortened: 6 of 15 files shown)


mihow (commented):
## Review: Direction and Approach

The worker protections (`--max-tasks-per-child`, `--max-memory-per-child`), autoheal pattern, and local dev improvements (`CELERY_DEBUG`, `CELERY_NO_RELOAD`) are solid additions. The layered defense approach is well-thought-out.

### Key gap: healthcheck strength

The healthcheck currently checks process existence (`pgrep`) and stopped state (`ps stat`). This catches crashed workers and `SIGSTOP`'d processes, but not the harder failure mode this PR describes: **stuck workers** (deadlocked, stale broker connection, frozen on I/O).

A worker with a stale Redis/RabbitMQ connection will still have a running process and pass the current healthcheck. From `docker ps`, everything looks healthy while jobs pile up unprocessed.

**Suggested upgrade**: Add `celery inspect ping` as a second check. It sends a control message through the broker and waits for the worker to respond — the standard way to verify a Celery worker is actually responsive, not just alive. See inline comment on `healthcheck.sh` for a concrete implementation. It needs no extra dependencies (Celery and Django settings are already in the container).

### Other items

- **Duplicate scripts**: Local and production healthcheck scripts are identical (except one comment word). Consider sharing a single script.
- **Dockerfile COPY scope**: The entire `celery/` directory is copied into the image, including start scripts already copied individually above. Should copy only `healthcheck.sh`.
- **Comment accuracy**: A few comments reference "ping" behavior that isn't implemented yet. See inline suggestions.

Overall the architecture is right — the fixes are incremental.

compose/local/django/celery/healthcheck.sh
```diff
@@ -0,0 +1,27 @@
...
+
+for pid in $CELERY_PIDS; do
+    state=$(ps -o stat= -p "$pid" 2>/dev/null | awk '{print substr($1,1,1)}')
+    if [ "$state" = "T" ]; then
+        echo "ERROR: Celery worker process $pid is stopped (state: $state)" >&2
+        exit 1
+    fi
+done
+
+# All checks passed
```
mihow: Consider adding a `celery inspect ping` check here. This is the main value-add — catching stuck/deadlocked workers that pass the process check. The PR description identifies this as the core problem ("workers can become stuck or unresponsive... but Docker doesn't detect this"), and `pgrep` alone can't solve it.

```bash
# Check 2: Is the worker actually responsive via the broker?
WORKER_NAME="celery@$(hostname)"
if ! timeout 12 celery -A config.celery_app inspect ping \
    --destination "$WORKER_NAME" \
    --timeout 8 > /dev/null 2>&1; then
    echo "ERROR: Celery worker $WORKER_NAME not responding to ping" >&2
    exit 1
fi
```

This uses the Django/Celery config already in the container — no extra deps needed. The `--timeout 8` (Celery-level) + `timeout 12` (process-level) stays within the 15s healthcheck timeout.

`celery inspect ping` is heavier than `pgrep` (roundtrip through broker), but the 30s interval is more than enough headroom.

compose/production/django/celery/healthcheck.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+#
+# Celery Worker Healthcheck Script (Production)
```
mihow: This file is identical to `compose/local/django/celery/healthcheck.sh` except for `(Production)` on this line. Consider sharing a single script to avoid divergence:

1. Place it in `compose/shared/celery/healthcheck.sh` and COPY from there in both Dockerfiles
2. Or have the production Dockerfile COPY from the local path

The healthcheck logic is environment-agnostic, so a single source of truth makes maintenance simpler.

compose/production/django/celery/worker/start
```diff
@@ -12,8 +12,18 @@ set -o nounset
 #
 # With prefork pool (default), each CPU gets a worker process.
 # Example: 8 CPUs × 2 GiB = 16 GiB max total worker memory
+#
+# These options work in conjunction with the Docker healthcheck:
+# - Healthcheck detects STUCK workers (not responding to ping)
```
mihow: The healthcheck currently uses `pgrep` — it detects dead workers (process not running) and stopped workers, not stuck ones responding to ping. If `celery inspect ping` is added later, this comment becomes accurate. For now:

```suggestion
# - Healthcheck detects DEAD workers (process no longer running)
```

mihow:
**Claude says:** (on behalf of @mihow, context from the celery beat hang debugged today in #1246)

Picking this back up. The shape is good — this should rebase on main and land rather than start over. One load-bearing change needed, plus a few smaller notes.

### 1. `docker-compose.production.yml` — celerybeat healthcheck needs a different signal

The `pgrep -f 'celery.*beat'` check misses exactly the failure mode we debugged today (see #1246): beat deadlocks inside redis-py's `ConnectionPool.release()` but the process, sockets, and file descriptors all stay up. pgrep passes, autoheal doesn't fire, beat stays wedged for hours until manual restart.

Suggest replacing with a schedule-liveness check:

```yaml
healthcheck:
  test:
    - CMD-SHELL
    - |
      python -c "
      import os, django, sys
      os.environ.setdefault('DJANGO_SETTINGS_MODULE','config.settings.production')
      django.setup()
      from django_celery_beat.models import PeriodicTask, IntervalSchedule
      from django.utils import timezone
      from datetime import timedelta
      min_interval = IntervalSchedule.objects.order_by('every').first()
      threshold = timedelta(seconds=min_interval.every * 2.5) if min_interval else timedelta(minutes=15)
      latest = PeriodicTask.objects.filter(enabled=True).exclude(last_run_at=None).order_by('-last_run_at').first()
      if latest is None or (timezone.now() - latest.last_run_at) > threshold: sys.exit(1)
      "
  interval: 60s
  timeout: 15s
  retries: 3
  start_period: 120s
```

This measures the actual "is beat scheduling?" signal. It's the only thing that catches the silent-hang class of bug. Starts firing 2 min after container start to avoid false positives during boot.

### 2. `compose/production/django/celery/healthcheck.sh` — overhead

`celery -A config.celery_app inspect ping` spawns a full Django process each invocation (~1–2s, ~100MB RSS). At `interval: 30s` that's ~3% baseline CPU just for health. Either raise interval to 60s (worker recovery SLA with `retries=3` still comes in under 3min) or drop the `-A` path overhead if there's a lighter option. Not a blocker, just a knob to consider.

### 3. `docker-compose.worker.yml` — autoheal env var duplication

`AUTOHEAL_START_PERIOD=60` on the autoheal sidecar overlaps with the service-level `start_period: 60s` on the worker healthcheck. One of them is redundant. Also worth setting `AUTOHEAL_DEFAULT_STOP_TIMEOUT=45` (default is 10s) so in-flight ML tasks get a real chance to finish before SIGKILL when autoheal restarts.

### 4. `docker-compose.production.yml` — celeryworker removal is a separate behavior change

Removing the `celeryworker` service from `production.yml` (in favor of `worker.yml` only) is a behavior change that deserves its own PR or at minimum a callout in the body. Anyone bringing up `production.yml` alone on an old deploy flow will silently lose worker capacity. Is `worker.yml` guaranteed up on every env that uses production.yml now?

### 5. Flower healthcheck — add explicit urlopen timeout

The `urllib.request.urlopen(...)` call doesn't pass a timeout arg. The 5s `timeout:` field in docker-compose bounds the total command but it's worth passing `timeout=3` to urlopen for clarity and to avoid relying on docker SIGKILL to free a hung connection.

### 6. `compose/local/django/celery/worker/start` — 1 GiB max-memory-per-child is tight for dev

ML tasks routinely exceed 1 GiB in dev when batch sizes aren't throttled. Suggest matching production (2 GiB) or going higher (4 GiB) to avoid surprise restarts mid-debug session.

---

### Merge plan

1. Rebase on main (6-week drift, low risk).
2. Swap the celerybeat healthcheck to the schedule-liveness check (item 1 above).
3. Address items 3, 5, 6 inline. 2 is a knob, 4 is a callout.
4. Land #1246 first (the underlying redis-py bug) so autoheal isn't papering over a known-fixable issue.
5. Ping for a human review once rebased.

If the schedule-liveness check gets pushback and stalls the PR, land the worker autoheal + flower healthcheck in this PR and split celerybeat into a follow-up rather than letting all three sit another 6 weeks.

mihow:
**Claude says:** Recurrence in a production deployment this evening (2026-05-27 ~15:00 UTC). Same fingerprint as the 2026-04-16 incident this PR was written for. Fresh evidence below, plus a correction to the upstream-fix analysis from the earlier incident — both sharpen the case for landing this PR and point at an underlying fix that's now actually available.

## Recurrence summary

A production celerybeat instance silently hung at ~15:00 UTC. `PeriodicTask.jobs.health_check.last_run_at` froze at 14:45 UTC and beat emitted no further log lines. The 15-min `jobs.health_check` task stopped firing for ~9 hours. A long-running async_api job hit a stage the reaper already knows how to finalize (last-batch images stranded with NATS drained but Redis pending sets not yet cleared), but the reaper never ran, so the job sat in `STARTED` overnight.

Container reported `Up` in `docker ps` with `RestartCount=0`. Default `restart: unless-stopped` did not catch the hang — exactly the failure mode this PR addresses.

## Fresh py-spy stack (main thread; all 16 threads on `futex_wait_queue`)

```
release                       redis/connection.py:1529   <- futex_wait_queue
reset                         redis/client.py:791
__del__                       redis/client.py:783        <- PubSub.__del__
_construct_command_packer     redis/connection.py:340    <- GC fired here
__init__                      redis/connection.py:318
make_connection               redis/connection.py:1524
get_connection                redis/connection.py:1478
mget                          redis/commands/core.py:2009
_reconnect_pubsub             celery/backends/redis.py:106
ensure                        celery/backends/redis.py:384
_consume_from                 celery/backends/redis.py:176
on_task_call                  celery/backends/redis.py:373
apply_async                   celery/beat.py:402
tick                          celery/beat.py:353
```

Same race as before: GC fires inside `AbstractConnection.__init__` (`self._command_packer = self._construct_command_packer(...)` at line 318 reaching 340), finalizes a stale `PubSub`, `PubSub.__del__` calls `connection_pool.release(...)`, which tries to re-acquire `ConnectionPool._lock` already held by the outer `get_connection`. Self-deadlock on a non-reentrant `threading.Lock`.

## Important: redis-py 5.3.x is still vulnerable

The 2026-04-16 incident notes referenced redis-py#3677 as the upstream fix and described it as "merged for 5.3.0." `requirements/base.txt` was bumped to `redis==5.3.1` on that basis. Verifying against the live container and upstream tags shows that claim was incorrect:

| redis-py | `ConnectionPool._lock` |
|---|---|
| 5.3.0 | `threading.Lock()` |
| 5.3.1 | `threading.Lock()` |
| 6.0.0 | `threading.Lock()` |
| 6.1.0 | `threading.Lock()` |
| **6.2.0** | **`threading.RLock()`** (when `cache is None`, which is our case) |

The actual fix landed in **redis-py 6.2.0** (released 2025-05-28). The 6.2.0 release notes credit redis-py#3654 ("Fixed potential deadlock from unexpected `__del__` call"). PR #3677 ("Removing the threading.Lock locks and replacing them with RLock objects to avoid deadlocks") was a further sweep and merged 2025-06-16, after 6.2.0 was already out. Neither change is present in 5.3.x.

So the 5.2.1 → 5.3.1 bump did nothing for this bug. The ~6 weeks since the 2026-04-16 incident were luck on GC timing.

## How this PR holds up against today's recurrence

The scheduler-tick heartbeat catches the failure mode cleanly. The deadlock lives inside `super().tick()`, so `/tmp/beat-heartbeat` would have frozen at the start of the first hung tick. With the 2-minute staleness threshold, the healthcheck would have flipped to `unhealthy` and autoheal would have restarted the container in ~3 minutes — instead of the ~9 hours it took for a human to notice. Design is correct.

## Suggested implementation for the underlying fix (separate PR, not this one)

Rather than monkeypatching the lock, the cleaner fix is to upgrade past the version that actually contains the patch. Two coordinated bumps are needed, because celery 5.4 caps redis-py:

| Package | Current | Suggested | Why |
|---|---|---|---|
| `celery` | 5.4.0 | 5.5.3 (or current 5.5.x) | 5.4 pins `redis>=4.5.2,<6.0.0` (`celery/requirements/extras/redis.txt` on v5.4.0); 5.5 drops the upper bound |
| `kombu` | 5.4.2 | 5.5.4 | matches celery 5.5 |
| `django-celery-beat` | 2.5.0 | 2.9.0 | celery 5.5 compatible |
| `redis` | 5.3.1 | 6.2.0 (or newer 6.x) | first version where `_lock` is `RLock` when `cache is None` |

This sequencing was already sketched as a follow-up in the 2026-04-16 incident notes. Today's recurrence moves it from "optional" to "needed."

After this PR lands, I'd still recommend the version bump as defense-in-depth: PR #1024 keeps the MTTR low if future redis-py or celery versions regress this. The bug already regressed at least once between 5.3.x and 6.1.x and was only fixed in 6.2.0.

## Small nit in this PR

`compose/{local,production}/django/celery/healthcheck-beat.sh` has a comment that says "we rely on a dedicated `ami.tasks.beat_heartbeat` task that runs every 60s via CELERY_BEAT_SCHEDULE and touches /tmp/beat-heartbeat." The actual implementation uses `HeartbeatDatabaseScheduler.tick()` — which is the better design (a beat-scheduled task would not catch a beat-scheduler-thread deadlock, since the task would never get sent). Worth updating the comment so it matches the code.

PR #981, opened 2025-10-07
Update status of disconnected and stale jobs

mihow:
## Summary

Introduces a new periodic status check for all incomplete jobs. 

Fixes: #721 

## Checklist

- [ ] I have tested these changes appropriately.
- [x] I have added and/or modified relevant tests.
- [ ] I updated relevant documentation or comments.
- [ ] I have verified that this PR follows the project's coding standards.
- [ ] Any dependent changes have already been merged to main.

## Release Notes

* **New Features**
  * Added debugging utilities for troubleshooting job and task issues.
  * Introduced automatic periodic checks of unfinished jobs every 3 minutes to maintain consistency.

* **Bug Fixes**
  * Improved concurrent job update handling with enhanced safety mechanisms.
  * Better detection and automatic retry of disappeared tasks.

* **Tests**
  * Added comprehensive test coverage for job status checking and concurrent operations.

PR #910, opened 2025-08-04
[integration] Enable async and distributed processing for the ML backend using Celery

vanessavmac:
## Summary

Currently, we run MLJobs as a single async celery task which is not scalable to large batch image processing jobs. For example, when processing a large number of image (i.e. 100+ images), the singular long running task is easily interrupted due to a failed worker and also takes a long time to complete (since images still have to be processed in a sequential manner).

This PR uses Celery as the task queue and RabbitMQ as the message broker to send batches of images as individual `process_pipeline_requests`. Each queue is dedicated to a specific ML pipeline (named after the pipeline slug). A celery worker can pick up tasks based on the pipelines they subscribe to. For example, the antenna celery workers may subscribe to all queues; an external celeryworker would subscribe to select pipelines.  

The `ami.ml.tasks.check_ml_job_status` task listens for completed `process_pipeline_requests` and enqueues `save_results` tasks; the `check_ml_job_status` reschedules itself when there are still in progress tasks.

### Related PR

This PR focuses on the "controller" side of the batch image processing system. This PR should be ready to deploy (pending further testing in a Staging environment) and tasks will be processed by antenna workers.

A follow up PR addresses adding additional consumers via the processing service template: https://github.com/RolnickLab/antenna/pull/1011

### List of Changes

* Add a `process_pipeline_request` task which takes a `PipelineRequest` and returns the model's results. The logic for this task is defined on the antenna side and can be added to external celeryworkers/processing services.
* Introduces a new `MLTaskRecord` model which stores the results and stats of a celery task.
* Update the `Job` model to include `ml_task_records` to track the celery tasks currently in-progress and completed (these can be either `process_pipeline_request` or `save_results` tasks)
* Add  `ami.ml.tasks.check_ml_job_status` which checks the subtasks of an `MLJob`, updates the job status, and schedules `save_results` tasks.
* Add a celery beat task `ami.ml.tasks.check_dangling_ml_jobs` which revokes MLJobs that are `STARTED` but has had not been checked by `check_ml_job_status` recently.

* All tests should now be run with the separate `docker-compose.ci` stack to ensure isolated testing.

### Related Issues

Addresses in part #515 

The planning of this feature was discussed in #515. See the comments beginning at https://github.com/RolnickLab/antenna/issues/515#issuecomment-3046972969

## Detailed Description

Note, this PR doesn't include external celery workers, it only updates the antenna celery worker.

```mermaid
%%{init: {"theme": "base", "themeVariables": { "background": "#ffffff" }}}%%
[...]

PR #1025, opened 2025-10-31
Add periodic status check for incomplete jobs 

mihow:
Simplified version of what's in https://github.com/RolnickLab/antenna/pull/981 & #910 

Primarily addressing jobs where the celery background job is disconnected the job is left running without progress. But also fixes other issues along the way. 

- Check incomplete jobs with a periodic task, mark them as failed when appropriate
- Update the displayed job status more consistently
- Update jobs admin for more visibility and troubleshooting

## Summary by CodeRabbit

* **New Features**
  * Added retry and cancel actions for jobs in the admin interface.
  * Introduced automatic periodic job status monitoring to detect and reconcile stuck, disappeared, or stale jobs.
  * Enhanced admin interface with improved filtering by job type, project, and pipeline; added search for source image collections.

* **Bug Fixes**
  * Jobs are now automatically checked and updated when task state mismatches occur.


carlosgjs (commented):
Couple of comments

mihow:
**Follow-up idea from #1222:** This periodic beat task would be a natural place to emit a mid-flight NATS consumer stats snapshot to the per-job log.

Right now #1222 only logs consumer stats at two points: when a fresh `TaskQueueManager` session reuses an existing consumer (which only happens at publish time, before workers have touched it), and at cleanup (job completion/cancellation/failure). During a long-running job there's no periodic visibility into redelivery pressure — the forensic snapshot only appears after the fact.

If `check_stale_jobs` (or whatever the beat task is called here) already iterates over active async_api jobs, it could fetch `consumer_info` for each and log a snapshot like:

```
NATS consumer job-1272-consumer (delivered=450 ack_floor=169 num_pending=200 num_ack_pending=43 num_redelivered=80)
```

This would make the "workers are pulling but not keeping up with acks" pattern (#1220) visible *during* a job, not just in the post-mortem cleanup line. Low cost — one `consumer_info` call per active job per beat interval.

mihow:
Closing in favor a simple periodic check for now in #1227. However there are some additional checks in this PR that maybe helpful to revisit if we continue to have the same problems.

Issue #721, opened 2025-02-11
Ensure that a processing job's status is always updated correctly


annavik:
## Summary

If something goes wrong with a job, we are seeing an error, but the job status still has the value "STARTED". Sometimes, we have the same problem for successful jobs (job is done, but the status is still "STARTED"). The problem affects the overall status, and also status for individual stages.

## Detailed Description

Not getting the correct status makes it difficult to present correct information to users, for example adjust the status bar color and labels to highlight there is a problem.

### Environment
- OS and/or device: Mac OS
- Browser Chrome
- Version: Production

## Screenshots

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/a1c567d9-8b6f-411e-b91e-77e0de0b8020" />

annavik:
Also a bit related: after doing a retry on the job above (https://antenna.insectai.org/projects/107/jobs/1244), things seem to go better, except the job never changed its status to "SUCCESS" and still looks "STARTED".

annavik:
This job (see https://antenna.insectai.org/projects/107/jobs/1246) also stopped due to problems, but did not had the status updated, so it still looks like it's running:

<img width="1728" alt="Image" src="https://github.com/user-attachments/assets/15f19056-a66b-4edc-bda2-a757f1e1b6c6" />

mihow:
This is related to #719

mihow:
This is addresses & fixed by the changes in #910

annavik:
I'm sorry, but I still have the same problems here. From a user perspective I wouldn't consider this as fixed. For example, this job I started this morning looks like it's running but is actually stuck. I have also recently seen jobs that has errors, but I can't understand from the status if still active or not.

<img width="1728" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c430fe6a-65b8-472f-9fc9-ae259bb25a0d" />

Can we keep this issue open until we have tested things a bit more?

mihow:
Sorry this wasn't suppose to be closed. I must have used a keyword in my PR. It will be fixed in https://github.com/RolnickLab/antenna/pull/910

The changes in #937 only affect the _initial_ status being wrong, when the job is first created.

The job status will never be updated once the background task worker disappears, so I believe that's the cause of this issue.

annavik:
Oh I see, that make sense, thanks for reopening!

mihow:
There are likely two reasons why the job status is not being updated:
[...]

PR #1026, opened 2025-10-31
[Draft] Don't overwrite logs & status in concurrent background tasks

mihow:
Pulled from https://github.com/RolnickLab/antenna/pull/981

More coming soon

### DIFF
--- a/ami/jobs/models.py
+++ b/ami/jobs/models.py
@@ -3,6 +3,7 @@ import logging
 import random
 import time
 import typing
+from contextlib import contextmanager
 from dataclasses import dataclass
 
 import pydantic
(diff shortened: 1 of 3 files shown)


mihow:
**Claude says:** Opened #1337 to capture the root cause this draft targets — concurrent unlocked read-modify-write of `Job.progress`/`status` — with the full lineage (#721 → #981 (closed) → this draft), the appearance-vs-lost-work breakdown, and four candidate fix directions. The key constraint it spells out: any fix has to stay atomic *without* re-adding the `select_for_update` that #1261 removed for contention reasons.

Worth deciding: either flesh this draft out against the #1337 design, or close it in favour of a fresh implementation and keep the spec in #1337. Happy to help compare the #981-derived approach here against the conditional-write / counter-column directions in #1337 before picking one.

mihow:
**Claude says:** Closing as superseded. The status half of this concurrent-write problem is now fixed on `main`:

- **#1338** (merged) makes the result handler's terminal transition a guarded statement-scope `UPDATE ... WHERE status IN finalizable_states()` — no row lock, can't be clobbered by a stale read-modify-write. It ships with a real concurrency regression test that reproduces the lost-update on `main` (red) and passes on the fix (green).
- **#1342** extends the same guard to the other lock-free terminal writers (`cancel()` and the celery task-signal handlers).

The periodic-check approach here (carried over from #981) is no longer the direction. Remaining work is tracked under the **#1337** root-cause hub (counter-accumulation "Layer 2", the `/next` starvation filter, reaper hardening). Thanks for the early groundwork — it framed the problem.

PR #1036, opened 2025-11-07
Speed up default captures list view

mihow:
## Summary

Increase performance of the captures list by removing the occurrence & taxa counts by default. 

### List of Changes

* Updates occurrence count & taxa count for captures, but it's still slow
* Remove the counts per image all together
* Add filesize & image resolution to the capture columns
* Counts can be rendered by passing `?with_counts=true` to the captures endpoint. This could be use to render the columns if the user selects them.
* Add missing config file for flake8 linter to make pre-commit match other uses of flake8

There are two other performance issues which I tested but did not deploy
- The count() query called for the paginator when calculating the number of captures in the response and generating the number pages is very slow as the table grows. I tested making this a "fuzzy" query but it didn't seem to work with filters. See this branch [`feat/countless-paginator`](https://github.com/RolnickLab/antenna/compare/main...feat/countless-paginator)
- The Guardian permissions are checked per object, which causes a lot of n+1 queries. This should be addressed, but cachalot automatically reduced these to one query within each request.  See this [branch fix/permission-n1-queries](https://github.com/RolnickLab/antenna/compare/main...fix/permission-n1-queries)

### How to test

- Comment out `cachalot` in the `config/settings/base.py:93`
- Ensure the capture list view & session detail pages render correctly.
- Test on staging, live or an environment with a large dataset.

I have tested locally on my machine with 15M captures

Test with and without counts
http://localhost:8000/api/v2/captures/?project_id=18&with_counts=true
http://localhost:8000/api/v2/captures/?project_id=18

### Screenshots

image

With counts (5 seconds for 20 images)
image

Without counts (<1 second for 20 images)
image

## Checklist

## Summary by CodeRabbit

* **New Features**
  * File size and resolution columns added to captures table; new actions column enables deletion dialogs.
  * Column-visibility settings let you toggle which capture columns are shown.
  * API can include per-capture occurrences_count and taxa_count when requested.

* **Refactor**
  * Count fetching and query behavior improved for more reliable pagination and large-result performance.

### DIFF
--- a/ami/main/models.py
+++ b/ami/main/models.py
@@ -20,6 +20,7 @@ from django.core.files.storage import default_storage
 from django.db import IntegrityError, models, transaction
 from django.db.models import Exists, OuterRef, Q
 from django.db.models.fields.files import ImageFieldFile
+from django.db.models.functions import Coalesce
 from django.db.models.signals import pre_delete
 from django.dispatch import receiver
 from django.template.defaultfilters import filesizeformat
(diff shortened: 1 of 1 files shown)


annavik (approved):
Thanks for adding those new columns, looking good! Also for exploring ways to optimize the query. To fetch counts on demand sounds like a good workaround for now.

To still make it possible so inspect and sort captures by counts (I do think it's a nice feature and I often use it!) I went on and included the column settings component and brought back the removed count columns (but made them hidden by default). This made it possible to include the "with_counts" query string as part of the fetch URL and automatically refetch data when columns are made visible or if sort settings considers.

As a small bonus I also made it possible to delete a capture from UI.

PR #1044, opened 2025-11-10
Apply Plotly code splitting

annavik:
The Plotly lib is very large, which means if we include it to our main JS bundle, it will affect our FCP metric (First Contentful Paint). By lazy loading this lib, we can present content to users sooner. Compared to other assets we load, a few MB might not seem like a big deal, but it's important to keep the main bundle as small as possible.

I had already prepared code for this, but forgot to apply it to the new species charts.

Before (Plotly part of main bundle):
Screenshot 2025-11-10 at 12 16 23

After (splitting Plotly code into a separate chunk):
Screenshot 2025-11-10 at 12 13 35

You can see here the library itself actually way larger than the rest of the frontend app.

## Summary by CodeRabbit

* **Refactor**
  * Updated internal component imports to utilize an optimized loading pattern for better efficiency.

### DIFF
--- a/ui/src/pages/species-details/species-details.tsx
+++ b/ui/src/pages/species-details/species-details.tsx
@@ -9,7 +9,7 @@ import {
   InfoBlockField,
   InfoBlockFieldValue,
 } from 'design-system/components/info-block/info-block'
-import Plot from 'design-system/components/plot/plot'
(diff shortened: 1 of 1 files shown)


mihow (approved):
That was fewer changes that I would have expected. Nice optimization!

PR #1046, opened 2025-11-11
New jobs API for accessing the ML task queue and posting back results

### LINKED ISSUE #970: Implement PSv2 pull API

carlosgjs:
## Summary

Defines an initial set of API endpoints for an external processing service to read & pull ML tasks from the job queue and post back results (detections & predictions).

- Add tasks() endpoint (stubbed) for NATS JetStream integration
- Add result() endpoint (stubbed) for pipeline result processing
- Add ids_only and incomplete_only query parameters to job list endpoint
- Add JobFilterSet with pipeline__slug filtering and search fields
- Add utility OpenApiParameter definitions for new query params

Important: The tasks() and result() endpoints are documented and stubbed out, awaiting NATS and Celery task dependencies from PR #987.

Implementation follows existing project patterns with `IncompleteJobFilter` backend, `MinimalJobSerializer`, and pipeline schemas (`PipelineProcessingTask`, `PipelineTaskResult`) in `ami/ml/schemas.py`

Closes #970 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Filter jobs by pipeline slug, search by job or pipeline name, and an "incomplete-only" filter
  * "IDs only" list mode returning job IDs and count via a minimal job representation
  * New endpoints to fetch queued tasks (with batching) and to submit pipeline results asynchronously
  * New public task/result data models and API params: ids_only, incomplete_only, batch

* **Tests**
  * Expanded test coverage and helpers for listing, filtering, searching, batching, results, and pipeline/ML job workflows


ami/jobs/views.py
```diff
@@ -153,6 +169,191 @@ def get_queryset(self) -> QuerySet:
...
+        3. POST to /jobs/{id}/result/ with the reply_subject to acknowledge
+
+        This stateless approach allows workers to communicate over HTTP without
+        maintaining persistent connections to the queue system.
+
+        NOTE: This endpoint requires NATS JetStream integration which is not yet
+        available in main. Implementation coming from PR #987.
+        """
+        job: Job = self.get_object()
+        batch = IntegerField(required=False, min_value=1).clean(request.query_params.get("batch", 1))
```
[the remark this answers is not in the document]
carlosgjs: query_params.get("batch", 1) will not return None since a default is provided

mihow (commented):
Comments from our call: Can we keep the API views lighter & abstracted from anything NATS specific? The endpoints & flow should be clear (get next 10 images, post back results) with the focus on view-layer things like validation and checking permissions.

ami/jobs/views.py
```diff
@@ -153,6 +168,137 @@ def get_queryset(self) -> QuerySet:
...
+
+        # Validate that the job has a pipeline
+        if not job.pipeline:
+            raise ValidationError("This job does not have a pipeline configured")
+
+        # TODO: Implement task queue integration
+        logger.warning(f"Task queue endpoint called for job {job.pk} but the implementation is not yet available.")
+
+        return Response(
```
mihow: Ideally we should use a Django REST Framework serializer or Pydantic dataclass for requests & responses. That will automatically give you validation and will allow you to use type checking in your IDE. Also it's helpful when writing tests for mocking requests & responses. See what we do for `PipelineResultsResponse` and `PipelineRequest`. These can go in `ami/jobs/schemas.py` or `ami/jobs/serializers.py`.  Then you can skip all of this logic:

```
            reply_subject = item.get("reply_subject")
            result_data = item.get("result")

            if not reply_subject:
                raise ValidationError(f"Item {idx}: reply_subject is required")

            if not result_data:
                raise ValidationError(f"Item {idx}: result is required")
```

ami/jobs/views.py
```diff
@@ -153,6 +169,132 @@ def get_queryset(self) -> QuerySet:
...
+            incomplete_only_param,
+        ]
+    )
     def list(self, request, *args, **kwargs):
+        # Get the base queryset
+        # Check if ids_only parameter is set
+        ids_only = url_boolean_param(request, "ids_only", default=False)
+
+        if ids_only:
+            queryset = self.filter_queryset(self.get_queryset())
```
mihow: I think this is it the most "native" way to do this in the Django Rest Framework. Confirm that the magic works and it only queries the DB for the IDs, not the whole object and filters out other fields in Python. 

```
def get_queryset():
   qs = super().get_queryset()
   if ids_only:
       qs.only("pk") # or values_list("pk")

def get_serializer():
    if ids_only:
        return JobsIDsOnlySerializer
    else:
        return super().get_serializer()
```

mihow (changes requested):
My preference is that the new view methods should fit into our existing patterns a bit more. The simplest change could just move logic in `filter_queryset` to `get_queryset` since that's the method we override everywhere else. And then use the `qs = qs.filter() ... return qs` pattern.

```python
   def get_queryset(self) -> QuerySet:
        """
        If a project is passed, only return taxa that have been observed
        and add extra data about the occurrences.
        Otherwise return all taxa that are active.
        """
        qs = super().get_queryset()
        if project:
            qs = self.attach_tags_by_project(qs, project)
        return qs
```

Also my guess is that "`ids_only`" might grow into "ids_only plus a few more things" (like `created_at`). I often create a "minimal" serializer like `MinimalProjectSerializer` with just id & name & created at. or a serializer for a particular purpose like `DeploymentNestedSerializerWithLocationAndCounts`.

Also it's cool that you already move out the `JobFilterSet`! You could put all of your conditional filters in that class. Or make individual filter classes. There are a bunch of examples of that for the TaxonViewSet

```python
class TaxonTagFilter(filters.BaseFilterBackend):
    """FilterBackend that allows OR-based filtering of taxa by tag ID."""

    def filter_queryset(self, request, queryset, view):
        tag_ids = request.query_params.getlist("tag_id")
        if tag_ids:
            queryset = queryset.filter(tags__id__in=tag_ids).distinct()
        return queryset

class TaxonViewSet(DefaultViewSet, ProjectMixin):
    # ...
    filter_backends = DefaultViewSetMixin.filter_backends + [
        CustomTaxonFilter,
        TaxonCollectionFilter,
        TaxonTaxaListFilter,
        TaxonBestScoreFilter,
        TaxonTagFilter,
        TagInverseFilter,
    ]
```

Also! It's really great to move reusable filters to the QuerySet class. Your incomplete_only filter is a good candidate for this since it has custom logic that we want to be exactly the same anywhere that we list or count "incomplete" jobs. 

```python
# ami/main/views.py
class OccurrenceViews:
    # ...
    def get_queryset():
        qs: QuerySet = super().get_queryset()
        qs = qs.with_detections_count().with_timestamps()  # type: ignore
        qs = qs.with_identifications()  # type: ignore
        qs = qs.apply_default_filters(project, self.request)  # type: ignore
        return qs

# ami/main/models.py
class OccurrenceQuerySet(BaseQuerySet):

    def filter_by_score_threshold(self, project: Project | None = None, request: Request | None = None):
        score_threshold = get_default_classification_threshold(project, request)
        logger.debug(f"Filtering occurrences by determination score threshold of {score_threshold}")
        filter_q = build_occurrence_score_threshold_q(score_threshold, occurrence_accessor="")
        return self.filter(filter_q)

    def valid(self):
        return self.exclude(detections__isnull=True)

    def with_timestamps(self):
        """
        These are timestamps used for filtering and ordering in the UI.
        """
        return self.annotate(
            first_appearance_timestamp=models.Min("detections__timestamp"),
            last_appearance_timestamp=models.Max("detections__timestamp"),
            first_appearance_time=models.Min("detections__timestamp__time"),
            duration=models.ExpressionWrapper(
                models.F("last_appearance_timestamp") - models.F("first_appearance_timestamp"),
                output_field=models.DurationField(),
            ),
        )
```

All those examples are from these files:
https://github.com/RolnickLab/antenna/blob/main/ami/main/api/views.py
https://github.com/RolnickLab/antenna/blob/main/ami/main/api/serializers.py
https://github.com/RolnickLab/antenna/blob/main/ami/main/models.py


I'm okay to merge sooner and refactor later if you think the first jobs API could change a lot. But otherwise it's really helpful to establish this structure. I am already fighting a lot of technical debt on this project and this structure makes it easier to track down performance issues and reduce inconsistencies (counts different in different views)

ami/jobs/views.py
```diff
@@ -153,6 +169,129 @@ def get_queryset(self) -> QuerySet:
...
+        # Filter to incomplete jobs if requested (checks "results" stage status)
+        if incomplete_only:
+            from django.db.models import Q
+
+            # Create filters for each final state to exclude
+            final_states = JobState.final_states()
+            exclude_conditions = Q()
+
+            # Exclude jobs where the "results" stage has a final state status
+            for state in final_states:
```
mihow: If this is how we will consider "incomplete" jobs in general then I suggest moving it to a QuerySet method

ami/jobs/views.py
```diff
@@ -1,25 +1,71 @@
...
+        # Filter to incomplete jobs if requested (checks "results" stage status)
+        if incomplete_only:
+            # Create filters for each final state to exclude
+            final_states = JobState.final_states()
+            exclude_conditions = Q()
+
+            # Exclude jobs where the "results" stage has a final state status
+            for state in final_states:
+                # JSON path query to check if results stage status is in final states
+                exclude_conditions |= Q(progress__stages__contains=[{"key": "results", "status": state}])
```
mihow: There is also a status field on the top-level job object. I know it's not reliable at the moment, but I don't think the stage statuses are more reliable. Either should work (once the status is reliable!), but the top-level status field will be faster to query.

ami/jobs/views.py
```diff
@@ -153,6 +196,99 @@ def get_queryset(self) -> QuerySet:
...
+        logger.warning(f"Task queue endpoint called for job {job.pk} but the implementation is not yet available.")
+
+        dummy_task = PipelineProcessingTask(
+            id="1",
+            image_id="1",
+            image_url="http://example.com/image1",
+            queue_timestamp=timezone.now().isoformat(),
+        )
+
+        return Response({"tasks": [task.dict() for task in [dummy_task] * batch]})
```
mihow: If this response gets more complex then you can make the Response be it's own schema as well. Then you will get the complete openAPI docs for free.

ami/jobs/views.py
mihow: Oh I see you are just checking the results stage! my bad, that's great.

mihow (approved):

PR #1057, opened 2025-11-19
Fix map zoom

annavik:
Adjusting the initial zoom to make it possible to also see stations that are close. Now the zoom is only based on bounds, not a max value (I'm not sure why we had this restriction before?).

Before (this is 13 stations):
Screenshot 2025-11-19 at 16 18 48

After:
Screenshot 2025-11-19 at 16 23 25

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map zoom behavior when displaying multiple markers by removing zoom constraints.


mihow (approved):
Excellent! Glad to see more ZAC stations being configured as well.

PR #1058, opened 2025-11-19
Handle missing occurrence image

annavik:
Sometimes, occurrences are missing images. In this case, it seems to be because of some issue during processing (see #1059), but it can also happen when processing is in progress (occurrence records are generated before crops are saved). We did not handle this in the new gallery view. In this PR, we fix this!

Before (gallery crashed):
Screenshot 2025-11-19 at 16 25 13

After:
Screenshot 2025-11-19 at 16 28 27

## Summary by CodeRabbit

* **Bug Fixes**
  * Occurrence gallery now displays placeholder icons for missing images instead of broken image references.


mihow (approved):

mihow:
Thanks for catching this and fixing!

PR #1068, opened 2025-11-27
Fix stations sorting

annavik:
## Summary

@mihow noticed that date sorting of stations didn't work as expected. We solve this by moving to server side sorting, which could also be nice for consistency (the stations table was the only one using client side sorting).

### List of Changes

* Add backend support for sorting stations by name
* Sort stations server side instead of client side
* Tweak backend sort settings to avoid errors

### How to Test the Changes

Go to a list of stations and try out different sort settings.

### Screenshots

No visual changes.

## Deployment Notes

Needs backend deploy!

## Summary by CodeRabbit

* **Refactor**
  * Migrated deployment sorting from client-side to server-side processing for improved performance.
  * Aligned deployment listing column and sort field names across frontend and backend systems.
  * Added "name" as a new sortable option for deployment listings alongside capture timestamp fields.

### DIFF
--- a/ami/main/api/views.py
+++ b/ami/main/api/views.py
@@ -236,8 +236,8 @@ class DeploymentViewSet(DefaultViewSet, ProjectMixin):
         "events_count",
         "occurrences_count",
         "taxa_count",
-        "first_date",
-        "last_date",
+        "first_capture_timestamp",
+        "last_capture_timestamp",
         "name",
     ]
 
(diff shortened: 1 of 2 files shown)


annavik:
> Awesome thanks for the quick fix! Will this enable stick sort options too? With the query params like other views have? https://antenna.insectai.org/projects/90/sessions?ordering=-duration

Yes!! :)

mihow (approved):

PR #1074, opened 2025-12-06
Fix the "null" string in the station update form

mihow:
## Summary

The `data_source_regex` field in the Station / Deployment form often gets saved with the string "null". This PR ensures the frontend keeps nulls as `null` rather than converting it to a string. A future enhancement could be to post/patch JSON data instead of form data -- for fields other than the image field. We also need to ensure "clearing" the field works so that PATCH sends `null` instead of omitting the field from the request.

Important: It's unclear when this happens! We are not able to manually reproduce the issue, but it happens regularly. I have updated all existing deployments in the production database that contained "null" in this field (48 in total).

## Screenshots
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented "null" strings from being added to form submissions by serializing nulls as empty values.
  * Stopped null values from being included in URL query parameters, avoiding unwanted "null" entries in searches.
  * Made filtering logic treat null like an absent value to avoid showing or applying null filters.
  * Allowed certain deployment fields to be empty (null) to better represent missing data.

### DIFF
--- a/ui/src/pages/deployment-details/deployment-details-form/section-source-images/section-source-images.tsx
+++ b/ui/src/pages/deployment-details/deployment-details-form/section-source-images/section-source-images.tsx
@@ -91,7 +91,7 @@ export const SectionSourceImages = ({
           <ConnectionStatus
             storageId={deployment.dataSource.id}
             subdir={deployment.dataSourceSubdir ?? ''}
-            regex={deployment.dataSourceRegex}
+            regex={deployment.dataSourceRegex ?? ''}
             showDetails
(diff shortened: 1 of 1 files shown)


annavik (approved):
Thanks Michael! I spent a lot of time today trying to reproduce the issue frontend side (without the fix in place) but I didn't manage. I tried different browsers and even computers haha. I'm still confused why the string shows up in the first place. Your fix is a good patch, but `null` values should not enter that function in the first place.

I fine trying this to see if it will help, but it still bothers me a bit I cannot understand when this happens.

PR #1093, opened 2026-01-20
Implement null detections (calculate what has been processed or not)

### LINKED ISSUE #484: Make it clear what images have not been processed
"Processed" means a source image (aka a "capture") was sent to an ML pipeline and returned a successful result. Right now we are _not_ tracking this. We can only see if a capture has associate detections or not (which do have a created data & algorithm info) but that doesn't say if an image was processed and has no detections or if it failed. So need to add that to the function that handles the pipeline response (`save_results()` in `ami/ml/models/pipeline.py`), which needs some maturation & cleanup overall. 

- When viewing data about multiple images, use a percentage. For example in the Deployment list, show the percent of each deployment's images processed by any pipeline. 
- When looking at a single capture with limited room, show the date that the image was last processed by any ML pipeline. Link to that ML job details. When more room is available, show a list of jobs that have processed this image

Session detail view: 

- [ ] Show when a capture was last processed in the capture details panel
- [ ] Add visualization to horizontal chart showing which captures have and have not been processed, and perhaps which ones failed to process. 

List views - Show percent captures processed

- [ ] Collections table
- [ ] Deployments table
- [ ] Sessions table
- [ ] Project overview?

vanessavmac:
## Summary

If a pipeline doesn't return detections but still successfully processed the image, it will have a "null" detection. This allows distinguishing "not yet processed" from "processed but found nothing."

### Image selection logic for processing

When a pipeline runs, `filter_processed_images()` decides which images to send. Each image is checked against its existing detections from the pipeline's algorithms:

**Yield (needs processing):**
1. No existing detections — image has never been processed by this pipeline
2. Has real detections without classifications — detector ran but classifier didn't
3. Has real detections with classifications, but not from all pipeline classifiers (e.g. a new classifier was added since last run)

**Skip (already processed):**
4. Only null detections exist (bbox=None) — pipeline ran but found nothing
5. Real detections exist and are fully classified by all pipeline classifiers

Null detections are sentinels that mark an image as "processed, nothing found." They are excluded from classification checks so they don't trigger reprocessing.

### List of Changes

**Backend (pipeline processing)**
* Made `bbox` field optional in `DetectionRequest` and `DetectionResponse` schemas
* Added `create_null_detections_for_undetected_images()` to generate null detection records for images where the pipeline found nothing
* Updated `save_results()` to append null detections before saving
* Updated `filter_processed_images()` to skip images that have only null detections (already processed, nothing found) while still yielding images that have real unclassified detections
* Added algorithm-specific lookup for null detections in `get_or_create_detection()` so different pipelines each track their own processed status per-image. Uses `bbox__isnull=True` (not `bbox=None`) to correctly match SQL NULL in Django's JSONField.
* Added `NULL_DETECTIONS_FILTER` constant and `DetectionQuerySet.null_detections()` for consistent null detection filtering
* Excluded null detections from `get_detections_count()`, `update_detection_counts()`, `DetectionViewSet`, and `prefetch_detections()`

**Backend (processed status)**
* Added `SourceImage.get_was_processed()` model method with optional `algorithm_key` parameter
* Added `SourceImageQuerySet.with_was_processed()` annotation for bulk use
* Updated `SourceImageAdmin` to use the annotation in `get_queryset()` to avoid N+1 queries from `get_was_processed` in `list_display`
* Added `was_processed` to timeline API response (`EventTimelineIntervalSerializer`)
* Added `source_images_processed_count` to `SourceImageCollectionSerializer`

**Frontend (capture sets)**
* Added "Processed captures" column to the capture sets table showing count and percentage

**Tests**
* Added `test_image_with_null_detection` for null detection creation and `get_was_processed()` behavior
* Added `test_filter_processed_images_skips_null_only_image` and `test_filter_processed_images_yields_image_with_null_and_real_unclassified` for the skip/yield logic
* Added `test_filter_processed_images_skips_null_and_fully_classified` for the null + real fully-classified scenario
* Added `test_null_detections_are_algorithm_specific` and `test_null_detection_deduplication_same_pipeline` for cross-pipeline isolation
* Updated `test_process_nats_pipeline_result_mixed_results` to include a detection algorithm

### Related Issues

Closes #484

### Follow-up PR

Session detail timeline UI: #1179

### How to Test the Changes

```
docker compose run --rm django python manage.py test ami.ml.tests.TestPipeline --keepdb
```

- View an existing Capture Set that has a % processed (currently this number will never go up and the % of captures with detections will always match % of captures processed)
- Process the Capture Set
- Now the % of captures processed should reach 100% (and be greater or equal to the % captures with detections)
- Process the Capture Set again, no images should be sent to the processing service (unless "always reprocess" has been manually turned on)

## Deployment Notes

No migrations required. Null detections will be created for new pipeline runs going forward. Existing images will not have null detections until they are reprocessed.

## Checklist

### DIFF
--- a/ami/ml/models/pipeline.py
+++ b/ami/ml/models/pipeline.py
@@ -445,12 +445,12 @@ def get_or_create_detection(
         assert detection_resp.algorithm, f"No detection algorithm was specified for detection {detection_repr}"
         try:
             detection_algo = algorithms_known[detection_resp.algorithm.key]
-        except KeyError:
+        except KeyError as err:
             raise PipelineNotConfigured(
                 f"Detection algorithm {detection_resp.algorithm.key} is not a known algorithm. "
                 "The processing service must declare it in the /info endpoint. "
                 f"Known algorithms: {list(algorithms_known.keys())}"
-            )
+            ) from err
         existing_detection = Detection.objects.filter(
             source_image=source_image,
             bbox__isnull=True,
@@ -483,12 +483,12 @@ def get_or_create_detection(
             assert detection_resp.algorithm, f"No detection algorithm was specified for detection {detection_repr}"
             try:
                 detection_algo = algorithms_known[detection_resp.algorithm.key]
-            except KeyError:
+            except KeyError as err:
                 raise PipelineNotConfigured(
                     f"Detection algorithm {detection_resp.algorithm.key} is not a known algorithm. "
                     "The processing service must declare it in the /info endpoint. "
                     f"Known algorithms: {list(algorithms_known.keys())}"
-                )
+                ) from err
 
         new_detection = Detection(
             source_image=source_image,


ami/ml/schemas.py
```diff
@@ -163,14 +163,14 @@ class Config:
 
 class DetectionRequest(pydantic.BaseModel):
     source_image: SourceImageRequest  # the 'original' image
-    bbox: BoundingBox
+    bbox: BoundingBox | None = None
```
carlosgjs: if I understand correctly `DetectionRequest` is used to for reprocessing previously found detections? If so, does it make sense to allow `bbox` to be `None`? Wouldn't we want to create `DetectionRequests` only for real detections?

ami/ml/models/pipeline.py
```diff
@@ -866,6 +871,39 @@ def save_results(
...
+    # Ensure all images have detections
+    # if not, add a NULL detection (empty bbox) to the results
+    source_images_with_detections = [detection.source_image_id for detection in results.detections]
+    source_images_with_detections = set(source_images_with_detections)
+    null_detections_to_add = []
+
+    for source_img in results.source_images:
+        if source_img.id not in source_images_with_detections:
+            detector_algorithm_reference = None
+            for known_algorithm in algorithms_known.values():
```
carlosgjs: this loop over known_algorithms doesn't need to happen per-image and could be pulled out. I.e, you could find the `known_detection_algorithm` before looping the images. Do pipelines allow more than one detection algorithm? If so then I agree with coderrabit comment below. If not you could assert here that only one detection algo is found.

vanessavmac: > Do pipelines allow more than one detection algorithm?

I think at the moment we only have one detection algorithm per pipeline. But maybe @mihow could confirm and comment here?

mihow:
### Code review

Found 3 issues:

1. `filter_processed_images` skips images with both null detections and real unclassified detections. The `elif existing_detections.null_detections().exists(): continue` fires before the `elif existing_detections.filter(classifications__isnull=True).exists(): yield image` check. Any image that has a null detection (processed, nothing found) plus a real detection lacking classifications will be skipped — its real detections never get classified. The null check should only skip when there are no real detections.

https://github.com/RolnickLab/antenna/blob/5585785410120c6c3dbe973ec200ff1eebe77eb8/ami/ml/models/pipeline.py#L83-L97

2. `get_or_create_detection` uses `filter(source_image=source_image, bbox=None)` for null detections, which matches any existing null-bbox detection regardless of algorithm. If two pipelines (A and B) both process the same image with no detections, pipeline B's call finds pipeline A's null detection and reuses it. The null detection retains algorithm A, so `get_was_processed(algorithm_key=<B-detector>)` returns False even though pipeline B processed the image. Algorithm should be included in the lookup when `bbox` is None.

https://github.com/RolnickLab/antenna/blob/5585785410120c6c3dbe973ec200ff1eebe77eb8/ami/ml/models/pipeline.py#L421-L427

3. `was_processed` is added to `EventTimelineIntervalSerializer` but the TypeScript `ServerTimelineTick` type is not updated to include it. The field is silently ignored by the frontend; no component reads or renders it.

https://github.com/RolnickLab/antenna/blob/5585785410120c6c3dbe973ec200ff1eebe77eb8/ui/src/data-services/models/timeline-tick.ts#L1-L13

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #1105, opened 2026-01-29
Update date picker version and tweak layout logic

annavik:
## Summary

Comment from @mihow : 
> If the calendar picker appears above the field, then when you click prev/next the buttons jump up or down based on the length of the month, which makes it very tricky to go back more than a few months. A lot of the data is from 2022 which takes quite a while to get to with this jumper!

In this PR we fix this by updating date picker version and the layout logic. The updated version (see [separate PR](https://github.com/RolnickLab/nova/pull/11)) will allow year and month to be set from dropdown. This will make it a lot quicker if you want go back far in time. We also force the date picker to go below the trigger, to keep the position stable for navigation buttons. One little drawback is that scrolling will be needed, if there is not enough space below the trigger in the current view, to fit the calendar.

We also discussed setting the date from a text input. I had a quick look and that solution would require a bit more client side formatting with the current setup. I skipped for now since I think the fixes included will take care of the biggest problems.

### How to Test the Changes

Test interact with the date picker in the filter sidebar.

### Screenshots

Before:
Screenshot 2026-01-29 at 13 18 58

After:
Screenshot 2026-01-29 at 13 07 56

## Summary by CodeRabbit

* **Chores**
  * Updated nova-ui-kit dependency.

* **Bug Fixes**
  * Enhanced date picker popover positioning and calendar month reset functionality when closing the picker.
  * Fixed page footer stacking to prevent visual overlap with other page elements.

### DIFF
--- a/ui/src/design-system/components/select/date-picker.tsx
+++ b/ui/src/design-system/components/select/date-picker.tsx
@@ -1,7 +1,7 @@
 import { format } from 'date-fns'
 import { AlertCircleIcon, Calendar as CalendarIcon } from 'lucide-react'
 import { Button, Calendar, Popover } from 'nova-ui-kit'
-import { useState } from 'react'
+import { useEffect, useMemo, useState } from 'react'
 
 const dateToLabel = (date: Date) => {
   try {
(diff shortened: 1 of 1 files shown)


mihow:
Hey this works really well! I like the month & year pickers. Thanks @annavik. I tested in the collection form as well. One additional enhancement would be to open the calendar to the date that has already been selected. On my system if I select April 1, 1984, then open the calendar again it starts from Jan 2026.

image

PR #1114, opened 2026-02-04
fix: Properly handle async job state with celery tasks 

### LINKED ISSUE #1084: PSv2: Decouple the celery task state from the job state
## Summary

Take a look at this [CR comment](https://github.com/RolnickLab/antenna/pull/987/files#r2550749606). It's correctly pointing out that the job will be marked as SUCCESS when the task finishes, even though at that point only the queueing has happened. I started looking into this (this is how I came across fix #1062), but looks like the Job status is generally tied to the state of the celery task that runs it. 

We can add something like this:
```
class Job:
...
    def update_from_task(task, status, retval, ...):
       # default implementation does what
       # ami/jobs/tasks:update_job_status() does
       # But then MLJob can override to handle the async case. 
```

## Related Issues

Depends on #987.

carlosgjs:
## Summary

This pull request improves the reliability of job status updates by ensuring that a job is only marked as SUCCESS when all its stages are truly complete. It adds a guard to prevent premature SUCCESS status in cases where asynchronous workers are still processing, and includes new tests to verify this behavior. Additionally, it allows FAILURE and REVOKED states to be set immediately, regardless of stage progress.

**Job completion logic improvements:**

* Added a new `is_complete` method to the `Job` model, which checks that all stages have both finished processing (progress >= 1.0) and reached a final state (SUCCESS, FAILURE, or REVOKED). This method is used to determine if a job is truly complete before setting its status to SUCCESS.

* Updated the `update_job_status` Celery signal handler to guard against setting the job status to SUCCESS unless `is_complete` returns True, preventing race conditions where the job could be marked as complete before all stages finish. FAILURE and REVOKED states bypass this guard and are set immediately.

**Testing and validation:**

* Added new tests to verify that the SUCCESS status is only set when all stages are complete, and that FAILURE and REVOKED states are set immediately even if stages are incomplete. These tests help ensure the new guard logic works as intended.

**Code maintenance:**

* Updated imports in the Celery task handlers to include `JobState` for proper state comparisons.

### Related Issues

Closes #1084

### Testing

Before the fix, the job is incorrectly set as successful, which makes the progress bar green and enables the Retry button:
image
After the fix: the job remains as pending, so the progress bar is yellow and the Cancel button is enabled:
image

Additional testing of populating a collection to ensure the fix doesn't affect non-ML jobs:

image
image

Verified in the debugger the additional check is not triggered in this case:
image

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented jobs from being marked as successful before all stages finish; error/retry/revoke states still apply immediately.
  * Ensures final completion is based on full stage completion rather than task timing.

* **Tests**
  * Added tests validating the completion guard and immediate error-state behavior.

* **Documentation**
  * Removed an internal planning document about async job-status handling.


mihow (approved):
This looks good! If we need to we can use the new `async_api` backend type on the Job model too!

PR #1119, opened 2026-02-05
fix: address taxa-list PR review feedback

mihow:
## Summary

Addresses review feedback (mihow, Copilot, CodeRabbit) on the taxa-lists feature branch. Grouped into six logical commits.

**Bug fixes**
- `entity.ts` — `createdAt` getter was guarding on `updated_at` instead of `created_at`
- `utils.ts` — reverted `project_id` back to `project`; existing serializers (Device, Site, StorageSource, SourceImageCollection) all declare the field as `project`
- `useTaxaListDetails.ts` — generic type changed from `TaxaList` (client shape) to `ServerTaxaList` (raw API shape); added `projectId` to query key to prevent cross-project cache collisions
- `taxa-list-details.tsx` — Table `isLoading` was gated on `!id` which is always falsy (route param), so loading state never showed

**Security / permissions**
- Added `IsProjectMemberOrReadOnly` permission class (`ami/base/permissions.py`). Safe methods are open; unsafe methods (POST/DELETE) require the user to be a member of the active project. Scoped tighter than `IsActiveStaffOrReadOnly` (which blocks non-staff project members) and looser than guardian object-level perms (which don't cover M2M-to-project models like TaxaList).
- Applied to `TaxaListTaxonViewSet` — was `permission_classes = []` (completely open).

**Backend cleanup**
- `TaxonTaxaListFilter` docstring corrected: `include_descendants` defaults to `true`, not `false`
- `get_taxa_list`: added `from None` to suppress chained `DoesNotExist` traceback
- `create`: replaced `Taxon.objects.get` with `get_object_or_404` to close the race between serializer validation and the lookup

**UX**
- "Add taxon" button in header now gated on `canUpdate` (the per-row remove button already was)
- Add button disabled while the add mutation is in flight
- Remove confirm button disabled while the remove mutation is in flight
- Taxa-lists page result count uses the paginated `total` from the API instead of current-page length

**Naming / dead code**
- Deduplicated `MESSAGE_MESSAGE_REMOVE_TAXA_LIST_TAXON_CONFIRM` → `MESSAGE_REMOVE_TAXA_LIST_TAXON_CONFIRM`
- Removed unused `API_ROUTES` import in `new-entity-dialog.tsx`
- Removed dead expression in `useSidebarSections.tsx` (computed value never used)
- Added missing `setMainBreadcrumb` to `useEffect` dep array in `breadcrumbs.tsx`

**Accessibility**
- Three dialog sites (`taxa-list-details`, `species`, `occurrences`) called `e.preventDefault()` in `onOpenAutoFocus` without transferring focus. Added explicit `.focus()` on the dialog content in each.

## Test plan

- [ ] `ami.main.tests.TaxaListTaxonAPITestCase` and `TaxaListTaxonValidationTestCase` — all 13 pass with non-staff project-member users
- [ ] Manually: create a taxa list as a non-staff project member, add and remove taxa
- [ ] Manually: confirm a non-member cannot POST/DELETE on the taxa-list taxon endpoint
- [ ] Visually confirm the "Add taxon" button disappears for read-only users
- [ ] Visually confirm Add and Remove buttons stay disabled while their requests are in flight

### DIFF
--- a/ami/base/permissions.py
+++ b/ami/base/permissions.py
@@ -88,6 +88,12 @@ class IsProjectMemberOrReadOnly(permissions.BasePermission):
         if request.method in permissions.SAFE_METHODS:
             return True
 
+        if not request.user or not request.user.is_authenticated:
+            return False
+
+        if request.user.is_superuser:  # type: ignore[union-attr]
+            return True
+
         # view must provide get_active_project (i.e. use ProjectMixin)
         get_active_project = getattr(view, "get_active_project", None)
         if not get_active_project:


ui/src/pages/occurrences/occurrences.tsx
```diff
@@ -281,6 +281,7 @@ const OccurrenceDetailsDialog = ({
         onOpenAutoFocus={(e) => {
           /* Prevent tooltip auto focus */
           e.preventDefault()
+          ;(e.currentTarget as HTMLElement).focus()
```
annavik: If we want to enable auto focus again, the correct way would be to skip `onOpenAutoFocus`. The code in this method will now first disable, then enable again, so a bit unnecessary.

I disabled it because I was annoyed to see the rank tooltip show up on focus :) However it's better for accessibility to have it, so maybe not a good idea.

I'll push an update here.

annavik:
Thanks for the fixes @mihow !! I pushed some tweaks and also tested things out again here. One thing I noticed in this branch is that if I'm a superuser, but not a member of the project, I cannot add or remove taxa from lists. This is not the case in the branch `feat/taxa-lists`.

Screenshot 2026-02-06 at 17 02 44

mihow:
> Thanks for the fixes @mihow !! I pushed some tweaks and also tested things out again here. One thing I noticed in this branch is that if I'm a superuser, but not a member of the project, I cannot add or remove taxa from lists. This is not the case in the branch `feat/taxa-lists`.
> 

Interesting discovery! I will look over the permissions for taxa lists as a whole again. I like the error handling you have now! "You do not have permission to perform this action" in the bar above the Add Taxon form! That is slick.

I also opened this ticket to look more into permissions for all data types that can be assigned to multiple projects. 
https://github.com/RolnickLab/antenna/issues/1120

Issue #1158, opened 2026-02-27
Expose session regrouping as a job stage with stats


mihow:
## Context

When a data storage sync job runs and adds new captures, session regrouping happens automatically via `deployment.save(regroup_async=False)`. However, the results of this regrouping (sessions created, merged, or updated) are not visible in the job progress UI.

## Current behavior

- Sync job shows "Data storage sync" stage with total file count
- Session regrouping happens silently inside `save()`
- No visibility into how many sessions were created or modified
- If grouping fails or produces unexpected results, there's no indication in the job

## Desired behavior

- Add a "Regroup sessions" stage to the sync job progress
- Report stats: sessions created, sessions updated, total sessions
- Log warnings for edge cases (e.g., very long sessions, sessions with single images)

## Related

- PR #904 (Fix incorrect session grouping) already refactors `group_images_into_events`
- The `group_images_into_events()` function in `ami/main/models.py` would need to return stats

PR #1159, opened 2026-02-27
Fix 2-digit year timestamp parsing, sync error reporting, and captures columns

mihow:
## Summary

- **2-digit year timestamp parsing**: Some cameras produce filenames with YYMMDDHHMMSS (12-digit) timestamps instead of the expected YYYYMMDDHHMMSS (14-digit). These were silently returning NULL, causing images to show "Dec 31, 1969" in the UI and preventing session grouping. Added a new regex pattern and `%y` format handling.

- **Skip null-timestamp images and report failures**: Images with unparseable timestamps are now skipped during sync rather than imported with NULL. They can't be grouped into events anyway (`group_images_into_events` already excluded them), so importing them just created orphaned records. Per-image processing is wrapped in try/except, a `failed` counter is tracked and reported as a "Failed" stage param in the job progress UI. Warns on suspicious pre-2000 timestamps. If timestamp parsing is improved later, a re-sync will pick up previously skipped images.

- **Filename/path columns**: Added two new columns to the captures table (hidden by default, toggled via column settings). Exposes the `path` field in the list serializer and derives `filename` from it on the frontend.

## Related

- Created #1158 to track exposing session regrouping as a job stage with stats (deferred to align with PR #904)

### Before: Null timestamps showing epoch date
Before - null timestamps

### After: Correct timestamps parsed from 2-digit year filenames
After - correct timestamps

### Filename and Path columns (toggled on via column settings)
Filename and path columns

### Sync job error reporting (Failed stage param in job progress)
Sync job error reporting

## Test plan

- [x] Backend unit tests pass for new timestamp patterns
- [x] Pre-commit hooks pass
- [x] Verify 2-digit year filenames parse correctly after re-sync
- [x] Verify sync job shows "Failed" param in job detail view
- [x] Toggle filename/path columns on in captures table

## Release Notes

* **New Features**
  * Added path and filename information to capture records with display columns in the interface
  * Enabled sorting captures by path
  
* **Bug Fixes**
  * Improved error handling during file synchronization with per-item failure tracking and logging
  * Added timestamp validation to prevent invalid data processing
  * Extended timestamp parsing to support 2-digit year formats

* **Tests**
  * Added test coverage for 2-digit year timestamp parsing formats

### DIFF
--- a/.agents/AGENTS.md
+++ b/.agents/AGENTS.md
@@ -650,6 +650,16 @@ images = SourceImage.objects.annotate(det_count=Count('detections'))
 - Use `@shared_task` decorator for all tasks
 - Check Flower UI for debugging: http://localhost:5555
 
+### E2E Testing & Monitoring Async Jobs
+
+Run an end-to-end ML job test:
+```bash
+docker compose run --rm django python manage.py test_ml_job_e2e \
+  --project 18 --dispatch-mode async_api --collection 142 --pipeline "global_moths_2024"
+```
+
+For monitoring running jobs (Django ORM, REST API, NATS consumer state, Redis counters, worker logs, etc.), see `docs/claude/reference/monitoring-async-jobs.md`.
+
 ### Running a Single Test
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -144,9 +144,13 @@ pip install -r requirements/local.txt
 
 #### Helpful Commands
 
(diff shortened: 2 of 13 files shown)


mihow:
I'm also unsure why we would ever want to import images with NULL timestamps. That will require a data migration, but seems like a good integrity change.

PR #1186, opened 2026-03-25
fix: handle null occurrence determination crashes in UI

mihow:
- [x] Update `OccurrenceQuerySet.valid()` to also exclude occurrences with `determination=null`
- [x] Add `TestOccurrencesWithoutDetermination` tests to verify behavior
- [x] Tests pass and existing threshold filter tests still pass

---

🔒 GitHub Advanced Security automatically protects Copilot coding agent pull requests. You can protect all pull requests by enabling Advanced Security for your repositories. [Learn more about Advanced Security.](https://gh.io/cca-advanced-security)

### DIFF
--- a/ui/src/data-services/models/occurrence-details.ts
+++ b/ui/src/data-services/models/occurrence-details.ts
@@ -126,7 +126,7 @@ export class OccurrenceDetails extends Occurrence {
 
   getDetectionInfo(id: string) {
     const detection = this._occurrence.detections.find(
-      (d: any) => `${d.id}` === id,
+      (d: any) => `${d.id}` === id
     )
 
     const classification = detection?.classifications?.[0]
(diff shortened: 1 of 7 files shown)


mihow:
@copilot instead of this fix, let's filter out occurrences without determinations from the occurrences list responses - currently they are filtered if apply default filters is true, but they should always be filtered. If we don't already have a QuerySet method on Occurrences, consider adding one called `valid()`. Occurrence.objects.valid()`

mihow:
Will filter more aggressively on backend and add a periodic data integrity check #1188

mihow:
@annavik and I discussed bringing this back, but in a simpler manner. The UI should anticipate that occurrences can have no detections or determination. But instead of adding null handling throughout the UI, just filter them out after the API response first. This will ensure when we switch to shared types between the BE & FE that the type checker is happy (Occurrences can always technically have no detections or no determination during intermediate processing stages), but also we don't add a lot of extra complexity for an edge case.

PR #1188, opened 2026-03-25
feat: add check_occurrences for occurrence data integrity

mihow:
## Summary

Adds a reusable data integrity check framework under `ami/main/checks/`. Each check is a pair of functions: a `get_*` queryset (side-effect free, composable) and a `reconcile_*` action returning an `IntegrityCheckResult`. Future checks add a new module under `checks/` and plug into the management command and Celery task.

This PR ships one check: **occurrences with classifications but no determination set** — the partial-save failure observed in the demo environment.

## Components

- **`ami/main/checks/occurrences.py`** — `get_occurrences_missing_determination()` and `reconcile_missing_determinations()`, scoped by `project_id`, `job_id`, or explicit `occurrence_ids`. Each row is fixed in its own `try/except` so one failure doesn't abort the batch.
- **`ami/main/management/commands/check_data_integrity.py`** — `--dry-run` (default) / `--no-dry-run`, plus `--project`, `--job`.
- **`ami/main/tasks.py`** — `check_data_integrity` Celery task, schedulable via django_celery_beat in the Django admin.
- **`ami/main/tests.py::TestIntegrityChecks`** — 7 tests against the real CI database.

## CLI

```bash
# Report-only (safe default)
manage.py check_data_integrity --project 5

# Apply fixes
manage.py check_data_integrity --project 5 --no-dry-run

# Scope to a specific job
manage.py check_data_integrity --job 123 --no-dry-run
```

## Tests

`TestIntegrityChecks` covers:
- `get_*` includes only occurrences with classifications but no determination
- `get_*` excludes occurrences with no classifications at all
- `reconcile_*` in `dry_run=True` mode reports counts but doesn't modify the DB
- `reconcile_*` in `dry_run=False` mode repairs nulled determinations
- Project scoping: other projects' rows are not touched
- Explicit occurrence-id scoping: only listed rows are fixed
- Clean state returns zero checked

All 7 tests pass against the CI Postgres stack.

## Related

Alternate implementation: #1185 (`fix/null-determination-resilience`). This PR adopts the same `get_*` / `reconcile_*` separation and `ReconcileResult`-style return, but lays it out as an extensible package (`checks/`) rather than a single file, with the expectation that future integrity checks add one module per domain.

### DIFF
--- a/ami/main/checks/__init__.py
+++ b/ami/main/checks/__init__.py
@@ -1,0 +1,20 @@
+"""Data integrity checks for the main app.
+
+Each module under this package defines one or more integrity check pairs:
+a ``get_*`` function returning a queryset of affected rows, and a
+``reconcile_*`` function that attempts to repair them. Both are exported
+from this package so callers can compose individual checks from
+management commands, post-job hooks, and periodic Celery tasks.
+"""
+
+from ami.main.checks.occurrences import (
+    IntegrityCheckResult,
+    get_occurrences_missing_determination,
+    reconcile_missing_determinations,
+)
+
+__all__ = [
+    "IntegrityCheckResult",
+    "get_occurrences_missing_determination",
+    "reconcile_missing_determinations",
+]
--- a/ami/main/checks/occurrences.py
+++ b/ami/main/checks/occurrences.py
@@ -1,0 +1,99 @@
+"""Integrity checks for Occurrence records."""
+
+import dataclasses
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class IntegrityCheckResult:
+    checked: int = 0
+    fixed: int = 0
+    unfixable: int = 0
+
+
+def get_occurrences_missing_determination(
+    project_id: int | None = None,
+    job_id: int | None = None,
+):
+    """Return occurrences that have classifications but no determination set.
+
+    Occurrences without any classifications are excluded because they
+    legitimately have no determination yet.
+    """
+    from ami.main.models import Occurrence
+
+    qs = Occurrence.objects.filter(
+        determination__isnull=True,
+        detections__classifications__isnull=False,
+    ).distinct()
+
+    if project_id is not None:
+        qs = qs.filter(project_id=project_id)
+
+    if job_id is not None:
+        from ami.jobs.models import Job
+
+        job = Job.objects.get(pk=job_id)
+        if job.pipeline_id:
+            qs = qs.filter(
+                detections__classifications__algorithm__in=job.pipeline.algorithms.all(),
+                project_id=job.project_id,
+            )
+
+    return qs
+
+
+def reconcile_missing_determinations(
+    project_id: int | None = None,
+    job_id: int | None = None,
+    occurrence_ids: list[int] | None = None,
+    dry_run: bool = True,
+) -> IntegrityCheckResult:
+    """Find occurrences missing determinations and repair them.
+
+    Re-runs ``update_occurrence_determination`` on each affected row so the
+    best available identification or prediction is promoted to the
+    determination field. Occurrences that can't be resolved (e.g. no viable
+    prediction) are counted as ``unfixable``.
+    """
+    from ami.main.models import Occurrence, update_occurrence_determination
+
+    if occurrence_ids is not None:
+        occurrences = Occurrence.objects.filter(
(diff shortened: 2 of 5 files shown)


mihow:
See another backend implementation at https://github.com/RolnickLab/antenna/pull/1185

mihow:
**Claude says:** Naming alignment note — we're landing a sibling beat task in #1227 (`jobs_health_check` umbrella, stale-job reconciler as the first sub-check) that intentionally adopts the `get_*` / `reconcile_*` split and `{checked, fixed, unfixable}` return shape from this PR.

One small asymmetry worth resolving eventually, in this PR or a follow-up:

- **Task name convention.** #1227 uses noun-phrase for the beat task (`jobs_health_check`) and reserves verb-first for management commands (`manage.py check_jobs` — planned follow-up). Here the beat task and command share `check_data_integrity`. Suggest renaming the task to `data_integrity_check` (keeping `manage.py check_data_integrity` as-is) so both PRs read symmetrically in flower / admin / logs.

- **Scheduling asymmetry (intentional, not a request to change).** #1227 auto-schedules via migration because stale-job reconciliation is a critical self-healing invariant; this PR leaves scheduling to admins because the integrity check is opt-in cleanup. Worth a line in this PR body so the difference is explicit.

Happy to address either of these on your side when you pick this up — flagging now so we stay aligned.

mihow:
**Claude says (on behalf of @mihow):** Heads-up — proposing a `JobLog` retention check as the second module under this framework once it lands. Context in #1259 (comment: https://github.com/RolnickLab/antenna/pull/1259#issuecomment-4348713570).

Shape would mirror the occurrences module: `get_joblog_retention_violations(project_id, max_rows_per_job, max_age_days)` + `reconcile_joblog_retention(..., dry_run=True)` returning `IntegrityCheckResult`. Plugs into `check_data_integrity` and the Celery beat hook this PR adds, no new surface.

Mostly a signal that the `checks/` package is the right home for this kind of cross-cutting hygiene work — happy to open a follow-up issue once this merges.

mihow:
another check: 
- look for missing occurrence thumnbnails (We have a management command for that)
- look for captures without an event

mihow:
**Claude says:** Flagging another follow-up check that's emerging from review on PR #1331 (thumbnail serializer rework), in case it's useful for shaping the framework here.

PR #1331 drops the synchronous `default_storage.exists()` HEAD that `SourceImage.find_or_generate_thumbnail_for_label` used to run on every cold-path request. The cached `SourceImageThumbnail` row is the warm signal we trust now — orphan rows (DB row points at a storage blob that was deleted out of band: TTL eviction, manual cleanup, backing-store outage) are explicitly handed off to a periodic reconciler instead of being caught per-request.

That reconciler is the natural shape this framework wants:

- `ami/main/checks/thumbnails.py::get_orphan_thumbnails(batch_size=N, last_checked_before=...)` — returns rows whose blob is missing (cursored by a `last_checked_at` once the table grows; sample-first while it's small).
- `reconcile_orphan_thumbnails(..., dry_run=True)` — deletes orphan rows, returns `IntegrityCheckResult(checked, fixed, unfixable)`. CASCADE FK + the existing `pre_delete` signal on `SourceImageThumbnail` clean the (already-missing) storage blob.
- Plugs into the same `check_data_integrity` management command + Celery beat hook this PR wires up.

Two things on the framework side that would shape this nicely if they're worth deciding now rather than later:

1. **Umbrella beat task vs per-check task.** `ami/jobs/tasks.py::jobs_health_check` composes `_safe_run_sub_check`-wrapped sub-checks with one failure isolated per branch. A `main_data_integrity_check` umbrella in `ami/main/tasks.py` that grows by sub-check (occurrences → joblog retention #1277 → orphan thumbnails → future) might be a cleaner shape than one beat entry per check. Otherwise every new check adds a beat row.
2. **Where the cursor lives** for `get_*` checks that scan an unbounded table. The occurrences check here is naturally bounded by classifications-without-determination. JobLog retention (#1277) and orphan thumbnails both potentially scan large tables; both want a `last_checked_at` column (or equivalent) so a tick costs O(batch) instead of O(table). Worth picking a convention before two checks pick different ones.

Will write up a dedicated issue for the thumbnails reconciler once #1331 lands (same shape as #1277). Mirrored a pointer on #1277 since it's the closest prior art for "specific check using this framework."

Cross-ref: #1277, #1331.

PR #1201, opened 2026-04-01
feat(ui): API key management and client info for processing services

mihow:
## Summary

- Adds "Generate API Key" button with copy/reveal UX in processing service details
- Shows "Authentication" section with API key prefix and service mode (push/pull)
- Shows "Last Known Worker" section with hostname, software, platform, and IP from client_info
- Passes projectId to processing service details hook for project-scoped queries
- Makes endpoint_url optional for pull-mode services that register themselves
- Fixes async service status to show ONLINE when heartbeat is recent

Depends on #1194 (backend auth).

## Permission and ownership notes for future UI work

The following considerations were surfaced during E2E testing and should inform the UI design:

**API key scoping**: The API key is tied to the ProcessingService (PS), not to a specific project. Since a PS has a M2M with projects, one API key grants access to all linked projects. Removing a PS from a project immediately denies access (checked live on each request).

**Who can generate/revoke keys**: Currently any project member who can list PSes can generate keys. This should probably be restricted to project admins/managers — key revocation affects all projects the PS serves.

**User-created vs global PSes**: Consider adding a `created_by` field to ProcessingService. Global PSes (admin-created) can be managed by admins; user-created PSes only by their creator + project admins. This prevents unauthorized key revocation.

**Multi-project PS management**: A user who creates a PS and belongs to multiple projects should be able to add the same PS to other projects they're a member of. The UI could show "My Processing Services" for self-service linking.

**Org-level ownership**: When the organization-level object is added above projects, PSes could also be org-owned, simplifying multi-project sharing within an organization.

## Test plan
- [ ] `cd ui && yarn build`
- [ ] Verify processing service details dialog shows Authentication section
- [ ] Generate API key and verify copy/reveal/regenerate flow
- [ ] Verify Last Known Worker section appears after a worker heartbeat

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/ui/src/data-services/hooks/processing-services/useGenerateAPIKey.ts
+++ b/ui/src/data-services/hooks/processing-services/useGenerateAPIKey.ts
@@ -1,0 +1,39 @@
+import { useMutation, useQueryClient } from '@tanstack/react-query'
+import axios from 'axios'
+import { API_ROUTES, API_URL } from 'data-services/constants'
+import { getAuthHeader } from 'data-services/utils'
+import { useUser } from 'utils/user/userContext'
+
+interface GenerateAPIKeyResponse {
+  api_key: string
+  prefix: string
+  message: string
+}
+
+export const useGenerateAPIKey = (projectId?: string) => {
+  const { user } = useUser()
+  const queryClient = useQueryClient()
+  const params = projectId ? `?project_id=${projectId}` : ''
+
+  const { mutateAsync, isLoading, isSuccess, error, data } = useMutation({
+    mutationFn: (id: string) =>
+      axios.post<GenerateAPIKeyResponse>(
+        `${API_URL}/${API_ROUTES.PROCESSING_SERVICES}/${id}/generate_key/${params}`,
+        undefined,
+        {
+          headers: getAuthHeader(user),
+        }
+      ),
+    onSuccess: () => {
+      queryClient.invalidateQueries([API_ROUTES.PROCESSING_SERVICES])
+    },
+  })
+
+  return {
+    generateAPIKey: mutateAsync,
+    isLoading,
+    isSuccess,
+    error,
+    apiKey: data?.data.api_key,
+  }
+}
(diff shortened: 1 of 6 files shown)


mihow:
Cross-ref with #1222 (per-job NATS lifecycle logging).

Now that #1201 will tie each request back to a specific `ProcessingService` via API key auth, we can extend the per-job log to surface **which** worker is actually consuming a job's messages. Today #1222 shows aggregate NATS state in the job log (`delivered=N`, `ack_floor=N`, `num_ack_pending=N`, `num_redelivered=N`), which is enough to see *that* delivery is happening but not *who* is doing the work — so if two workers are attached and one is stuck, the job log can't tell you which one.

Concrete suggestion for a small follow-up PR once #1201 + #1194 are in:

In the pull view that calls `TaskQueueManager.reserve_tasks(job_id, count)`, after a successful non-empty fetch, emit one line to the job's logger:

```python
if tasks:
    job = Job.objects.get(pk=job_id)
    job.logger.info(
        f"Worker '{request.user.processing_service.name}' reserved {len(tasks)} tasks from job {job_id}"
    )
```

Gated on `len(tasks) > 0` so idle polls (every few seconds, per worker) stay silent — otherwise a job with 4 workers polling would drown its own log. The existing per-message logs in `reserve_tasks` stay on the module logger only.

With this plus #1222 in place, a triage reader looking at a job's UI log would see the full story:

```
Created NATS stream job_1267
Created NATS consumer job-1267-consumer (max_deliver=5, ack_wait=30s, ...)
Queuing 18 images to NATS stream for job '1267'
Worker 'Classifier Worker A' reserved 8 tasks from job 1267
Worker 'Classifier Worker B' reserved 8 tasks from job 1267
Worker 'Classifier Worker A' reserved 2 tasks from job 1267
Finalizing NATS consumer job-1267-consumer before deletion (delivered=18 ack_floor=18 num_pending=0 num_ack_pending=0 num_redelivered=0)
Deleted NATS consumer job-1267-consumer for job '1267'
```

Not in-scope for #1201 itself — just flagging so it doesn't get lost after the auth plumbing lands.

Issue #1219, opened 2026-04-11
`async_api` jobs killed by transient Redis errors during `update_state` — `RedisError` and "state actually missing" are conflated into a single fatal path


mihow:
> **Note on scope**
>
> This ticket is the **code-path** half of the bug: a single transient Redis error
> kills an async job irrecoverably because of how `AsyncJobStateManager.update_state()`
> and its caller handle `RedisError`. There is a **sibling config-path bug** (missing
> `socket_keepalive` on the Django Redis cache connection) captured separately in
> `2026-04-10-antenna-new-issue-redis-cache-keepalive.md`. Fixing the config bug
> reduces the frequency of the transient; fixing this code-path bug ensures the
> remaining transients don't destroy user data.
> 
> Both should be fixed. Either one alone is insufficient.

## Summary

In `ami/ml/orchestration/async_job_state.py`, `AsyncJobStateManager.update_state()` returns `None` for **two very different conditions**:

1. `redis.exceptions.RedisError` raised during the pipeline call (transient: connection reset, timeout, broker restart)
2. `total_raw is None` — the `job:<id>:pending_images_total` key genuinely doesn't exist (state was actually cleaned up)

```python
# async_job_state.py
try:
    ...
    results = pipe.execute()
except RedisError as e:
    logger.error(f"Redis error updating job {self.job_id} state: {e}")
    return None                                   # ← transient

...
if total_raw is None:
    return None                                   # ← state actually gone
```

The caller in `ami/jobs/tasks.py:process_nats_pipeline_result()` treats both the same way — no retry, no distinguishing log, immediate call to `_fail_job`:

```python
# tasks.py, first call (before save_results)
progress_info = state_manager.update_state(processed_image_ids, stage="process", ...)
if not progress_info:
    _ack_task_via_nats(reply_subject, logger)
    _fail_job(job_id, "Redis state missing for job")
    return

# tasks.py, second call (after save_results + NATS ack)
progress_info = state_manager.update_state(processed_image_ids, stage="results")
if not progress_info:
    _fail_job(job_id, "Redis state missing for job")
    return
```

`_fail_job` sets job status to FAILURE, marks `finished_at`, and calls `cleanup_async_job_resources`, which deletes the Redis state AND the NATS stream + consumer. So a single 153ms connection blip to Redis doesn't just incorrectly fail the job — it **destroys the job's work queue**, making recovery impossible without re-running the whole job from scratch.

## Observed production incident

[...]

PR #1228, opened 2026-04-14
fix(celery): update worker concurrency defaults

mihow:
## Summary

- Add explicit `CELERY_WORKER_CONCURRENCY = env.int("CELERY_WORKER_CONCURRENCY", default=8)` to `config/settings/base.py`, next to the existing `CELERY_WORKER_PREFETCH_MULTIPLIER` / `CELERY_WORKER_ENABLE_PREFETCH_COUNT_REDUCTION` block.
- Overridable per deployment via the `CELERY_WORKER_CONCURRENCY` env var.
- Production should override to **16** — committed as an example in `.envs/.production/.django-example` with guidance.
- Local dev can lower to **2–4** on memory-constrained laptops — guidance added to `.envs/.local/.django`.

## Why

The default celery worker concurrency when the setting is unset is `os.cpu_count()`. On the current production celery worker host (8 cores) this means an 8-process prefork pool. The dominant tasks on the `antenna` queue — `process_nats_pipeline_result` and `create_detection_images` — are DB/Redis-bound rather than CPU-bound: each task spends most of its time waiting on postgres/pgbouncer and Redis round-trips, not crunching numbers.

Direct observation during a high-throughput `async_api` job:

- Queue ingress rate ~640/min (two active jobs with 280–900 images each; redelivery amplifies ingress further when any of the bugs around #1219/#1221 trigger)
- Single-consumer drain rate ~283/min at 8-way concurrency
- Net accumulation ~360/min; the queue grew into the tens of thousands of messages over ~30 min
- Container-level utilisation at 8-way concurrency: ~1.6% CPU, ~1.3 GiB of 29 GiB resident. Plenty of headroom.
- RabbitMQ `consumer_utilisation` on the antenna queue: ~0.0016, i.e. the single AMQP consumer's prefetch window is fully occupied essentially all the time. This is the "worker pool too small" signature, not a broker-side issue.

Raising the prefork pool size directly addresses the bottleneck. A hotfix override of 16 was applied in production via the env var ahead of this PR and confirmed to drain the backlog on the active jobs.

## Why default=8, prod override=16

The previous revision of this PR used `default=16` everywhere. Review flagged the blast radius across local dev laptops and the smaller staging/demo VMs — 16 prefork workers = 16 separate Python processes with imports + persistent DB connections, which is a real memory bump where it isn't needed.

The current default of **8** is Celery's typical out-of-the-box behaviour on an 8-core host (matches `os.cpu_count()` there) and a safer memory footprint everywhere else. Production overrides to **16** via env var — committed as an example in `.envs/.production/.django-example` with the rationale inline. This keeps the bottleneck fix in place where it matters without making every other environment pay for it.

16 is the smallest power-of-2 step above 8 that roughly matches the empirical gap between ingress and drain on the production incident that motivated this PR, without risking pgbouncer saturation. A larger default can be considered once we have measured postgres connection-pool headroom (see "what we still need to verify" below).

## What this does not change

- **Prefetch multiplier** stays at `1` — that was already set and fairness behaviour is unchanged.
- **Routing / queue topology** is unchanged. Splitting the `antenna` queue into a dedicated "ingest fast path" vs "housekeeping / status-check" queue is a larger follow-up, filed separately.
- **Pool class** stays `prefork`. Switching to `gevent` for this queue may give much higher effective concurrency on an IO-bound workload, but every task on this queue would need to be audited for gevent-safety (blocking C extensions, thread-locals in PyTorch paths, etc.) first. Out of scope here.

## Deploy note

The production host is already running with `CELERY_WORKER_CONCURRENCY=16` in its env file from the earlier hotfix — lowering the settings default to 8 does **not** regress prod. Confirm the env var is still set at next deploy; if for any reason it isn't, prod will drop to 8 until the env is fixed.

- Postgres / pgbouncer connection pool usage after deploy — 16 prefork workers × persistent connections should be well within pgbouncer's `default_pool_size`, but worth confirming under load.
- Whether the 16-prod override is still the right ceiling, or whether larger backlogs warrant raising it (gated on pgbouncer headroom).
- Whether this change exposes any new memory-pressure pattern at peak load (current `--max-tasks-per-child=100` / `--max-memory-per-child=2 GiB` already bound each process).

## Related

- #1219 — code-path brittleness that lets a single transient Redis error mark an active job FAILURE and delete state (independent of this PR).
- #1221 — django-redis cache connection keepalive fix that reduces how often the #1219 path triggers (independent of this PR).
- #1231 — code-path fix for #1219 (retry transient Redis errors instead of failing the job).

## Summary by CodeRabbit

* **New Features**
  * Added a configurable worker concurrency setting to control parallel background task processing (default: 8, adjustable via environment variable).

* **Documentation**
  * Included example environment entries for local (commented example: 4) and production (example: 16) configurations.

### DIFF
--- a/.envs/.local/.django
+++ b/.envs/.local/.django
@@ -19,6 +19,11 @@ NATS_URL=nats://nats:4222
 CELERY_FLOWER_USER=QSocnxapfMvzLqJXSsXtnEZqRkBtsmKT
 CELERY_FLOWER_PASSWORD=BEQgmCtgyrFieKNoGTsux9YIye0I7P5Q7vEgfJD2C4jxmtHDetFaE2jhS7K7rxaf
 
+# Celery worker concurrency (prefork pool size).
+# Default in settings is 8. Lower to 2-4 on memory-constrained dev laptops —
+# each prefork worker is a separate Python process with its own DB connection.
+# CELERY_WORKER_CONCURRENCY=4
+
 # RabbitMQ
 CELERY_BROKER_URL=amqp://rabbituser:rabbitpass@rabbitmq:5672/
 CELERY_RESULT_BACKEND=rpc:// # Use RabbitMQ for results backend
--- a/.envs/.production/.django-example
+++ b/.envs/.production/.django-example
@@ -20,6 +20,13 @@ REDIS_URL=redis://redis:6379/0
 # Celery
 # ------------------------------------------------------------------------------
 
+# Worker concurrency (prefork pool size). Settings default is 8.
+# Recommended for production: 16 on an 8-core host — process_nats_pipeline_result
+# and create_detection_images are DB/Redis-bound, so ~2x cpu_count roughly matches
+# observed drain vs ingress on the antenna queue without saturating pgbouncer.
+# Tune higher only after confirming postgres connection-pool headroom.
+CELERY_WORKER_CONCURRENCY=16
+
 # Flower
 CELERY_FLOWER_USER=
 CELERY_FLOWER_PASSWORD=
--- a/config/settings/base.py
+++ b/config/settings/base.py
@@ -393,6 +393,13 @@ CELERY_REDIS_BACKEND_HEALTH_CHECK_INTERVAL = 30  # Check health every 30s
 CELERY_WORKER_PREFETCH_MULTIPLIER = 1
 CELERY_WORKER_ENABLE_PREFETCH_COUNT_REDUCTION = True
 
+# Worker concurrency (prefork pool size)
+# https://docs.celeryq.dev/en/stable/userguide/configuration.html#worker-concurrency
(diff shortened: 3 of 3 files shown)


mihow:
**Claude says:** Addressed the nitpick about env-template documentation in [07a1c2b2](https://github.com/RolnickLab/antenna/pull/1228/commits/07a1c2b2). Added `CELERY_WORKER_CONCURRENCY` with guidance to both `.envs/.local/.django` (suggesting 2-4 on memory-constrained dev laptops) and `.envs/.production/.django-example` (recommending 16 for prod, with rationale inline). Also took the opportunity to lower the settings default from 16 → 8 after review feedback about blast radius on smaller deployments; prod keeps its 16 via the existing env var override.

Issue #928, opened 2025-08-27
Investigate Database optimization options to better scale up our database operations


f-PLT:
## Summary

Our database is becoming a kind of bottleneck for some operations. It is getting quite big, with very large tables. We therefore need to explore ways of optimizing our database so it doesn't become a full blocker down the line

## Implementation Details

I'll be hosting my own local antenna, starting out from a full database backup in order to replicate the same scale.

This task will be a slow one, as it requires a lot of continuous time. Hopefully, the VISS collaboration can also help with this.

PR #1251, opened 2026-04-17
Add composite index main_detection(occurrence_id, timestamp DESC)

mihow:
## Summary

Add a composite index on `main_detection (occurrence_id, timestamp DESC)` to speed up the `/captures/` list endpoint, which joins `main_detection` and `main_occurrence`, filters by `project_id`, and sorts by timestamp DESC.

Without this index PostgreSQL falls back to a nested-loop join that scans ~517k detection rows before filtering by project, then sorts the full set.

## Measured impact (external perf report, warm DB, caching disabled)

| Stage | Exec time | vs. baseline |
|---|---|---|
| Before — nested loop, ~517k rows scanned | ~1,470 ms | baseline |
| After — parallel hash join + index-ordered scan | ~251 ms | **−83%** |

This is a straight-up plan change: the query planner picks a parallel hash join because it can read detections ordered by timestamp for each occurrence directly from the index, instead of collecting all qualifying detections and sorting.

## Implementation

- `ami/main/models.py` — add `Meta.indexes = [Index(fields=[\"occurrence\", \"-timestamp\"], name=\"detection_occurrence_ts_desc\")]` on `Detection`.
- `ami/main/migrations/0084_index_detection_occurrence_timestamp.py` — uses `django.contrib.postgres.operations.AddIndexConcurrently` with `atomic = False`.

Verified locally that the migration emits the correct SQL:

```sql
CREATE INDEX CONCURRENTLY \"detection_occurrence_ts_desc\"
ON \"main_detection\" (\"occurrence_id\", \"timestamp\" DESC);
```

…and the resulting index is visible on the table:

```
\"detection_occurrence_ts_desc\" btree (occurrence_id, \"timestamp\" DESC)
```

## Why AddIndexConcurrently

`main_detection` is one of the largest tables in the system. A plain `CREATE INDEX` takes an `ACCESS EXCLUSIVE`-level table lock while it builds, blocking writes for the duration. `CREATE INDEX CONCURRENTLY` trades a longer build for an online build that only needs a `SHARE UPDATE EXCLUSIVE` lock — safe during production traffic.

Gotcha: `AddIndexConcurrently` can't run inside a transaction, so the migration sets `atomic = False`. If the build fails partway, the index is left in an invalid state and must be dropped manually (`DROP INDEX CONCURRENTLY`) before re-running — standard for concurrent index builds.

## Test plan

- [x] `python manage.py migrate main` applies cleanly in CI stack
- [x] `sqlmigrate` emits `CREATE INDEX CONCURRENTLY`
- [x] Post-apply `\\d main_detection` shows the composite index
- [ ] Confirm on staging before prod rollout: `EXPLAIN (ANALYZE, BUFFERS)` on `/captures/` query with a large project, compare plan before/after

## Related

Pairs with #1250 (require `project_id` on hot list endpoints) as the second half of the \"fix the major culprits now\" pass. Neither depends on the other — can merge independently.

## Summary by CodeRabbit

* **Performance**
  * Improved speed and responsiveness when viewing detection records grouped by occurrence, especially where the newest detections are shown first.
  * Queries that rely on ordering detections within each occurrence should now run more efficiently.

* **Bug Fixes**
  * Reduced the chance of slow or timing-out database operations during index creation on large datasets.


mihow:
@coderabbitai review this please if you haven't already

mihow:
I have tested locally on a large dataset, will test in staging & demo env next

mihow:
**Claude says:** Worth a bit more verification before merge — index itself is low-risk and the migration setup (`AddIndexConcurrently` + `atomic = False`) looks right, but a couple things to confirm so we know it's actually moving the needle:

1. **Confirm the index is being picked by the real query.** The benchmark in the PR shows a big plan change (nested loop → parallel hash join), but the SQL it was run against isn't shown. The `/captures/` list path (`SourceImageViewSet`) mostly filters detections by `source_image_id`, not `occurrence_id`, so the new index probably isn't relevant there. Where it does help is the Occurrence detail prefetch (`views.py:1233`) which explicitly does `Detection.objects.order_by("-timestamp")` keyed on `occurrence_id` — that's a perfect match for `(occurrence_id, timestamp DESC)`. Worth running `EXPLAIN (ANALYZE, BUFFERS)` on staging against whichever endpoint the original perf report came from and checking the plan actually flips after the index lands.

2. **Use Sentry to pick the next target.** Performance issues on the `OccurrenceViewSet` and `SourceImageViewSet` transactions in Sentry should show which queries are dominating wall time post-index. If we're still seeing the same long spans after this index, that's a signal the plan never changed for the hot path and we need a different index (or a different fix entirely).

3. **Operational note for prod rollout.** If the `CREATE INDEX CONCURRENTLY` fails partway, Postgres leaves an INVALID index named `detection_occurrence_ts_desc` and the migration can't be retried until it's dropped (`DROP INDEX CONCURRENTLY detection_occurrence_ts_desc;`). Worth a line in the deploy notes.

Not blocking — just want to make sure the merge buys us what the report claims before we move on.

(For context on #1271: that's a separate N+1 angle on `OccurrenceViewSet.list` — different fix, different code path. Mentioning it as another lever for the "slow queries" theme, not as a dependency of this PR.)

mihow:
**Claude says:** Recommending we close this one, based on a measurement pass on a production-snapshot dataset (part of the #928 slow-query work). Summary of what we found:

**The index doesn't help the query it's aimed at, at our data distribution.** On the snapshot, detections-per-occurrence is averaging ~1.0 (p95 = 1, max 37), so there is essentially nothing to sort per occurrence. On the expensive occurrence aggregate the planner *ignored* the index (626 ms → 643 ms, within noise). Where it does help — first/last-appearance subqueries — the win is ~13 ms → ~2 ms over a couple hundred occurrences, i.e. single-digit milliseconds in absolute terms.

**It adds write cost on the hottest insert path.** `main_detection` is where ML ingestion writes thousands of rows per job; a sixth index there is btree maintenance on every one of those inserts, for a read win the data distribution makes trivial.

**Heads-up on the description:** the PR/commit body cites a `/captures/` 1470 ms → 251 ms improvement, but we couldn't map that to an actual query in the code path — it looks like it may have come from a different change. Worth disregarding so it doesn't muddy the epic's evidence base.

Net: at current scale this is redundant. If we later get projects where occurrences genuinely accumulate many detections each, the first/last-appearance case could justify it — but the data isn't there today.

For context: the occurrence-list slowness that *is* real turned out to be a missing `(project, -determination_score)` index for the default list sort, which is now up as #1351 (proven ~150× on the page query and stable under concurrency). The branch here has been rebased onto current `main`, so it's trivially re-openable if we want it later.

PR #1258, opened 2026-04-20
fix(jobs): throttle and defer pipeline heartbeat update

mihow:
## Problem

Under concurrent async_api jobs, N gunicorn workers call `_mark_pipeline_pull_services_seen` simultaneously from both the `/tasks` (GET) and `/result` (POST) endpoints. The function issued an unguarded `QuerySet.update()` that hit every async `ProcessingService` row matching the `(pipeline, project)` pair — so every request wrote the same shared rows with no throttle, no `F()` expression, and no coordination.

py-spy dumps taken 26s and 57s before a gunicorn SIGSEGV in a deployment under concurrent load showed two threads blocked in this exact `.update()` → psycopg `wait()`. The function runs inside cachalot's monkey-patch, which layers cache invalidation on top of the concurrent UPDATE. Whether the lock contention is a direct contributor to the C-level crash is still being investigated separately; this PR is scoped to relieve the concurrency hotspot regardless of that outcome.

The heartbeat is a visibility feature — `check_processing_services_online` reads `last_seen` against a 60s window — not a correctness-critical path. Writing on every request is unnecessary.

## Goal

Get this DB write off the gunicorn request path entirely, and stop firing it on every request when once per ~30s is enough to keep the 60s offline check honest.

## Changes

**1. Fire-and-forget via Celery.** The view now dispatches `update_pipeline_pull_services_seen.delay(job.pk)` (new task in `ami/jobs/tasks.py`) instead of running the UPDATE inline. The HTTP request returns without waiting on the DB write, so the gunicorn path — where the SIGSEGV was observed — no longer touches this UPDATE at all. The task itself is minimal: one `QuerySet.update()`.

**2. Throttle at the view, before dispatch.** Before `.delay()`, the view takes a Redis cache lock via `cache.add(key, timeout=HEARTBEAT_THROTTLE_SECONDS)` keyed on `(pipeline_id, project_id)`. `cache.add` is cross-process atomic on Redis, so concurrent gunicorn workers see exactly one winner per 30s window; the rest short-circuit before hitting the broker. This is what collapses the write rate — the task itself has no throttle because it doesn't need one.

**3. Broker tolerance.** If `.delay()` raises (broker down, connection reset, OSError), the view logs and continues. The heartbeat is non-critical; a broker blip should not fail the request.

**4. Scope note.** The existing filter — `pipeline.processing_services.async_services().filter(projects=job.project_id)` — is the tightest scope available until per-service application-token auth lands. Once #1117 is merged, the cache key and the UPDATE target should narrow to the individual calling service; the docstrings call this out and a tracking comment has been posted on #1117.

- **Whether the SIGSEGV recurs after deploy.** The `psycopg wait()` → C-level crash link is inferred from py-spy stacks, not a controlled experiment. Moving the write off the request path should remove the specific contention signal py-spy saw, but the crash could have co-factors. Next step if it recurs: gdb + core dump on the crashing container.
- **That `last_seen` stays fresh enough under the 30s gate.** The offline check uses a 60s window, so the margin is 2×. Should be fine, but flagging as a theoretical edge case if Celery queues back up.
- **Broker load.** The gate collapses repeated polls for the same `(pipeline, project)` to one `.delay()` per 30s, so task volume should scale roughly with `active_jobs / 30s` rather than request rate.

## Tests

4 tests in `TestPipelineHeartbeatTask`:

- `test_tasks_endpoint_dispatches_heartbeat_task` — `.delay(job.pk)` fires from `/tasks`
- `test_result_endpoint_dispatches_heartbeat_task` — `.delay(job.pk)` fires from `/result`
- `test_tasks_endpoint_tolerates_heartbeat_dispatch_failure` — `/tasks` still returns 200 when the broker raises
- `test_view_gate_suppresses_redundant_dispatches` — 5 rapid calls collapse to 1 `.delay()`

Pre-existing failures in `TestJobView` (duplicate-key) are unrelated and present on `main`.

## Relation to upcoming work

Once application-token auth (#1117) lands, this can be scoped from `(pipeline, project)` to the individual calling service — cutting the UPDATE target from N services to 1 and letting each service have its own independent 30s gate.

## Summary by CodeRabbit

* **Performance Improvements**
  * Implemented intelligent throttling for service heartbeat tracking, reducing database operations while maintaining accurate status monitoring.
  * Service status updates now execute asynchronously for improved system responsiveness and resource efficiency.

* **Testing**
  * Expanded test coverage for heartbeat tracking and service monitoring across job operations.

### DIFF
--- a/ami/jobs/tests/test_jobs.py
+++ b/ami/jobs/tests/test_jobs.py
@@ -11,6 +11,7 @@ from ami.base.serializers import reverse_with_params
 from ami.jobs.models import Job, JobDispatchMode, JobProgress, JobState, MLJob, SourceImageCollectionPopulateJob
 from ami.main.models import Project, SourceImage, SourceImageCollection
 from ami.ml.models import Pipeline
+from ami.ml.models.processing_service import ProcessingService
 from ami.ml.orchestration.jobs import queue_images_to_nats
 from ami.users.models import User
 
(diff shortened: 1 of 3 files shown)


ami/jobs/tests/test_jobs.py
```diff
@@ -1016,3 +1017,214 @@ def test_tasks_endpoint_rejects_non_async_jobs(self):
         resp = self.client.post(tasks_url, {"batch_size": 1}, format="json")
         self.assertEqual(resp.status_code, 400)
         self.assertIn("async_api", resp.json()[0].lower())
+
+
+class TestPipelineHeartbeatTask(APITestCase):
```
mihow, line 1022: This is a of tests for a "nice to have" feature that is slowing down our required core features. What is necessary?

Issue #1256, opened 2026-04-20
feat(job): refactor job logging so it isn't a bottleneck


mihow:
## Observation

During a moderately-sized async ML job (~742 images), the Celery worker running `process_nats_pipeline_result` begins producing a recurring pair of errors once multiple NATS results are in flight:

```
ERROR Failed to save logs for job #N: SoftTimeLimitExceeded()
ERROR Failed to save logs for job #N: sending query failed: another command is already in progress
ERROR Error processing pipeline result for job N: sending query failed: another command is already in progress. NATS will re-deliver the task message.
```

The same 5-minute window also shows Django containers logging `terminating connection due to idle-in-transaction timeout`. Errors are spread across all `ForkPoolWorker-*` processes (not clustered on one) and are only ever attached to the single actively-processing job row. A live `pg_stat_activity` snapshot during the error window showed no lock contention and plenty of connection headroom (~90 of 500) — i.e. the DB is not saturated.

This appears (tentatively) to be a logical within-connection collision, not a capacity problem.

## Current implementation

`ami/jobs/models.py:335–361` — `JobLogHandler.emit`:

```python
def emit(self, record: logging.LogRecord):
    logger.log(record.levelno, self.format(record))
    try:
        self.job.refresh_from_db(fields=["logs"])          # SELECT
        # ...in-memory insert at position 0, dedupe check, truncate to 1000...
        self.job.save(update_fields=["logs"], update_progress=False)  # UPDATE
    except Exception as e:
        logger.error(f"Failed to save logs for job #{self.job.pk}: {e}")
```

The `refresh_from_db` exists for a real reason (see PR #1162): each concurrent worker holds its own stale in-memory `JobLogs` pydantic object, and without the refresh the last `save()` wins and earlier entries are silently dropped. So the current design is trading concurrency-safety of **log entries** for concurrency-risk at the **connection** level.

## Likely failure path (hypothesis)

`_update_job_progress` (same file, ~L179–209) wraps a `transaction.atomic()` + `Job.objects.select_for_update().get(pk=job_id)` and then calls `job.logger.info(...)` **inside that block**. `JobLogHandler.emit` then issues:

1. `SELECT` (refresh_from_db) on the same connection that already holds a `SELECT … FOR UPDATE` cursor on the same row
2. An `UPDATE` on the same row for `logs` only

[...]

PR #1261, opened 2026-04-21
fix(jobs): fixes for concurrent ML processing jobs

mihow:
## Summary

Tuning PR for `async_api` row-lock and keep-up contention surfaced under
concurrent ML jobs (issue #1256). Scope is what we can safely change while
PR #1259 (the proper `JobLog` append-only child table) and #1256 option D
(atomic `jsonb_set` progress update) land separately.

1. **Drop `select_for_update()` in `_update_job_progress`** — the worker-side
   serialization point. Keeps `transaction.atomic()` and the `max()` guard
   that prevents progress regression.
2. **Add `JOB_LOG_PERSIST_ENABLED` feature flag** (default `True`) — gates
   the per-record UPDATE in `JobLogHandler.emit`. Setting to `False` is a
   time-bounded escape hatch until PR #1259's append-only `JobLog` child
   table is in place.
3. **Narrow `_update_job_progress`'s save to `update_fields`** — limits the
   write to `progress`, `status`, `finished_at`, `updated_at` so a concurrent
   worker's mutation of `logs` / `progress.errors` isn't clobbered by a
   stale read-modify-write. Addresses PR review feedback.
4. **Wire `NATS_TASK_TTR` env var, default 300s** — the ack-wait timeout in
   `nats_queue.py` was reading `settings.NATS_TASK_TTR` via `getattr` with a
   `30` fallback, but the setting was never declared. Effective ack-wait was
   pinned at 30s everywhere, which is too tight for 24 MB images on GPU
   cold-start. Initially raised to 60s, then to 300s (commit `259c6592`) to
   match what staging/demo are actually running — 60s was still tight for
   cold-start GPU pipelines on large images. `max_deliver` (prod=5, demo=2)
   bounds how long a genuinely crashed worker can hold tasks invisibly.
   See "NATS redelivery amplifies Django contention" below.
5. **Docs + script**: `docs/claude/debugging/row-lock-contention-reproduction.md`
   and `scripts/load_test_result_endpoint.py` for local reproduction.

6. **Fair worker polling via `updated_at` — then swapped for random shuffle**
   (commits `014b8c0c`, `7445f6ea`). Worker polling through
   `?ids_only=1&incomplete_only=1` was returning jobs in insertion order, so
   concurrent pollers deterministically converged on the same head-of-queue
   job and later jobs starved. Initial fix was `ORDER BY updated_at, pk` —
   leveraging the fact that `updated_at` freshens on every progress/log
   write, so active jobs drop to the bottom and starved jobs bubble to the
   top. That has a degenerate case at startup (freshly queued jobs share
   near-identical timestamps, tie-broken by `pk`, so simultaneous polls
   still pick the same job). Replaced with `ORDER BY RANDOM()` which gives
   probabilistic disjoint assignment across concurrent pollers without
   writing a poll-stamp column. Fairness is probabilistic rather than
   strictly oldest-first; strict fairness is a follow-up (dedicated
   `last_polled_at` column with `UPDATE ... RETURNING` semantics).

7. **Default `?ids_only=1` to `limit=1`** (commit `7b7dfa5d`). Worker
   polling used to get a paginated batch of 10 jobs and drain them
   serially in `for job_id in jobs: _process_job(job_id)` (see
   `ami-data-companion/trapdata/antenna/worker.py:106`). The cached list
   caused later jobs in the batch to starve until the earlier jobs
   finished. Changing the ids_only response to length-1 forces a re-poll
   per job, which re-samples the random order every iteration. Callers
   that genuinely want a batch can still pass `?limit=N` explicitly —
   back-compat preserved for any non-polling `ids_only` consumers.
   Implemented by overriding `JobViewSet.paginator` to return a
   `LimitOffsetPagination` with `default_limit=1` when `ids_only=1` is
   set and no explicit `limit`/`page_size` is in the query string.
   Combined with the random ordering above, each worker poll becomes an
   independent "pick any unfinished job" draw.

8. **Heartbeat on idle `?ids_only=1` polls** (commits `44d4c322`, `695f7616`).
   `_mark_pipeline_pull_services_seen` only fires from `/tasks/` and
   `/result/` — i.e., from workers with active work — so a worker sitting
   on `GET /jobs/?ids_only=1` between jobs had no heartbeat path and would
   age past `PROCESSING_SERVICE_LAST_SEEN_MAX` (60s), flipping its UI
   status to offline despite still polling.

   The initial commit (`44d4c32`) added a project-scoped heartbeat
   (`_mark_async_services_seen_for_project`). On demo that heartbeat never
   fired: the real ADC worker request is
   `GET /jobs/?pipeline__slug__in=<slugs>&ids_only=1&...` with **no**
   `project_id` in the query string (one worker serves pipelines across
   multiple projects and has no single project to nominate), so
   `get_active_project()` returned `None` and the dispatch silently
   short-circuited.

   Follow-up commit `695f7616` adds a sibling path —
   `_mark_async_services_seen_for_pipelines` +
   `update_async_services_seen_for_pipelines` — that derives scope from
   the `pipeline__slug__in` values in the query and bumps
   `last_seen` / `last_seen_live` on every async `ProcessingService`
   whose `pipelines` match. The `list()` heartbeat branch now prefers
   the pipeline-slug path when those slugs are present and falls back to
   project scope for callers that do send `?project_id=`. Same Redis gate
   (`HEARTBEAT_THROTTLE_SECONDS = 30s`), keyed on the sorted slug set
   (`heartbeat:list:pipelines:<sorted-slugs>`) so concurrent pollers for
   the same pipelines share a single dispatch per window.

   Verified on demo post-deploy: `last_seen_live` flipped from `false` →
   `true` on all three async worker entries (gpu-02, gpu-03, gpu-04)
   within one heartbeat window. Narrowing the scope via
   `AMI_ANTENNA_SERVICE_NAME` (header-based, precise per-service
   identity) is tracked as #1264.

9. **Disable cachalot around `process_nats_pipeline_result`** (commit `826e2bb2`).
   The pipeline-result Celery task runs inside a hot UPDATE path on
   `jobs_job` and triggers cachalot's post-write invalidation across a
   large set of tables on every call. Wrapping the task body in
   `cachalot_disabled()` skips those invalidations for this specific code
   path — the task doesn't query the cached tables anyway, so there's
   nothing to invalidate. Cuts per-task overhead in concurrent-result
   bursts. No behavior change to callers.

**Reviewer takeaway addressed in `9cee66d8`**: the `?ids_only=1` branch of
`list()` (random ordering, `limit=1` paginator override, heartbeat dispatch)
has quietly turned this endpoint into a claim-next-job call. Correct shape
is a dedicated `GET /jobs/next/` action, tracked as #1265. The three
overrides now carry `⚠️ TEMPORARY HACK — remove by 2026-04-24` markers in
code with back-references to #1265, so the expiration stays visible until
`/next` ships and ADC migrates.

Related follow-up issue on the worker side:
[ami-data-companion#143](https://github.com/RolnickLab/ami-data-companion/issues/143)
— once the `/jobs` endpoint is out of the way, starvation moves one
layer down to the NATS-task consumption pattern: a worker that has
reserved tasks from job A via `max_ack_pending` then serializes job B's
batch leaves A's reservations idle until ack_wait expires and NATS
redelivers. Orthogonal to this PR.

## Context

Under concurrent `async_api` result processing, three paths all mutate the
same `jobs_job` row:

1. ML celery workers — `select_for_update` in `_update_job_progress`.
   **(addressed by commit 1 here)**
2. `JobLogHandler.emit` — UPDATE on `jobs_job.logs` per log line. The dominant
   bottleneck after commit 1 lands, because the view's per-iter
   `job.logger.info("Queued pipeline result: …")` fires once per result in a
   batched POST body, stacking N UPDATEs in one `ATOMIC_REQUESTS` tx.
   **(escape hatch via flag in commit 2; proper fix in PR #1259)**
3. `ATOMIC_REQUESTS` holding the `/result` and `/tasks` view tx across broker
   RPCs. **(out of scope — future PR with `@transaction.non_atomic_requests`)**

Commit 1 alone shifts the bottleneck from #1 to #2 rather than eliminating
it (observed: `blocker_chain` grows from ~10 deep to ~56, queue still backs
up). The flag in commit 2 provides a way to actually get the workshop
through the weekend while PR #1259 goes through review.

### NATS redelivery amplifies Django contention

With `ack_wait=30s` pinned, any task that took longer than 30s to finish
(S3 download + GPU cold-start + batch inference + synchronous POST) would
be redelivered by NATS before the original worker got a chance to ack.
Observed in the most recent 6-job load test: one 100-image job had **100
of 100 tasks redelivered** — every task ran twice end-to-end.

The feedback loop that matters for this PR: each redelivered task produces
a **second** POST to `/api/v2/jobs/<id>/result/`, which enqueues a second
`process_nats_pipeline_result` celery task, which contends for the same
`jobs_job` row as the first. So a short ack-wait directly multiplies the
very row-lock load that commits 1-3 are trying to relieve — turning a
throughput shortfall on the ML side into doubled Django-side contention.

Wiring `NATS_TASK_TTR` and raising the default (60s → 300s across the
branch, final value landed in `259c6592`) cuts the redelivery rate;
deployments seeing cold-start cost on 24 MB images can still bump further
via env var. Demo runs the code default (300s).

### Evidence

**py-spy before commit 1**: 7/8 `ForkPoolWorker` threads stuck at
```
wait (psycopg/connection.py:957)
...
_update_job_progress (ami/jobs/tasks.py:493)   ← select_for_update
process_nats_pipeline_result (ami/jobs/tasks.py:224)
```

**py-spy after commit 1, flag still True**: same threads now stuck at
```
save (ami/jobs/models.py:1088)
emit (ami/jobs/models.py:359)
_update_job_progress (ami/jobs/tasks.py:570)   and from
result (ami/jobs/views.py:392)
tasks (ami/jobs/views.py:330)
```

**Staging-equivalent deployment, 5 concurrent `async_api` jobs** (10/100/500
images, same pipeline):

| signal | baseline | commit 1 only | commit 1 + flag=False |
|---|---|---|---|
| `pg_stat_activity` blocker-chain depth | ~10 | 56 | 0 |
| `idle_in_tx` holding `UPDATE jobs_job.logs` for 60s+ | yes | yes | no |
| `ml_results` queue under load | climbs unboundedly | climbs to 200+, does not drain | peaks ~1150, drains to 0 |
| 5-job end state at 500s | stuck / silent FAILURE | 4/5 still at 0% after 210s | 5/5 SUCCESS |

**Local reproduction** (dev compose, `WEB_CONCURRENCY=1`, 8 ml fork workers,
batched POST test at 50 results × 10 concurrent):

| flag | `blocker_chain` | POST success (120s timeout) | p95 latency |
|---|---|---|---|
| `true` (default) | 37 | 0/10 | 120s+ |
| `false` | 1 (transient) | 20/20 | ~5s |

**6-job concurrent retest on demo** (commits 1-3, flag=False, NATS_TASK_TTR=30
i.e. unwired): all 6 jobs eventually SUCCESS, confirming commits 1-3 are
correctness-safe under concurrent load. First job solo ran at 1.4 img/s
(356s for 500); subsequent concurrent jobs ran 2.3-11× slower with high
redelivery (j172 = 100/100 tasks redelivered, j171 = 37/100, j173 = 93/500).
Motivates commit 4.

Reproduction steps, prereqs, the Postgres query, and the load script are in
`docs/claude/debugging/row-lock-contention-reproduction.md`. A
single-result-per-POST curl loop does not reproduce the pathology;
batching results per POST is load-bearing because the per-iter
`job.logger.info` inside the view is where the UPDATE stacking happens.

## Trade-offs

**Commit 1 — count drift under race.** The `max()` guard still prevents
progress-percentage regression between concurrent workers. Accumulated
counts (detections, classifications, captures) can drift by one batch if
two workers race the read-modify-write cycle. Cosmetic: underlying
`Detection` and `Classification` rows are written authoritatively by
`save_results()`, which runs *before* `_update_job_progress`. A fully
atomic version with `jsonb_set` and server-side arithmetic is possible as a
follow-up (#1256 option D).

**Commit 2 — UI log feed goes blank while flag is off.** Container stdout
still captures every line via the module-level app logger, so ops
observability is unchanged. Only the per-job UI log view loses new
entries for the duration the flag is disabled. Default stays `True` so
existing deployments aren't affected.

**Commit 4 — longer ack-wait delays crash detection.** If a worker truly
dies mid-task, the `NATS_TASK_TTR` seconds are wasted before redelivery.
At the 300s default this is still acceptable for GPU cold-start on 24 MB
images — the redelivery-amplification cost dominates the crash-detection
latency under the workloads we see. Deployments with smaller payloads or
stricter freshness SLAs can tune down via env. `max_deliver` (prod=5,
demo=2) is the cap that prevents infinite retries.

## Rollout plan

1. Land this PR on `main`; keep the flag default `True` and `NATS_TASK_TTR`
   default 300. Matches what demo/staging are already running.
2. Deployments hitting the contention pattern set
   `JOB_LOG_PERSIST_ENABLED=false` on their Django container env and
   restart. Revert by unsetting or switching to `true`.
3. Deployments with extreme cold-start cost can bump `NATS_TASK_TTR`
   further via env; the 300s default is sized for the 24 MB image +
   H100 cold-start case we've been testing.
4. When PR #1259 lands, the default-`True` log path migrates to the new
   `JobLog` table write; the flag can be removed in a cleanup PR.
5. When #1256 option D (atomic `jsonb_set` progress update) lands, commit 1's
   count-drift trade-off goes away.

## Tested / verified

Unit tests (local, `docker compose -f docker-compose.ci.yml run --rm django
python manage.py test ami.jobs.tests.test_jobs`):

- [x] 49/49 tests pass, including the new `TestListEndpointHeartbeat`
  class covering both the project-scoped and pipeline-slug heartbeat
  paths (7 cases). Also green on CI for the latest SHA.

Local reproduction (dev compose, `WEB_CONCURRENCY=1`, 8 ml fork workers,
`scripts/load_test_result_endpoint.py` at 50 results × 10 concurrent):

- [x] Commit 1 alone shifts contention from `select_for_update` to
  `JobLogHandler.emit`; `blocker_chain` grows from ~10 to ~37. 0/10
  batched POSTs succeed in 120s.
- [x] Commit 1 + `JOB_LOG_PERSIST_ENABLED=false`: `blocker_chain` ≤ 1
  transient, 20/20 batched POSTs succeed, p95 ~5s. Numbers match the
  evidence table above.

Staging-equivalent deployment (Beluga), 5 concurrent `async_api` jobs
(10/100/500 images, same pipeline):

- [x] Baseline (main): 4/5 jobs still at 0% after 210s, `idle_in_tx`
  holding `UPDATE jobs_job.logs` for 60s+, `ml_results` queue climbs
  unboundedly.
- [x] This branch with flag=False: 5/5 SUCCESS in ~500s, queue peaks
  ~1150 then drains to 0, no `idle_in_tx` holders. Confirms commits 1-3
  are correctness-safe under concurrent load.

Demo env (`demo.antenna.insectai.org`):

- [x] 6-job concurrent retest on demo pre-`NATS_TASK_TTR` (j171-j176,
  mixed sizes 10/10/100/100/500/500): all 6 jobs eventually SUCCESS
  despite high redelivery (j172 = 100/100 tasks redelivered). Motivated
  commit 4.
- [x] Branch deployed to demo, all six containers recreated on SHA
  `9cee66d8`. Verified via `curl demo.antenna.insectai.org/api/v2/ml/processing_services/?project_id=27`
  that `last_seen_live` flipped from `false` (stuck for hours) → `true`
  on all three async worker entries (gpu-02, gpu-03, gpu-04) within one
  heartbeat window post-deploy — confirms the pipeline-slug heartbeat
  path fires for the real ADC worker request shape.

antenna-dev-serbia + antenna-dev-arctia (full-stack dev benches running
the branch end-to-end):

- [x] Concurrent e2e via `test_ml_job_e2e` SUCCESS on both boxes on the
  merged-with-#1259 integration branch (CPU-only serbia and H100-24GB
  arctia, respectively).

Not yet verified:

- [x] Overnight soak under real ADC load on demo.
- [ ] Production rollout with `JOB_LOG_PERSIST_ENABLED` flipped on (prod
  will default `True`; the flag-off path is covered by the staging
  5-job test above, not by a prod run).

## Related

- Issue #1256 — original contention analysis, full path breakdown; option D is the follow-up atomic UPDATE that retires commit 1's drift caveat
- PR #1259 (draft) — append-only `JobLog` child table (permanent fix that deprecates the flag)
- PR #1258 (merged) — heartbeat throttle (predecessor that surfaced this underlying contention)

## Summary by CodeRabbit

* **New Features**
  * Added a configuration flag to toggle per-job log persistence to the database.
  * Added a CLI tool to load-test the job result endpoint.
  * Adjusted job list behavior: ID-only queries get a tighter default pagination.

* **Performance**
  * Reduced database row-lock contention during job progress updates.
  * Prevented unnecessary post-write cache invalidation during pipeline result handling.

* **Documentation**
  * Added a runbook for reproducing and debugging row-lock contention and testing the log-persistence flag.

### DIFF
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -578,7 +578,13 @@ def _update_job_progress(
             job.progress.summary.status = complete_state
             job.finished_at = datetime.datetime.now()  # Use naive datetime in local time
         job.logger.info(f"Updated job {job_id} progress in stage '{stage}' to {progress_percentage*100}%")
-        job.save()
+        # Narrow the write to the fields we actually mutated. Without this, a full
+        # save() would also overwrite `updated_at`, `logs`, and any other field on
+        # the instance fetched at the top of this block — so a concurrent worker's
+        # append to `progress.errors` (via `_reconcile_lost_images`) or log line
+        # (via JobLogHandler) could be clobbered by a stale read-modify-write.
+        # See PR #1261 review feedback.
+        job.save(update_fields=["progress", "status", "finished_at", "updated_at"])
         try:
             _log_job_throughput(job, stage)
         except Exception as e:
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -91,6 +91,11 @@ def update_async_services_seen_for_pipelines(pipeline_slugs: list[str]) -> None:
     pipelines across many projects), so scope the heartbeat by the pipelines it
     asked about. Marks every async ProcessingService linked to any of those
     pipelines as seen.
+
+    TODO: once #1194 (client-ID / application-token auth) lands, scope this
+    update to the specific calling ProcessingService rather than every service
+    matching the slugs. Currently one poller's heartbeat falsely marks its
+    peers live.
     """
     from ami.ml.models import ProcessingService  # avoid circular import
 
@@ -114,10 +119,14 @@ def update_async_services_seen_for_project(project_id: int) -> None:
     """
     Heartbeat for idle worker polls on ``GET /api/v2/jobs/?ids_only=1``.
 
-    Unlike ``update_pipeline_pull_services_seen`` — which is pipeline-scoped and
-    only fires when a worker hits /tasks/ or /result/ for an active job — this
-    marks every async processing service attached to the polling project as
-    seen. The list endpoint has no pipeline context, so scope is the project.
+    Fallback path used only when the request carries ``?project_id=`` without
+    ``pipeline__slug__in`` — the ADC worker does not currently send this shape,
+    so in practice the pipeline-slug task above is the one that fires.
+
+    TODO: once #1194 (client-ID / application-token auth) lands, scope this
+    update to the specific calling ProcessingService rather than every async
+    service attached to the project. Currently one poller's heartbeat falsely
+    marks its peers live.
     """
     from ami.ml.models import ProcessingService  # avoid circular import
 
@@ -629,11 +638,12 @@ def _update_job_progress(
             job.finished_at = datetime.datetime.now()  # Use naive datetime in local time
         job.logger.info(f"Updated job {job_id} progress in stage '{stage}' to {progress_percentage*100}%")
         # Narrow the write to the fields we actually mutated. Without this, a full
-        # save() would also overwrite `updated_at`, `logs`, and any other field on
-        # the instance fetched at the top of this block — so a concurrent worker's
-        # append to `progress.errors` (via `_reconcile_lost_images`) or log line
-        # (via JobLogHandler) could be clobbered by a stale read-modify-write.
-        # See PR #1261 review feedback.
+        # save() would overwrite `logs` and any other field on the instance
+        # fetched at the top of this block — so a concurrent worker's append to
+        # `progress.errors` (via `_reconcile_lost_images`) or log line (via
+        # JobLogHandler) could be clobbered by a stale read-modify-write.
+        # `updated_at` is listed explicitly because Django skips `auto_now` bumps
+        # when `update_fields` is provided. See PR #1261 review feedback.
         job.save(update_fields=["progress", "status", "finished_at", "updated_at"])
         try:
             _log_job_throughput(job, stage)
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -7,6 +7,7 @@ from collections.abc import Callable
 from typing import TYPE_CHECKING
 
 from asgiref.sync import async_to_sync, sync_to_async
+from cachalot.api import cachalot_disabled
 from celery.signals import task_failure, task_postrun, task_prerun
 from django.db import transaction
 from redis.exceptions import RedisError
@@ -158,6 +159,15 @@ def _log_worker_availability(job) -> None:
     soft_time_limit=300,  # 5 minutes
     time_limit=360,  # 6 minutes
 )
+# Disable cachalot cache invalidation for this task. Each call writes
+# Detection/Classification rows and UPDATEs jobs_job; under concurrent
+# async_api load, cachalot's post-write invalidation added ~2.5s/task
+# (measured on demo, issue #1256 Path 4). This is a pure write path —
+# nothing inside benefits from the query cache — so skipping invalidation
+# is strictly a throughput win. Celery task decorator stack order matters:
+# @celery_app.task wraps the cachalot-wrapped function, so Celery sees the
+# cachalot context manager enter/exit on every task execution.
+@cachalot_disabled()
 def process_nats_pipeline_result(self, job_id: int, result_data: dict, reply_subject: str) -> None:
     """
     Process a single pipeline result asynchronously.
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -77,6 +77,28 @@ def update_pipeline_pull_services_seen(job_id: int) -> None:
     )
 
 
+@celery_app.task(
+    soft_time_limit=10,
+    time_limit=15,
+    ignore_result=True,
+)
+def update_async_services_seen_for_project(project_id: int) -> None:
+    """
+    Heartbeat for idle worker polls on ``GET /api/v2/jobs/?ids_only=1``.
+
+    Unlike ``update_pipeline_pull_services_seen`` — which is pipeline-scoped and
+    only fires when a worker hits /tasks/ or /result/ for an active job — this
+    marks every async processing service attached to the polling project as
+    seen. The list endpoint has no pipeline context, so scope is the project.
+    """
+    from ami.ml.models import ProcessingService  # avoid circular import
+
+    ProcessingService.objects.async_services().filter(projects=project_id).update(
+        last_seen=datetime.datetime.now(),
+        last_seen_live=True,
+    )
+
+
 @celery_app.task(bind=True, soft_time_limit=default_soft_time_limit, time_limit=default_time_limit)
 def run_job(self, job_id: int) -> None:
     from ami.jobs.models import Job
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -509,7 +509,7 @@ def _update_job_progress(
     # (detections/classifications/captures) can drift by one batch under race —
     # cosmetic only, since the underlying `Detection`/`Classification` rows are
     # written authoritatively by `save_results` before this function runs.
-    # See docs/claude/planning/jobs-row-lock-remediation.md and issue #1256.
+    # See issue #1256 and PR #1261.
     with transaction.atomic():
         job = Job.objects.get(pk=job_id)
 
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -489,8 +489,19 @@ def _update_job_progress(
 ) -> None:
     from ami.jobs.models import Job, JobState  # avoid circular import
 
+    # NOTE: Previously this used `select_for_update()` inside `transaction.atomic()`
+    # to serialize concurrent progress updates for the same job. Under concurrent
+    # async_api result processing that serialization became a bottleneck: every
+    # ML result task queued a contending exclusive lock on the `jobs_job` row,
+    # stacking behind gunicorn view threads also holding the row under
+    # ATOMIC_REQUESTS. The `max()` guard below still prevents progress regression
+    # between concurrent workers; the trade-off is that accumulated counts
+    # (detections/classifications/captures) can drift by one batch under race —
+    # cosmetic only, since the underlying `Detection`/`Classification` rows are
+    # written authoritatively by `save_results` before this function runs.
+    # See docs/claude/planning/jobs-row-lock-remediation.md and issue #1256.
     with transaction.atomic():
-        job = Job.objects.select_for_update().get(pk=job_id)
+        job = Job.objects.get(pk=job_id)
 
         # For results stage, accumulate detections/classifications/captures counts
         if stage == "results":
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -82,6 +82,34 @@ def update_pipeline_pull_services_seen(job_id: int) -> None:
     time_limit=15,
     ignore_result=True,
 )
+def update_async_services_seen_for_pipelines(pipeline_slugs: list[str]) -> None:
+    """
+    Heartbeat for idle worker polls on
+    ``GET /api/v2/jobs/?pipeline__slug__in=...&ids_only=1``.
+
+    The ADC worker sends pipeline slugs but no project_id (one worker may serve
+    pipelines across many projects), so scope the heartbeat by the pipelines it
+    asked about. Marks every async ProcessingService linked to any of those
+    pipelines as seen.
+    """
+    from ami.ml.models import ProcessingService  # avoid circular import
+
+    if not pipeline_slugs:
+        return
+
+    ProcessingService.objects.async_services().filter(
+        pipelines__slug__in=pipeline_slugs,
+    ).distinct().update(
+        last_seen=datetime.datetime.now(),
+        last_seen_live=True,
+    )
+
+
+@celery_app.task(
+    soft_time_limit=10,
+    time_limit=15,
+    ignore_result=True,
+)
 def update_async_services_seen_for_project(project_id: int) -> None:
     """
     Heartbeat for idle worker polls on ``GET /api/v2/jobs/?ids_only=1``.
--- a/ami/jobs/views.py
+++ b/ami/jobs/views.py
@@ -60,12 +60,16 @@ def _actor_log_context(request) -> tuple[str, str | None]:
 
 def _mark_async_services_seen_for_pipelines(pipeline_slugs: tuple[str, ...]) -> None:
     """
-    Heartbeat for idle worker polls that send ``pipeline__slug__in=...`` but no
-    ``project_id`` — the real ADC worker shape, where one worker may serve
-    pipelines across many projects and has no single project to nominate.
+    Redis-throttled wrapper around the ``update_async_services_seen_for_pipelines``
+    celery task. The wrapper does no DB work itself — it gates dispatch so at
+    most one heartbeat is enqueued per sorted slug set per
+    ``HEARTBEAT_THROTTLE_SECONDS`` window (currently 30s), keeping the HTTP
+    request path cheap under concurrent polling.
 
-    Redis throttle keyed on the sorted slug set so concurrent pollers for the
-    same pipelines share a single dispatch per window.
+    Called from the ``?ids_only=1`` branch of ``JobViewSet.list()`` — the real
+    ADC worker shape, which sends ``pipeline__slug__in=<slugs>`` and no
+    ``project_id`` (one worker may serve pipelines across many projects and
+    has no single project to nominate).
     """
     if not pipeline_slugs:
         return
@@ -80,17 +84,18 @@ def _mark_async_services_seen_for_pipelines(pipeline_slugs: tuple[str, ...]) -> 
 
 def _mark_async_services_seen_for_project(project_id: int) -> None:
     """
-    Heartbeat for idle worker polls on ``GET /api/v2/jobs/?ids_only=1``.
+    Redis-throttled wrapper around ``update_async_services_seen_for_project``.
+    Same shape as ``_mark_async_services_seen_for_pipelines`` above — gates
+    celery dispatch to at most one per-project enqueue per
+    ``HEARTBEAT_THROTTLE_SECONDS`` window — but keyed by project id for
+    callers that send ``?project_id=`` without ``pipeline__slug__in``.
 
-    The pipeline-scoped heartbeat in ``_mark_pipeline_pull_services_seen`` only
-    fires when a worker hits /tasks/ or /result/ on an active job; workers idling
-    on the list endpoint between jobs had no heartbeat path at all, so their
-    ``last_seen`` would age out of ``PROCESSING_SERVICE_LAST_SEEN_MAX`` and the
-    UI would flip them to offline despite being actively online.
-
-    Scope: marks every async service attached to the polling project. The list
-    endpoint has no pipeline context to narrow by. Once application-token auth
-    lands (PR #1117), this should be scoped to the specific calling service.
+    The ADC worker does not currently use this shape, so this is a fallback.
+    Background on why idle-poll heartbeats exist at all: the other heartbeat
+    (``_mark_pipeline_pull_services_seen``) only fires from ``/tasks/`` and
+    ``/result/`` — i.e., from workers with active work — so a worker sitting
+    on ``GET /jobs/?ids_only=1`` between jobs would otherwise age past
+    ``PROCESSING_SERVICE_LAST_SEEN_MAX`` and flip to offline in the UI.
     """
     cache_key = f"heartbeat:list:project:{project_id}"
     if not cache.add(cache_key, 1, timeout=HEARTBEAT_THROTTLE_SECONDS):
@@ -302,6 +307,7 @@ class JobViewSet(DefaultViewSet, ProjectMixin):
             status=JobState.failed_states(),
             updated_at__lt=cutoff_datetime,
         )
+        # ⚠️ TEMPORARY HACK — remove by 2026-04-24.
         # Worker-polling call path (`ids_only=1`): randomize order so concurrent
         # pollers don't all converge on the same head-of-queue job. An
         # `updated_at`-based sort has a degenerate case at startup — freshly
@@ -310,18 +316,32 @@ class JobViewSet(DefaultViewSet, ProjectMixin):
         # ordering gives probabilistic disjoint assignment without writing a
         # poll-stamp column. Combined with `limit=1` below, each poll is an
         # independent "pick any unfinished job" draw.
+        #
+        # The whole `ids_only=1` branch (this ordering override, the paginator
+        # override in `paginator` below, the heartbeat dispatch in `list()`)
+        # exists because the ADC worker currently repurposes this list endpoint
+        # as a claim-next-job call. Correct shape is a dedicated `/next` action
+        # (tracked as #1265). Once `/next` ships
+        # and ADC is migrated, delete this `order_by("?")` override along with
+        # the paginator override and the list() heartbeat branch.
         if self.action == "list" and url_boolean_param(self.request, "ids_only", default=False):
             jobs = jobs.order_by("?")
         return jobs
 
     @property
     def paginator(self):
+        # ⚠️ TEMPORARY HACK — remove by 2026-04-24.
         # Treat `?ids_only=1` as a pop()-style handoff ("what job is next?")
         # rather than a list() dump: default to one job per response unless the
         # caller explicitly asks for a batch via ?limit=N or ?page_size=N.
         # Concurrent pollers drain a cached list serially and starve later jobs;
-        # forcing a re-poll per job lets the `updated_at` fairness sort rotate
+        # forcing a re-poll per job lets the random-shuffle fairness sort rotate
         # work across jobs every iteration. No ADC-side change required.
+        #
+        # This override exists only because `list(ids_only=True)` is being used
+        # as a claim-next-job call. Replace with a dedicated `/next` action
+        # (tracked as #1265); once ADC is migrated,
+        # drop this override so the list endpoint goes back to normal pagination.
         if not hasattr(self, "_paginator"):
             if (
                 self.action == "list"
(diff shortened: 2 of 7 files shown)


ami/jobs/tasks.py
```diff
@@ -158,6 +159,15 @@ def _log_worker_availability(job) -> None:
...
+# Disable cachalot cache invalidation for this task. Each call writes
+# Detection/Classification rows and UPDATEs jobs_job; under concurrent
+# async_api load, cachalot's post-write invalidation added ~2.5s/task
+# (measured on demo, issue #1256 Path 4). This is a pure write path —
+# nothing inside benefits from the query cache — so skipping invalidation
+# is strictly a throughput win. Celery task decorator stack order matters:
+# @celery_app.task wraps the cachalot-wrapped function, so Celery sees the
+# cachalot context manager enter/exit on every task execution.
+@cachalot_disabled()
 def process_nats_pipeline_result(self, job_id: int, result_data: dict, reply_subject: str) -> None:
```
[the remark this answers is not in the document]
mihow: **Claude says:** This claim is empirically incorrect for the pinned version. Python's `@contextmanager`-produced `_GeneratorContextManager` has inherited from `ContextDecorator` since Python 3.2, so any `@contextmanager` function can be used as either a `with` block or a decorator.

Verified directly against the installed `cachalot==X.Y` in this repo's container:

```python
from cachalot.api import cachalot_disabled, LOCAL_STORAGE

def probe():
    return getattr(LOCAL_STORAGE, "cachalot_enabled", "unset")

# Context manager form
print("outside:", probe())
with cachalot_disabled():
    print("inside:", probe())
print("after:", probe())

# Decorator form
@cachalot_disabled()
def dec_fn():
    return probe()

print("outside:", probe())
print("inside dec_fn() call 1:", dec_fn())
print("after call 1:", probe())
print("inside dec_fn() call 2:", dec_fn())
print("after call 2:", probe())
```

Output (both in a local Django shell and in the `celeryworker_ml` container):
```
=== context manager ===
outside: unset
inside: False
after: True

=== decorator ===
outside: True
inside dec_fn() call 1: False
after call 1: True
inside dec_fn() call 2: False
after call 2: True
```

Both forms flip `LOCAL_STORAGE.cachalot_enabled` to `False` during execution and restore on exit, across repeated invocations.

Also validated on the staging-equivalent deployment with a 500-image async_api job: the decorator form moves `ml_results` drain rate from ~5 tasks/sec (baseline, cachalot on) to ~8.5 tasks/sec (cachalot off via decorator). Measured over the final 65s of clean drain (queue fell 556 → 269 → 0). Matches the predicted Path 4 improvement from issue #1256 (~2.5s → ~0.9s per task).

Happy to switch to `with cachalot_disabled():` if preferred for stylistic reasons, but the correctness claim (decorator is a no-op) doesn't hold.

ami/jobs/tasks.py
```diff
@@ -489,8 +489,19 @@ def _update_job_progress(
...
+    # to serialize concurrent progress updates for the same job. Under concurrent
+    # async_api result processing that serialization became a bottleneck: every
+    # ML result task queued a contending exclusive lock on the `jobs_job` row,
+    # stacking behind gunicorn view threads also holding the row under
+    # ATOMIC_REQUESTS. The `max()` guard below still prevents progress regression
+    # between concurrent workers; the trade-off is that accumulated counts
+    # (detections/classifications/captures) can drift by one batch under race —
+    # cosmetic only, since the underlying `Detection`/`Classification` rows are
+    # written authoritatively by `save_results` before this function runs.
+    # See docs/claude/planning/jobs-row-lock-remediation.md and issue #1256.
```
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in 9d15813 — dropped the path to the non-existent planning doc; the inline NOTE now just references issue #1256 and PR #1261. Thanks for the catch.

ami/jobs/tasks.py
```diff
@@ -489,8 +489,19 @@ def _update_job_progress(
...
     with transaction.atomic():
-        job = Job.objects.select_for_update().get(pk=job_id)
+        job = Job.objects.get(pk=job_id)
```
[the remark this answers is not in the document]
mihow: **Claude says:** Accepted as a known trade-off — see the inline NOTE in `_update_job_progress` for the reasoning, and the PR body for the observed before/after. Short version: the `max()` guard below still prevents progress regression on the same stage, and the accumulated counters (detections/classifications/captures) can drift by at most one batch under a RMW race — cosmetic only, since the underlying `Detection`/`Classification` rows are written authoritatively by `save_results` before this function runs. The alternative (optimistic compare-and-swap with retry, or a server-side `jsonb_set` atomic update — issue #1256 option D) is a larger change and is tracked as a follow-up; under current load (5 concurrent async_api jobs validated on demo + Serbia) the lockless path is holding up with `blocker_chain=0`. If concurrency climbs meaningfully past that, we'll revisit. Resolving.

ami/jobs/tasks.py
```diff
@@ -489,8 +499,19 @@ def _update_job_progress(
...
```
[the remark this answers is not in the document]
mihow: **Claude says:** Acknowledged, and agreed the window is small (reconcile only runs for idle jobs past the cutoff). Narrowing `job.save()` to `update_fields=["progress", "status", "finished_at", "updated_at"]` doesn't actually fix the `progress.errors` clobbering concern — `progress` is a single JSONB column, so a narrowed save still writes the whole dict back. The real mitigation is optimistic compare-and-swap or a server-side `jsonb_set` atomic update (issue #1256 option D), which is tracked as a follow-up. Keeping the full save for now since neither shape helps here; noting for the follow-up PR. Resolving.

ami/jobs/tasks.py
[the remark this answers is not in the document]
mihow: **Claude says:** Right that the race is real — with `select_for_update()` gone, two concurrent `_update_job_progress` calls can both read, mutate, and last-writer-win the `progress` JSON. Pushing back on the specific suggestion though: `select_for_update(nowait=True)` / `skip_locked` isn't the right fix for this path.

- `nowait=True` raises `DatabaseError` when the row is locked → needs a retry loop, which re-serializes onto the same hot row and puts us back in the blocker chain we just removed (just failing faster).
- `skip_locked` silently drops the progress update → worse than last-write-wins, because the drop is permanent rather than overwritten on the next call.
- Both still require an exclusive lock on `jobs_job` per task, which is exactly what the async_api result POST path cannot afford at 5+ concurrent workers (measured on demo: blocker_chain of 37 with `select_for_update`, ~0 without).

The right fix for the race is an atomic SQL update that avoids the read-modify-write entirely — `Job.objects.filter(pk=id).update(progress=Func(..., jsonb_set(..., GREATEST(...))))` — which is tracked as option D on issue #1256 and is a separate PR. For this PR the remaining risk is bounded because: (a) the window between `get()` and `save()` is a single UPDATE with no RPCs (microseconds), (b) counters are incremental so any loss is overwritten by the next call a few hundred ms later, (c) `progress.errors` is only appended by `_reconcile_lost_images`, which has a `updated_at < cutoff` guard that disqualifies actively-progressing jobs.

Partial hardening landed in ef78fb50: narrow `job.save()` to `update_fields=["progress","status","finished_at","updated_at"]` so the read-modify-write can't clobber unrelated fields (`logs`, etc.). Leaving this thread open as a pointer to the option-D follow-up.

ami/jobs/tasks.py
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in 9d15813 — swapped the stale planning-doc reference for a pointer to issue #1256 and PR #1261.

scripts/load_test_result_endpoint.py
```diff
@@ -0,0 +1,91 @@
...
+produce (``AMI_LOCALIZATION_BATCH_SIZE=4``, ``AMI_CLASSIFICATION_BATCH_SIZE=150``).
+
+A single-result-per-POST loop does NOT reproduce the contention. Batching
+is load-bearing.
+
+Usage:
+
+    python scripts/load_test_result_endpoint.py <job_id> <token> \\
+        [--batch 50] [--concurrency 10] [--rounds 3] \\
+        [--host http://localhost:8000]
```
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in 9d15813 — token now defaults to `$ANTENNA_TOKEN` (positional removed), `--batch`/`--concurrency`/`--rounds` validated as int > 0 via a custom `_positive_int` type, `--host` validated to http(s) scheme with non-empty netloc via `_http_url`, and per-request errors now carry the exception type + reason (`HTTPError: ...`, `URLError: ...`, generic fallback with class name) instead of collapsing to -1. Up to 5 errors printed to stderr per round with an overflow count.

ami/jobs/tasks.py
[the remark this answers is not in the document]
mihow: **Claude says:** Good catch — accepted. Narrowed the write in ef78fb50 to `job.save(update_fields=["progress","status","finished_at","updated_at"])` (included `updated_at` so `auto_now` still fires).

Didn't go the full "re-read `progress.errors` inside the atomic block" route because the only appender (`_reconcile_lost_images`) has a `updated_at < cutoff` guard that disqualifies actively-progressing jobs, so the two write paths don't overlap in practice. The race on the shared JSON column remains (tracked as issue #1256 option D — atomic `jsonb_set` UPDATE); this just stops the save from clobbering unrelated fields like `logs`.

scripts/load_test_result_endpoint.py
```diff
@@ -0,0 +1,91 @@
...
+        method="POST",
+    )
+    t0 = time.time()
+    try:
+        with urllib.request.urlopen(req, timeout=120) as resp:
+            return (idx, resp.status, time.time() - t0)
+    except urllib.error.HTTPError as e:
+        return (idx, e.code, time.time() - t0)
+    except Exception:
+        return (idx, -1, time.time() - t0)
```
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in 9d15813 — token now defaults to `$ANTENNA_TOKEN` (positional removed), `--batch`/`--concurrency`/`--rounds` validated as int > 0 via a custom `_positive_int` type, `--host` validated to http(s) scheme with non-empty netloc via `_http_url`, and per-request errors now carry the exception type + reason (`HTTPError: ...`, `URLError: ...`, generic fallback with class name) instead of collapsing to -1. Up to 5 errors printed to stderr per round with an overflow count.

ami/jobs/tasks.py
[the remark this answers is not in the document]
mihow: **Claude says (update):** @mihow landed ef78fb5 after my earlier reply, narrowing the save to `update_fields=["progress", "status", "finished_at", "updated_at"]`. That does limit blast radius to fields other than `progress` (e.g. the `logs` JSONB before PR #1259 lands, or any future scalar field on Job). The progress-subfield clobbering concern is unchanged and still tracked as a follow-up (issue #1256 option D). Thanks for the nudge.

ami/jobs/views.py
```diff
@@ -261,6 +344,20 @@ def get_queryset(self) -> QuerySet:
         ]
     )
     def list(self, request, *args, **kwargs):
```
mihow, line 366: This list function became very specific if you pass ids_only=True. What we really want is a `/next` action which pops / returns a random job's list of tasks. And then let the list view have normal list behavior. But the running ADCs will need to make that change as well. Let's make both work. Make a new action or function called next() that list() calls if ids_only is true. Or a shared function called _get_next_ml_job() and mark / note that the list(ids_only=True) path is depreciated.

config/settings/base.py
```diff
@@ -568,3 +575,12 @@ def _celery_result_backend_url(redis_url):
 # Default taxa filters
 DEFAULT_INCLUDE_TAXA = env.list("DEFAULT_INCLUDE_TAXA", default=[])  # type: ignore[no-untyped-call]
 DEFAULT_EXCLUDE_TAXA = env.list("DEFAULT_EXCLUDE_TAXA", default=[])  # type: ignore[no-untyped-call]
+
+# When True, ``JobLogHandler.emit`` persists each log line to ``jobs_job.logs``
```
mihow, line 582: Add a plain language purpose of this flag: e.g. Disable jobs logs in the database and UI

ami/jobs/tasks.py
```diff
@@ -76,6 +77,56 @@ def update_pipeline_pull_services_seen(job_id: int) -> None:
...
+    soft_time_limit=10,
+    time_limit=15,
+    ignore_result=True,
+)
+def update_async_services_seen_for_pipelines(pipeline_slugs: list[str]) -> None:
+    """
+    Heartbeat for idle worker polls on
+    ``GET /api/v2/jobs/?pipeline__slug__in=...&ids_only=1``.
+
+    The ADC worker sends pipeline slugs but no project_id (one worker may serve
```
mihow, line 90: Add TODO scope this last_seen update to the calling processing service, once we merge #1194 and can identify it with a client ID

ami/jobs/tasks.py
```diff
@@ -76,6 +77,56 @@ def update_pipeline_pull_services_seen(job_id: int) -> None:
...
+@celery_app.task(
+    soft_time_limit=10,
+    time_limit=15,
+    ignore_result=True,
+)
+def update_async_services_seen_for_project(project_id: int) -> None:
+    """
+    Heartbeat for idle worker polls on ``GET /api/v2/jobs/?ids_only=1``.
+
+    Unlike ``update_pipeline_pull_services_seen`` — which is pipeline-scoped and
```
mihow: Add TODO scope this last_seen update to the calling processing service, once we merge #1194 and can identify it with a client ID

ami/jobs/tasks.py
```diff
@@ -557,7 +628,13 @@ def _update_job_progress(
             job.progress.summary.status = complete_state
             job.finished_at = datetime.datetime.now()  # Use naive datetime in local time
         job.logger.info(f"Updated job {job_id} progress in stage '{stage}' to {progress_percentage*100}%")
-        job.save()
+        # Narrow the write to the fields we actually mutated. Without this, a full
+        # save() would also overwrite `updated_at`, `logs`, and any other field on
```
mihow: we _are_ updating `updated_at` here. update the comment

ami/jobs/views.py
```diff
@@ -51,6 +58,49 @@ def _actor_log_context(request) -> tuple[str, str | None]:
     return user_desc, token_fingerprint
 
 
+def _mark_async_services_seen_for_pipelines(pipeline_slugs: tuple[str, ...]) -> None:
+    """
+    Heartbeat for idle worker polls that send ``pipeline__slug__in=...`` but no
```
mihow: Clarify these 2/4 nearly identically functions in views and tasks (_mark* and update*). These two are just wrapping the core update / actual work in a redis cache so that new updates are queued at most every 30 seconds.

ami/jobs/views.py
mihow: **Claude says:** Agreed — the `list(ids_only=1)` overloading became a claim-next-job call by accretion. Filed follow-up issue #1265 to introduce a dedicated `GET /jobs/next/` action and deprecate this path. In the meantime (9cee66d8), the three overrides (random ordering, paginator default_limit=1, heartbeat dispatch on list()) carry a prominent `⚠️ TEMPORARY HACK — remove by 2026-04-24` marker with a back-reference to #1265, so the expiration stays visible until `/next` lands and ADC is migrated.

config/settings/base.py
mihow: **Claude says:** Added a one-line plain-language lead: `# Purpose: master switch for per-job logs in the database and UI. Set to False to disable them as a contention escape hatch.` (9cee66d8)

ami/jobs/tasks.py
mihow: **Claude says:** Added the TODO in 9cee66d8 on `update_async_services_seen_for_pipelines` (new) and `update_async_services_seen_for_project` (fallback), referencing #1194. Kept per-service scoping out of this PR since it depends on client-ID auth.

ami/jobs/tasks.py
mihow: **Claude says:** Same fix — TODO added on both task bodies in 9cee66d8 referencing #1194.

ami/jobs/tasks.py
mihow: **Claude says:** Good catch — the comment was wrong. Rewrote in 9cee66d8: we include `updated_at` in `update_fields` intentionally because Django skips `auto_now` bumps when `update_fields` is provided. The fields we're avoiding overwriting are `logs` and `progress.errors` on a concurrent worker's read-modify-write.

ami/jobs/views.py
mihow: **Claude says:** Rewrote both `_mark_async_services_seen_*` docstrings in 9cee66d8 to state the contract explicitly: "Redis-throttled wrapper around the matching celery task — no DB work in the wrapper, only dispatch gating." This leaves the two layers (throttle at the HTTP path, update in the worker) clearly separated in the documentation.

Issue #1265, opened 2026-04-23
Add GET /jobs/next action; deprecate list(ids_only=1) claim-semantics


mihow:
## Context

PR #1261 bundled a set of `async_api` concurrency fixes. One of those — the
worker fairness rework — reshaped `GET /jobs/?ids_only=1` into a
claim-next-job endpoint: random ordering, `limit=1` default via a paginator
override, and a heartbeat side-effect on every call. Reviewers (see
PR #1261) flagged that the `list()` endpoint has quietly become a pop()
queue with three hacks bolted on.

The right shape is a dedicated action.

## Goal

Introduce a dedicated `GET /api/v2/jobs/next/` action that:

1. Returns the next claim for a worker: a single job's id (or the
   `{id, pipeline_slug}` pair the worker currently expects).
2. Accepts the same filter params workers use today
   (`pipeline__slug__in`, `dispatch_mode`, `incomplete_only`).
3. Runs the heartbeat on every call (same Redis-throttled helpers already
   in `ami/jobs/views.py`).
4. Uses random ordering + `limit=1` internally — no paginator override.

Refactor plan:

- Extract the shared work into a helper (e.g. `_get_next_ml_job(request)`)
  called by both the new `next()` action and the existing
  `list(ids_only=True)` path.
- Mark `list(ids_only=True)` as deprecated in the OpenAPI schema and in
  comments, but keep it working until every running ADC worker has
  migrated.
- Once ADC is migrated (companion issue in ami-data-companion), remove:
  - the `order_by(\"?\")` override in `get_queryset()`
  - the paginator override in `JobViewSet.paginator`
  - the `ids_only=1` branch of `list()` — the heartbeat moves to `next()`

Each of the three removals is marked \`⚠️ TEMPORARY HACK — remove by
2026-04-24\` in code. That deadline is aspirational, not hard — treat as
\"start this as soon as #1261 lands.\"

## Related

- PR #1261 — the PR that introduced the hacks
- #1264 — scoping the heartbeat to the specific calling service (needs #1194)
- #1194 — client-ID / application-token auth

## Acceptance

- [ ] `GET /jobs/next/` action exists and returns a single claim
- [ ] ADC `trapdata/antenna/client.py::get_jobs` migrated to `/next`
- [ ] All three `⚠️ TEMPORARY HACK` markers in `ami/jobs/views.py` removed
- [ ] `list(ids_only=True)` either deleted or documented as the old contract

mihow:
[...]

PR #1276, opened 2026-04-29
fix(jobs): fix dangling jobs from going to revoked

mihow:
## Summary

The stale-job reaper (`check_stale_jobs`) is the safety net that finalizes jobs stuck in a running state past the cutoff. For async (pull-based) ML jobs it was making two unsafe calls: it trusted a Celery task's terminal state too readily, and it judged completeness from `Job.progress.is_complete()` — a JSONB blob that has been racy under concurrent result writes since #1261 dropped the row lock. The visible symptom was healthy async jobs being REVOKED even though their work had actually finished (NATS and Redis fully drained).

This PR makes the reaper consult the real source of truth (Redis pending-set completeness) before accepting a success, stops it from forcing a terminal FAILURE that the result handler is supposed to decide, and stops its writes from clobbering fields a concurrent writer owns. It also keeps the reaper's network calls off the database row lock so it doesn't contend with the live result handler.

### List of Changes

| # | Change (effect) | How |
|---|---|---|
| 1 | A stalled async job whose work is actually done is finalized as SUCCESS instead of being wrongly revoked. | For async_api jobs the reaper accepts a Celery SUCCESS only when `AsyncJobStateManager.all_tasks_processed()` reports the Redis pending sets are drained, instead of reading the racy `progress.is_complete()`. When Redis state is unavailable it falls back to `progress.is_complete()` and logs a warning. |
| 2 | A stalled async job whose Celery task ended in FAILURE is revoked rather than force-failed. | Celery FAILURE is no longer fast-pathed to a terminal status for async_api jobs — `update_job_failure()` defers the terminal outcome to the result handler, which decides SUCCESS/FAILURE from the final processed/failed counts against `FAILURE_THRESHOLD`. |
| 3 | The reaper no longer overwrites a concurrent writer's `logs` / `progress.errors`. | Terminal and revoke writes use `save(update_fields=[...])` (status, progress, finished_at, updated_at) instead of a full-row `save()` from a stale snapshot. |
| 4 | The reaper no longer prolongs lock contention with the live result handler. | Celery/Redis lookups now run in an unlocked first phase; a short `select_for_update()` second phase re-checks the staleness predicate and persists. If the job progressed or finalized in between, the row no longer matches and it is skipped. |
| 5 | New tri-state completeness helper. | `AsyncJobStateManager.all_tasks_processed()` → `True` (both pending sets empty / total 0), `False` (work remains), `None` (state unavailable: cleaned up, expired, never initialized, or RedisError). |

## Detailed Description

This sits alongside the status-write hardening in #1338 / #1342: those make the lock-free *status* writers safe; this makes the reaper's *decision* trustworthy and keeps its non-status JSONB writes (`logs`, `progress.errors`) from being clobbered. The reaper remains the deliberate last-resort writer — it keeps a broader from-set under its row lock and only *accelerates* a SUCCESS when Redis confirms the work is drained, otherwise it revokes.

`all_tasks_processed()` returns `None` for several distinct causes (missing keys vs transient RedisError). The single caller treats `None` as "cannot confirm" and falls back to `progress.is_complete()`, which is a safe degradation regardless of cause. If a future caller needs to distinguish causes, `AsyncJobStateManager.diagnose_missing_state()` (added in #1241) already exposes the richer detail.

### How to Test the Changes

- `pytest ami/jobs/tests/test_tasks.py::TestCheckStaleJobsReaperGuard`
- Full `ami/jobs/tests/test_tasks.py` + `ami/ml/tests.py` pass locally (107 passed, 2 skipped).

## Checklist

Refs #1337. Closes #1276's dangling-revoke symptom; complements #1338 / #1342.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened job status tracking for asynchronous operations by consulting multiple system state sources, preventing jobs from remaining in incorrect or stalled states when processing is disrupted.

* **Tests**
  * Added extensive test coverage for asynchronous job state management, including scenarios where internal systems may fall out of sync.

### DIFF
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -909,9 +909,10 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
 
     For each stale job, checks Celery for a terminal task status. REVOKED is
     always trusted. For async_api jobs, SUCCESS and FAILURE are only accepted
-    when job.progress.is_complete() — NATS workers may still be delivering
-    results after the Celery task finishes. All other cases result in revocation.
-    Async resources (NATS/Redis) are cleaned up in both branches.
+    when AsyncJobStateManager.all_tasks_processed() reports True (i.e. the
+    Redis pending sets are drained). When Redis state is absent, falls back to
+    job.progress.is_complete(). All other cases result in revocation. Async
+    resources (NATS/Redis) are cleaned up in both branches.
 
     Returns a list of dicts describing what was done to each job.
     """
@@ -961,12 +962,25 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
                     # Treat as unknown state — job will be revoked below.
 
             # Only trust terminal Celery states. For async_api jobs, SUCCESS and
-            # FAILURE are only accepted when progress is complete — NATS workers may
-            # still be delivering results after the Celery task finishes.
+            # FAILURE are only accepted when all NATS tasks are processed — workers
+            # may still be delivering results after the Celery task finishes.
+            # Consult Redis (source of truth for SREM completeness) directly rather
+            # than Job.progress.is_complete(), which mirrors a JSONB blob racy under
+            # concurrent _update_job_progress writes since #1261.
             is_terminal = celery_state in states.READY_STATES
             is_async_api = job.dispatch_mode == JobDispatchMode.ASYNC_API
-            if is_async_api and celery_state in {states.SUCCESS, states.FAILURE} and not job.progress.is_complete():
-                is_terminal = False
+            if is_async_api and celery_state in {states.SUCCESS, states.FAILURE}:
+                processed = AsyncJobStateManager(job.pk).all_tasks_processed()
+                if processed is False:
+                    is_terminal = False
+                elif processed is None:
+                    logger.warning(
+                        "Reaper for job %s: Redis state absent, falling back to " "progress.is_complete()",
+                        job.pk,
+                    )
+                    if not job.progress.is_complete():
+                        is_terminal = False
+                # processed is True -> trust Celery's terminal state
 
             previous_status = job.status
             if is_terminal:
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -975,11 +975,15 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
     healthy. Default cutoff is :attr:`Job.STALLED_JOBS_MAX_MINUTES`.
 
     For each stale job, checks Celery for a terminal task status. REVOKED is
-    always trusted. For async_api jobs, SUCCESS and FAILURE are only accepted
-    when AsyncJobStateManager.all_tasks_processed() reports True (i.e. the
-    Redis pending sets are drained). When Redis state is absent, falls back to
-    job.progress.is_complete(). All other cases result in revocation. Async
-    resources (NATS/Redis) are cleaned up in both branches.
+    always trusted. For async_api jobs, only SUCCESS is fast-pathed to a
+    terminal status, and only when AsyncJobStateManager.all_tasks_processed()
+    reports True (i.e. the Redis pending sets are drained); when Redis state is
+    unavailable it falls back to job.progress.is_complete(). Celery FAILURE is
+    not trusted for async_api jobs — update_job_failure() defers the terminal
+    outcome to the async result handler (FAILURE_THRESHOLD logic), so a stale
+    FAILED async_api job is revoked rather than forced to FAILURE. All other
+    cases result in revocation. Async resources (NATS/Redis) are cleaned up in
+    both branches.
 
     Returns a list of dicts describing what was done to each job.
     """
@@ -1028,33 +1032,54 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
                     )
                     # Treat as unknown state — job will be revoked below.
 
-            # Only trust terminal Celery states. For async_api jobs, SUCCESS and
-            # FAILURE are only accepted when all NATS tasks are processed — workers
-            # may still be delivering results after the Celery task finishes.
-            # Consult Redis (source of truth for SREM completeness) directly rather
-            # than Job.progress.is_complete(), which mirrors a JSONB blob racy under
-            # concurrent _update_job_progress writes since #1261.
+            # Only trust terminal Celery states. For async_api jobs, a SUCCESS
+            # Celery state is only accepted when all NATS tasks are processed —
+            # workers may still be delivering results after the Celery task
+            # finishes. Consult Redis (source of truth for SREM completeness)
+            # directly rather than Job.progress.is_complete(), which mirrors a
+            # JSONB blob racy under concurrent _update_job_progress writes since
+            # #1261.
+            #
+            # Celery FAILURE is deliberately NOT fast-pathed to a terminal
+            # status here: update_job_failure() defers post-queue run_job
+            # failures for async_api jobs to the async result handler, which
+            # decides the terminal outcome from the final processed/failed
+            # counts against FAILURE_THRESHOLD (a drained-but-failed Celery task
+            # can still resolve to SUCCESS). Trusting Celery FAILURE here would
+            # force the job to FAILURE and bypass that threshold logic, so a
+            # stale async_api job whose Celery task ended FAILURE falls through
+            # to the revoke branch instead.
             is_terminal = celery_state in states.READY_STATES
             is_async_api = job.dispatch_mode == JobDispatchMode.ASYNC_API
-            if is_async_api and celery_state in {states.SUCCESS, states.FAILURE}:
+            if is_async_api and celery_state == states.SUCCESS:
                 processed = AsyncJobStateManager(job.pk).all_tasks_processed()
                 if processed is False:
                     is_terminal = False
                 elif processed is None:
                     logger.warning(
-                        "Reaper for job %s: Redis state absent, falling back to " "progress.is_complete()",
+                        "Reaper for job %s: Redis state unavailable, falling back to " "progress.is_complete()",
                         job.pk,
                     )
                     if not job.progress.is_complete():
                         is_terminal = False
-                # processed is True -> trust Celery's terminal state
+                # processed is True -> trust Celery SUCCESS
+            elif is_async_api and celery_state == states.FAILURE:
+                # Don't treat Celery FAILURE as authoritative for async_api jobs
+                # (see comment above); revoke instead of forcing FAILURE.
+                is_terminal = False
 
             previous_status = job.status
             if is_terminal:
                 if not dry_run:
                     job.update_status(celery_state, save=False)
                     job.finished_at = datetime.datetime.now()
-                    job.save()
+                    # Narrow the write to the fields the reaper actually mutates.
+                    # A full save() would re-write the whole row from the snapshot
+                    # fetched at select_for_update() time, clobbering `logs` and
+                    # `progress.errors` that a concurrent _update_job_progress (no
+                    # row lock since #1261) may commit. update_status() touches
+                    # status + progress.summary.status, so `progress` is included.
+                    job.save(update_fields=["status", "progress", "finished_at", "updated_at"])
             else:
                 # Per-job diagnostic: surface enough state at revoke time that an
                 # operator can answer "why was this stalled?" without grepping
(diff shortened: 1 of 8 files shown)


mihow:
**Claude says:** (posted on Michael's behalf — part of a structural review pass over the four open PSv2-readiness PRs: #1276, #1279, #1312, #1324, toward the umbrella #910.)

The direction here is right: the reaper was making a destructive decision (REVOKE) based on the one mirror known to lie (`Job.progress` after a concurrent-writer clobber), and re-pointing it at the atomic Redis tracker is the correct re-ordering of trust. The tri-state design is careful — distinguishing `RedisError` from genuinely-absent state and refusing to REVOKE on a Redis blip — and the tests are strong, including a faithful reproduction of the job-clobber incident with raw-update staleness. The dispatch-mode gate keeps sync jobs on the old path, and the `updated_at`-based staleness trigger remains the right one for long-running jobs.

Things worth discussing before merge:

1. **The `is_terminal` fast-path may widen a known-wrong behavior for the FAILURE case** (echoing CodeRabbit's open comment, which looks legitimate). For `async_api`, celery's FAILURE on `run_job` is deliberately non-authoritative — `update_job_failure` defers it because the result handler owns the threshold-derived outcome. Post-PR, a job with drained Redis + celery FAILURE gets force-FAILED even when the failure count was below threshold and the natural outcome was SUCCESS. All five reaper tests use `celery_states.SUCCESS`, so the FAILURE branch is untested. Scoping the fast-path to SUCCESS-only looks like a two-line change that removes the concern.

2. **The fallback arm re-imports the bug this PR works around.** The same clobber race has a second shape: writer A sees completion → sets SUCCESS → cleanup deletes the Redis keys; a slower writer B then commits a stale snapshot including `status=STARTED`, resurrecting the job *after* Redis is gone. The reaper then gets `None` from `all_tasks_processed()` and falls back to the clobbered `progress.is_complete()` — same wrong REVOKE, new WARNING line. Not an argument against this PR (it's an honest mitigation), but it does mean the underlying fix — making progress writes atomic instead of read-modify-write — should be tracked as a first-class issue rather than a "follow-up" note. As far as we can find, no open issue owns that fix today: #1256 contains the contention analysis (its option D), #1285 names the race but scopes itself to diagnostics. We'll file a dedicated issue and link it here.

3. **Interaction with #1324 is worth a joint look before both merge.** #1324's `acks_late`/`reject_on_worker_lost` is built so a crashed worker's `run_job` redelivers and resumes; this PR's reaper REVOKEs anything stale past the cutoff, which quietly nullifies that resume for any outage longer than the window. Separately, #1324's cancel path tears down Redis while the job sits in CANCELING — making this PR's "Redis absent → fallback" branch much more common than the rare case it's described as. The guards do compose without zombies as far as we can trace, but neither PR currently mentions the other, and the combined behavior should be a deliberate choice.

4. **The bundled `docs/claude/planning/` and `NEXT_SESSION_PROMPT.md` files contain internal deployment specifics** (instance log paths, references to internal infra docs) that probably shouldn't live in a public repo, and they make the reviewable diff much larger than the code change. Suggest moving them out or trimming before merge.

Smaller cleanups, low priority: the plan doc's race analysis says a just-missed SREM defers to the next reaper tick, but the code REVOKEs in the same tick when `processed is False` — tiny window and matches pre-PR behavior, just worth correcting in the doc; and the open bot threads (None-conflation, Redis I/O under the row lock, double-WARNING) could use a reply or a fix each.

With item 1 scoped to SUCCESS-only and the docs trimmed, this looks mergeable as the mitigation it sets out to be — and the atomic-progress-writes issue becomes the load-bearing follow-up for the PSv2 default flip.

---

**Claude says:** Filed the atomic-progress-writes root-cause issue as promised: #1337 ("Saving job progress concurrently is the root of multiple issues related to incorrect job statuses"). It collects this PR, #1324, #1287, #1288, the prior attempts (#981 closed, #1026 draft), and the appearance-vs-failure breakdown — including why a wrong status isn't just cosmetic (it keeps the job claimable via `/next`, see #1265/#1282).

ami/jobs/tasks.py
```diff
@@ -961,12 +962,25 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
...
+            # Consult Redis (source of truth for SREM completeness) directly rather
+            # than Job.progress.is_complete(), which mirrors a JSONB blob racy under
+            # concurrent _update_job_progress writes since #1261.
             is_terminal = celery_state in states.READY_STATES
             is_async_api = job.dispatch_mode == JobDispatchMode.ASYNC_API
-            if is_async_api and celery_state in {states.SUCCESS, states.FAILURE} and not job.progress.is_complete():
-                is_terminal = False
+            if is_async_api and celery_state in {states.SUCCESS, states.FAILURE}:
+                processed = AsyncJobStateManager(job.pk).all_tasks_processed()
+                if processed is False:
```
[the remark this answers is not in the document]
mihow: **Claude says:** Done in 6c399176. The reaper no longer holds the row lock across the Celery/Redis round-trips. It now runs in two phases: Phase 1 fetches the Celery task state and queries Redis (`all_tasks_processed`) with no row lock; Phase 2 opens a short `select_for_update()` transaction, re-fetches with the same staleness predicate (still running + `updated_at < cutoff`), and persists. If the job progressed or finalized while we were off the lock, the row no longer matches and it's skipped, so the Phase 1 decision is never applied to a job that's no longer stale.

ami/ml/orchestration/async_job_state.py
```diff
@@ -228,6 +228,36 @@ def get_pending_image_ids(self) -> set[str]:
...
+        try:
+            redis = self._get_redis()
+            with redis.pipeline() as pipe:
+                for stage in self.STAGES:
+                    pipe.scard(self._get_pending_key(stage))
+                pipe.get(self._total_key)
+                results = pipe.execute()
+        except RedisError as e:
+            logger.warning(f"Redis error reading all_tasks_processed for job {self.job_id}: {e}")
+            return None
```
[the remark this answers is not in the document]
mihow: **Claude says:** Kept `bool | None` deliberately rather than adding an enum. The single caller (the reaper) treats `None` as "can't confirm via Redis" and falls back to `progress.is_complete()` — a safe degradation whether `None` came from missing keys or a transient RedisError, since the resulting decision is identical. The docstring already documents the three causes folded into `None`. If a second caller ever needs to branch on the cause, the richer signal already exists via `AsyncJobStateManager.diagnose_missing_state()` (added in #1241), so we can expose it then without changing this return type pre-emptively.

ami/jobs/tasks.py
```diff
@@ -961,12 +962,25 @@ def check_stale_jobs(minutes: int | None = None, dry_run: bool = False) -> list[
...
+                    is_terminal = False
+                elif processed is None:
+                    logger.warning(
+                        "Reaper for job %s: Redis state absent, falling back to " "progress.is_complete()",
```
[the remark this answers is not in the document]
mihow: **Claude says:** The warning now reads "Redis state unavailable, falling back to progress.is_complete()" (not "absent"), which covers both missing keys and a RedisError. On a RedisError there are two lines by design — `all_tasks_processed` logs the underlying error detail, and the reaper logs the fallback action it took — one names the cause, the other names the response.

ami/ml/migrations/0028_normalize_empty_endpoint_url_to_null.py
```diff
@@ -7,7 +7,6 @@ def normalize_empty_endpoint_url(apps, schema_editor):
 
 
 class Migration(migrations.Migration):
-
```
mihow, line 10: don't modify existing migrations please

ami/ml/migrations/0027_rename_last_checked_to_last_seen.py
```diff
@@ -2,7 +2,6 @@
 
 
 class Migration(migrations.Migration):
-
```
mihow, line 5: don't modify existing migrations please

ami/ml/migrations/0028_normalize_empty_endpoint_url_to_null.py
mihow: **Claude says:** Reverted in `27e24f8b` — both migrations are restored byte-for-byte to their form on main. They had only been touched by black removing a blank line after the `Migration` class declaration; nothing in this PR depends on them.

ami/ml/migrations/0027_rename_last_checked_to_last_seen.py
mihow: **Claude says:** Reverted in `27e24f8b` — both migrations are restored byte-for-byte to their form on main. They had only been touched by black removing a blank line after the `Migration` class declaration; nothing in this PR depends on them.

mihow:
**Claude says:** Closing this and folding the reaper hardening into #1337, where the root fix (atomic progress writes) lives.

Why: the core idea here — the stale-job reaper consulting Redis (`all_tasks_processed()`) as the source of truth before it REVOKEs, instead of the clobberable `progress.is_complete()` — is sound and is genuinely **not** covered by the merged #1338 / #1342 (those guard the status-write side). But a review pass found two blockers that make it net-risky to land standalone:

1. **It introduces a cancel-override bug.** The SUCCESS fast-path re-fetches with `status__in=running_states()` (which includes `CANCELING`) and writes via a plain `update_status()`, not the guarded transition. So a stale `CANCELING` job whose Celery task reports SUCCESS gets force-flipped to SUCCESS — overriding an in-flight cancel, the exact resurrection race #1338 was built to stop.
2. **The "does not clobber concurrent `progress.errors`" claim is false.** `save(update_fields=[…, "progress", …])` rewrites the whole `progress` JSONB from the reapers stale snapshot, and `errors` lives *inside* `progress` — so the narrow write protects other columns but not the two fields the comment names.

Since it is a mitigation (not the root fix) and adds a new bug, the reaper-consults-Redis improvement is better done as part of #1337 Layer 2. Captured the salvageable design there. The `all_tasks_processed()` tri-state helper is a good starting point and can be lifted from this branch (`fix/job-complete-check`).

mihow:
**Claude says:** Reopening this — but see the correction below, which retracts the main reason I originally gave.

---

### Correction (posted same day, after auditing my own claim against the branch)

My original comment said this PR "would have prevented the 99.997% revoke outright" and called its approach "the exact discrimination the reaper failed to make." **Both statements are wrong.** Retracting them.

The progress percentage and `all_tasks_processed()` read the **same** Redis pending set. On `main`, `ami/ml/orchestration/async_job_state.py` computes `remaining = scard(pending_key)` and then `processed = total - remaining`. So a job logged at `processed=37283/37284` requires `scard(pending) == 1` — that is what the leaked tracker entry *is*.

Feed that into this branch's `all_tasks_processed()` (`ami/ml/orchestration/async_job_state.py:231`), whose contract is explicit: *"False — at least one pending set has members."* One member means **False**, which means not terminal, which means the reaper revokes the job exactly as it did. The two revocations at zero percent are more clear-cut still: their pending sets were full.

This branch's own test asserts that outcome. `is_terminal` is also seeded from `celery_state in states.READY_STATES`, so the new guard can only turn True into False, never the reverse. **It changes the result for none of the three revocations.** I should have traced that before asserting it, and I appreciate the push to check.

I also wrote that reopening is "a rebase-and-review rather than a rewrite." That was too optimistic. Both blockers raised when this was closed are still unfixed at the branch head: the cancel-override path (`CANCELING` remains in `running_states()` and is written via a plain `update_status`), and the claim that Phase 2 does not clobber `progress.errors`.

### What this PR is still worth reopening for

The third commit, which splits the reaper's work into an unlocked Phase 1 doing the Celery and Redis I/O and a short locked Phase 2 doing the write, is worth having on its own merits. A production incident on 2026-07-29 produced a measured convoy of 125 database backends queued on a single `jobs_job` row, so shortening lock holds on that row now has evidence behind it rather than only reasoning.

Reopening on that basis, with the completeness-check argument withdrawn.

### A better lead for the underlying bug

While checking the above, something more promising turned up. The mechanism for reconciling exactly this failure is **already on `main`**: `mark_lost_images_failed` / `_reconcile_lost_images` (`ami/jobs/tasks.py:846-965`) runs *before* the stale check, fetches the pending image ids, marks them failed, and lets the job finalise. For a single leaked id that is precisely the right behaviour, and it should have saved the job.

So the question worth answering is not "how do we detect completeness differently" but **"why did the reconciler not fire?"** One candidate visible in the code is that it skips when the NATS consumer lookup returns `None` (`tasks.py:813-817`). That is a hypothesis, not a finding — but it is a cheaper and better-targeted line of attack than this PR, and anyone picking up the reaper work should start there.

PR #1281, opened 2026-04-30
Update agree logic for human identifications

### LINKED ISSUE #1226: UI: Agree button on human identification card sends wrong FK type
## Summary

The "Agree" button rendered inside a **human identification card** passes the Identification PK as `predictionId`, but `predictionId` is routed to the backend field `agreed_with_prediction_id` — a FK to `Classification`, not `Identification`.

## Location

`ui/src/pages/occurrence-details/identification-card/human-identification.tsx:105`

```tsx
<Agree
  ...
  agreeWith={{ predictionId: identification.id }}  // ← identification.id is a human Identification PK
  ...
/>
```

## Expected

When agreeing with a human identification, the UI should send:

```tsx
agreeWith={{ identificationId: identification.id }}
```

so the backend stores the value in `agreed_with_identification_id` (FK to `Identification`).

## Consequence

Depending on backend validation:

1. FK validation fails → the Agree click errors silently or with a generic rejection.
2. FK validation is lenient → the identification row ends up with a misrouted `agreed_with_prediction_id` pointing at whatever `Classification` happens to share that PK (or orphaned).

Either way, the data on `agreed_with_identification` is never populated from this surface, which also means analytics / exports that rely on that FK (e.g. the `agreed_with_user` column added in PR #1214) will undercount transitive agreement chains.

## Correct call sites (for reference)

These three call sites already pass both FKs correctly based on the current determination source:

- `ui/src/pages/occurrence-details/occurrence-details.tsx:176-179`
- `ui/src/pages/occurrences/occurrence-actions.tsx:48-51`
- `ui/src/pages/occurrences/occurrences-actions.tsx:73-76`

And the sibling file `machine-prediction.tsx:99,130` uses `predictionId` correctly because `identification.id` in that file refers to a `Classification`.

## Fix

One-line change at `human-identification.tsx:105`:

```diff
- agreeWith={{ predictionId: identification.id }}
+ agreeWith={{ identificationId: identification.id }}
```

## How to verify after fix

```sql
SELECT COUNT(*) FROM main_identification
WHERE agreed_with_identification_id IS NOT NULL AND withdrawn = FALSE;
```

should start growing as users click Agree on human identifications from the occurrence details page.

Discovered while reviewing PR #1214.

annavik:
### Background
When confirming a human identification from the history view, we were passing the wrong key to the API.

Screenshot 2026-04-30 at 15 24 26

### The solution
In this PR we adjust the key. This will allow the update, and also we can see more details in the export. Fixes #1226.

Screenshot 2026-04-30 at 15 33 01

### Notes
- I see some inconsistencies in the export. Sometimes we refer to users by name and sometimes by email. No biggie but could be nice to streamline.
- When confirming IDs from the list view, we also pass "agree with". If "agree with" will be a machine prediction or human identification depends on how the final determination was derived. Is this inline with how you would like it to be @mihow?

### DIFF
--- a/ui/src/pages/occurrence-details/identification-card/human-identification.tsx
+++ b/ui/src/pages/occurrence-details/identification-card/human-identification.tsx
@@ -102,7 +102,7 @@ export const HumanIdentification = ({
                           )
                         : false
                     }
-                    agreeWith={{ predictionId: identification.id }}
+                    agreeWith={{ identificationId: identification.id }}
                     applied={identification.applied}
                     occurrenceId={occurrence.id}
                     taxonId={identification.taxon.id}


mihow:
Thanks @annavik 
> When confirming IDs from the list view, we also pass "agree with". If "agree with" will be a machine prediction or human identification depends on how the final determination was derived. Is this inline with how you would like it to be @mihow?

That's tricky! The list view and the detail view both have a Confirm button which isn't next to a specific human or machine label. Do you have the data to clearly tell which type of source and which `id` was used for the final determination? 

image

mihow:
I'm glad to know that the validator was working! But also if the ID happens to exist on the other source type (there is a Classification and an Identification with id `1`) then I think it will outsmart the validation.

mihow:
> I see some inconsistencies in the export. Sometimes we refer to users by name and sometimes by email. No biggie but could be nice to streamline.

Should we make a full display name for the user field like Michael <notbot@gmail.com>? Not all users have a complete or unique value in the name field. Or just go with email?

mihow (approved):
Approved, but good questions in the PR description. I answered them in comments.

annavik:
> That's tricky! The list view and the detail view both have a Confirm button which isn't next to a specific human or machine label. Do you have the data to clearly tell which type of source and which `id` was used for the final determination?

Yes! The API returns an object `determination_details` for each occurrence. It has fields `identification` and `prediction`. One of them will be defined, depending on the determination source. This is what I inspect, to populate `agree_with` when a user confirms the current determination. This has been the current logic for some time, but I wanted to bring it up, since from UI, the user is not agreeing with an individual identification or prediction, only the current determination.

annavik:
> > I see some inconsistencies in the export. Sometimes we refer to users by name and sometimes by email. No biggie but could be nice to streamline.
> 
> Should we make a full display name for the user field like Michael [notbot@gmail.com](mailto:notbot@gmail.com)? Not all users have a complete or unique value in the name field. Or just go with email?

I like that idea! Optional name + email.

@coderabbitai can you open an issue about this? Include the related screenshot from the PR description and describe the suggested update.

annavik:
> Approved, but good questions in the PR description. I answered them in comments.

Thanks! I will merge this now, but I added a follow up task about the user display name: https://github.com/RolnickLab/antenna/issues/1290

Issue #1283, opened 2026-04-30
Clean up task queue when job is revoked or re-started, fix duplicate tasks


mihow:
## Bug

NATS task queue (stream + consumer) is not cleaned up when an async_api job is revoked or re-started. Two distinct lifecycle gaps:

### 1. Job marked REVOKED → stream + consumer left behind

The job-watchdog / healthcheck flips a stuck job to `REVOKED`, but the per-job NATS stream and consumer are untouched. Messages sit in the stream indefinitely.

### 2. User clicks "Start" on a non-running job → tasks re-published, stream not purged

Re-starting (from `REVOKED`, `FAILURE`, or `CANCELLED`) re-runs the `collect` stage and publishes a fresh round of tasks **on top of** whatever is still in the stream. The existing consumer's `delivered_seq` continues from where it was, so:

- Stream message count roughly doubles (or worse on multiple restarts).
- Already-ACKed tasks from the previous attempt that were re-published become duplicates that workers will process again, producing duplicate detections / classifications.
- Stage params are overwritten with fresh values, which manifests in the UI as the "Processed" counter visibly **resetting to 0** while the user is watching.

### Concrete observation

A real job hit both gaps in sequence today:

1. Job started, `collect` completed with 44,286 tasks → stream `job_<N>` published 44,286 messages.
2. Workers were starved by an unrelated `/next` filter bug (separate ticket: #1282), never delivered. Healthcheck flipped the job to `REVOKED` after ~18 min. Stream + consumer still present, 44,286 pending.
3. After the upstream filter bug was unblocked manually, user clicked "Start". Second `collect` ran, found 43,597 tasks (689 source images had soft-deleted in between), and published them on top of the existing 44,286.
4. Resulting NATS state: `messages = 87,883`, `delivered_seq` = wherever the consumer had reached, ~half the workload now duplicated across the two rounds.
5. UI showed the `process` stage counter snap from ~3,000 back to 0 and start counting up again, while detection / classification counts ran ahead of the processed-image count (reflecting re-detection of already-processed images).

### Suggested cleanup hooks

- **On status transition into `REVOKED` / `FAILURE` / `CANCELLED`** (whether by user, watchdog, or admin tooling): delete the consumer, purge or delete the stream. After that point the job is terminal; nothing should still be drainable.
[...]

PR #1284, opened 2026-04-30
fix(perms): restrict job deletion to superusers

mihow:
## Summary

The `MLDataManager` role granted `delete_job` from the initial roles framework (#693), and `ProjectManager` inherited it via the permission-set union. Both roles could `DELETE /api/v2/jobs/{id}/` and remove Job rows.

Job records are kept in the DB for audit / traceability — only superusers should delete them. Roles that need to stop in-flight work already have the `run_*_job` permission, which the cancel endpoint uses to revoke and clean up without dropping the row.

## Changes

- `ami/users/roles.py` — drop `DELETE_JOB` from `MLDataManager.permissions`. `ProjectManager` loses it via inherited union, no separate edit.
- `ami/main/migrations/0084_revoke_delete_job_from_roles.py` — data migration that strips `delete_job` from existing per-project `MLDataManager` and `ProjectManager` guardian groups, so deployed environments lose the perm immediately rather than waiting for the next `post_migrate` to re-run `create_roles_for_project`.
- `ami/jobs/tests/test_jobs.py` — new `TestJobDeletePermission` covering:
  - role permission sets no longer contain `delete_job`
  - `MLDataManager` and `ProjectManager` users get `403` on DELETE
  - superuser gets `204` and the row is gone

The `delete_job` permission codename remains a valid permission on `Project.Meta.permissions` — it's just not granted to any role. Superusers bypass via `is_superuser`.

## Out of scope (followup)

Currently a superuser DELETE drops the Job row but does **not** revoke the Celery task or clean up NATS/Redis async state. That belongs to issue #1283 (cleanup-on-revoke / cleanup-before-restart). Posted an implementation proposal at #1283#issuecomment-4356728574 with a generic `Job.terminate()` helper and the call sites that should funnel through it.

## Test plan

- [x] `python manage.py test ami.jobs.tests.test_jobs.TestJobDeletePermission --keepdb` — 5 tests pass locally
- [x] `python manage.py makemigrations --check --dry-run` — no missing migrations
- [x] `python manage.py migrate main` — 0084 applies cleanly on dev DB
- [ ] CI green
- [x] Manual: log in as MLDataManager / ProjectManager and confirm DELETE on a job returns 403 in the UI; superuser DELETE works

## Summary by CodeRabbit

* **Bug Fixes**
  * Job deletion is now restricted to superusers only. Users with data manager and project manager roles can no longer delete jobs.

* **Tests**
  * Added test coverage for job deletion permission enforcement.


mihow:
Confirmed that a project manager or owner does not see the Trash icon for jobs in the UI and cannot delete Job records, only cancel them.

image

PR #1300, opened 2026-05-12
Speed up the captures list view

mihow:
## Summary

Adds `select_related("deployment__data_source")` to `SourceImageViewSet.list`. Net diff is a single line in `ami/main/api/views.py` plus a regression test.

`SourceImage.public_url()` reads 8 S3-credential columns off `deployment.data_source` when the deployment has no `public_base_url`. Without the eager join, every such row fires one extra query during list serialization. Preloading collapses those into the main SELECT.

## What changed since the original draft

The earlier draft of this PR layered three things together: this `select_related`, an explicit `.only(*SOURCE_IMAGE_LIST_ONLY_FIELDS)` whitelist on the list queryset, and a denormalization of `SourceImageCollection` counts. Following review:

- **`.only()` was removed** (commit 04999abe). Per @mihow's pushback: the N+1 fix is `select_related`, not `.only()`. Future heavy fields will be caught by `django-zen-queries` + perf testing, and the whitelist friction outweighed the marginal row-width win on a 26-of-52-column table where the bulk are already small scalars.
- **Collection-count denormalization moved to #1301** (stacked on this branch). The paginator-COUNT win from rewriting `with_*_count` annotations as correlated subqueries was invisible to UX since the collections list does not paginate. #1301 denormalizes the 3 counts as cached columns instead.

What stays here: the `select_related` join + a regression test that pins the row shape (`url`, `size_display`, `deployment.name`, `event`, `detections_count`, `path`) and asserts no per-row lazy loads fire.

## Tests

- `TestSourceImageListQueryCount.test_list_response_shape_has_no_lazy_loads` — fails if the eager join is dropped or a serializer field is wired through a method that re-fetches.
- Existing `TestSourceImage*` and `TestOccurrence*QueryCount` suites pass unchanged.

## Out of scope

- Composite index `(project_id, timestamp DESC)` on `main_sourceimage` — needs `CREATE INDEX CONCURRENTLY` on prod + measured before/after.
- The `SOURCE_IMAGE_LIST_ONLY_FIELDS` constant and its `SOURCE_IMAGE_PUBLIC_URL_DEPENDENCIES` sub-group were committed and then reverted (8dfc3a0b → 04999abe). The reasoning trail lives in the planning docs under `docs/claude/planning/`.

## Test plan

- [x] `docker compose run --rm django python manage.py test ami.main.tests.TestSourceImageListQueryCount` — 4/4 pass cold-cache
- [ ] Staging arbutus-2026 deploy + NR p50/p95/p99 check on `/api/v2/captures/`


ami/main/api/views.py
```diff
@@ -542,6 +542,32 @@ def get_queryset(self) -> QuerySet:
...
+            # is loaded so the FK can short-circuit to None without a lazy fetch.
+            queryset = queryset.only(
+                "id",
+                "path",
+                "timestamp",
+                "width",
+                "height",
+                "size",
+                "public_base_url",
+                # `detections_count` is cached on the row; `occurrences_count` and
```
mihow: detections_count is different than `occurrences_count` and `detections_count` is included 2 lines below this one anyway. fix the comment?

ami/main/api/views.py
```diff
@@ -542,6 +542,32 @@ def get_queryset(self) -> QuerySet:
         ).order_by("timestamp")
 
         if self.action == "list":
+            # Trim row width to fields SourceImageListSerializer reads.
+            # SourceImage.public_url() needs `path` + `public_base_url`; `deployment.data_source_id`
+            # is loaded so the FK can short-circuit to None without a lazy fetch.
+            queryset = queryset.only(
```
mihow: how was this list determined? how can we maintain it? would exclude on the hot columns work instead? or a constant for list view fields? or detect the list by the serializer's needs? happy to keep it simple if we can make it clear, even if only with some instructions in the docstring

mihow:
@coderabbitai ready for review! our outstanding question is this even necessary if we cache the detections_count filed

ami/main/api/views.py
mihow: **Claude says:** Fixed in 8ec3e81e. Dropped the misleading inline comment entirely. Maintenance notes for the whole field list now live in the docstring above the new `SOURCE_IMAGE_LIST_ONLY_FIELDS` constant — clarifies that `detections_count` is a cached DB column while `occurrences_count` / `taxa_count` are annotation-only (not in the constant).

ami/main/api/views.py
mihow: **Claude says:** Refactored in 8ec3e81e. Pulled the field list into a module-level `SOURCE_IMAGE_LIST_ONLY_FIELDS` constant with a docstring describing the maintenance rules: every DB column the serializer reads + every column model methods touch on the serialized fields. The test `test_list_response_shape_preserved_after_only` fails loudly if a missing column causes a lazy load during serialization, so the constant has a behavioral guard.

Considered `.defer()` on hot columns instead — rejected because the hot column set is small relative to the cold one (e.g. checksum/last_modified/test_image/uploaded_by_id) and the inclusion list is more explicit about *why* each column is needed. Considered deriving from `serializer.Meta.fields` — rejected because model methods (e.g. `public_url()`) read columns the serializer doesn't directly declare, so full derivation isn't possible. The constant + docstring + test guard felt like the best compromise.

ami/main/api/views.py
```diff
@@ -542,6 +542,32 @@ def get_queryset(self) -> QuerySet:
...
+                "deployment__id",
+                "deployment__name",
+                "deployment__project_id",
+                "deployment__data_source_id",
+                "event_id",
+                "event__id",
+                "event__start",
+                "event__end",
+                "event__deployment_id",
+            )
```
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in 8ec3e81e. Added `select_related("deployment__data_source")` and extended the preload list with the seven S3StorageSource columns `SourceImage.public_url()` touches (bucket, region, prefix, access_key, secret_key, endpoint_url, public_base_url). Rows whose cached `public_base_url` is blank now resolve the presigned-URL fallback without firing a per-row query.

mihow:
**Claude says:** Pushed 8ec3e81e addressing all 4 review comments:

1. **mihow:556** (misleading comment) — dropped, maintenance notes moved to docstring above the new `SOURCE_IMAGE_LIST_ONLY_FIELDS` constant.
2. **mihow:548** (`.only()` maintainability) — refactored into module-level constant + docstring with maintenance rules. Behavioral guard already exists via `test_list_response_shape_preserved_after_only`.
3. **coderabbit:570** (`data_source` N+1) — added `select_related("deployment__data_source")` + the 7 S3StorageSource columns `public_url()` reads.
4. **coderabbit nitpick (tests.py:3297)** — bumped `TestSourceImageCollectionListQueryCount` fixtures from 2 → 30 collections so `limit=25` exercises a real page boundary.

All 16 tests in `TestSourceImage*` / `TestSourceImageCollection*` pass locally.

The collection-counts denormalize-vs-CTE question (E1/E2/E3 in the PR body) is still open — flagged as follow-up work, not blocking this PR.

ami/main/api/views.py
```diff
@@ -472,6 +472,55 @@ def list(self, request, *args, **kwargs):
...
+#   2. Run TestSourceImageListQueryCount — the response-shape test fails loudly if a
+#      lazy load fires during serialization, and the scaling tests fail if a missing
+#      column causes per-row queries.
+#
+# Notes on what is and is NOT here:
+#   - `detections_count` is a cached DB column on SourceImage and IS preloaded.
+#   - `occurrences_count` and `taxa_count` are NOT columns on SourceImage — they exist
+#     only when `with_counts=true` adds them as annotations. Listing them here would
+#     raise FieldDoesNotExist.
+#   - The `deployment__data_source__*` set is needed by SourceImage.public_url()'s
```
mihow: when we have an image resize service, perhaps we can remove the raw/authenticated image URLs and make this field _not_ dynamic

ami/main/api/views.py
```diff
@@ -472,6 +472,55 @@ def list(self, request, *args, **kwargs):
...
+#     presigned-URL fallback when SourceImage.public_base_url is blank but the
+#     deployment's data_source has credentials.
+SOURCE_IMAGE_LIST_ONLY_FIELDS = (
```
mihow: Do we really need this? why can't we use exclude?

ami/main/api/views.py
```diff
@@ -472,6 +472,76 @@ def list(self, request, *args, **kwargs):
...
+    # Core fields read by the serializer (id, timestamps, dimensions, cached counts, FKs).
+    "id",
+    "timestamp",
+    "width",
+    "height",
+    "size",
+    "detections_count",
+    "project_id",
+    "deployment_id",
+    # Nested DeploymentNestedSerializer ({id, name, details}) and permission walk.
```
mihow: Now I like the idea of assigning each of these dependent groups to a variable (DeploymentNestedSerializer) etc. However they can be part of the class instead of module variables. reduce the doc string a bit. Make it clear why we can't use exclude in this case, link to an external django reference if there's a helpful one.

ami/main/api/views.py
mihow: **Claude says:** Captured in the new in-class comment block in 3fc8b730d312d662830bfb443bb84ed002adb889 — the public-url group is now `_LIST_PUBLIC_URL_FIELDS` and the docstring above it marks the whole tuple TEMPORARY pending the image-resize/CDN layer. When that lands, dropping the group should be a single deletion.

ami/main/api/views.py
mihow: **Claude says:** Yes — needed to keep the SELECT narrow. Reframed in 3fc8b730d312d662830bfb443bb84ed002adb889 so the rationale is in the code:

`.only()` is whitelist, `.defer()` is blacklist (`exclude()` is for filtering rows, not columns). Two reasons whitelist wins here:
1. **Stable surface.** The serializer fields are the contract — they change rarely. The model gains wide/heavy columns over time (JSONField, ArrayField, text). Whitelisting pins us to the contract; blacklisting forces us to chase every new model column.
2. **Failure mode.** Forget to add a new model column to a blacklist → silent perf regression. Forget to add a new serializer column to a whitelist → loud test failure in `TestSourceImageListQueryCount` because the lazy load fires.

Docs ref: https://docs.djangoproject.com/en/4.2/ref/models/querysets/#only

ami/main/api/views.py
mihow: **Claude says:** Done in 3fc8b730d312d662830bfb443bb84ed002adb889. Both constants are now class attributes on `SourceImageViewSet`, split into 4 groups by purpose:

- `_LIST_CORE_FIELDS` — id/dimensions/cached counts/FKs
- `_LIST_DEPLOYMENT_FIELDS` — nested DeploymentNestedSerializer
- `_LIST_EVENT_FIELDS` — nested EventNestedSerializer
- `_LIST_PUBLIC_URL_FIELDS` — `SourceImage.public_url()` (marked TEMPORARY)

`LIST_ONLY_FIELDS` is just `*all four`. Docstring trimmed to one paragraph above the groups; the why-not-defer note links to the Django ref. `get_queryset` now calls `self.LIST_ONLY_FIELDS`.

mihow: **Claude says:** Refined further in 8dfc3a0beb9521e4b2c12ec04b20c6580858e5e7 per follow-up: dropped the `LIST_ONLY_FIELDS` aggregate (redundant when the groups are named) and dropped the `_LIST_` prefix from each group. `get_queryset` now unpacks all four directly: `only(*self._CORE_FIELDS, *self._DEPLOYMENT_FIELDS, *self._EVENT_FIELDS, *self._PUBLIC_URL_FIELDS)`.

ami/main/api/views.py
mihow: I don't think this is correct. If 1 & 2 are the only reasons, then .only() is not needed. Regressions by new big fields or methods will be caught by django-zen-queries and perf testing. To me only() is more frustrating when we have this many fields. From your previous explanation: there a single field that is exploding the query: the public_url() method field - which requires we prefetch fields to avoid the lazy n+1 queries. Is that not the true reason? So .only() is acting as a prefetch, not a whitelist mechanism.

ami/main/api/views.py
```diff
@@ -483,6 +483,59 @@ class SourceImageViewSet(DefaultViewSet, ProjectMixin):
     GET /captures/1/
     """
 
+    # Columns preloaded by `.only()` for the list response. We whitelist (`only`) instead of
+    # blacklist (`defer`) because the serializer surface is the stable contract, while
+    # SourceImage gains wide/heavy columns over time that we don't want pulled into every
+    # list row. See https://docs.djangoproject.com/en/4.2/ref/models/querysets/#only
```
mihow: I don't believe this is the reason.

ami/main/api/views.py
```diff
@@ -483,6 +483,59 @@ class SourceImageViewSet(DefaultViewSet, ProjectMixin):
...
+    # Columns read by the nested EventNestedSerializer ({id, name, details, date_label}).
+    _EVENT_FIELDS = (
+        "event_id",
+        "event__id",
+        "event__start",
+        "event__end",
+        "event__deployment_id",
+    )
+
+    # TEMPORARY — columns `SourceImage.public_url()` reads when `public_base_url` is blank
```
mihow: All of the only() fields are temporary. Once the public_url() method is removed from the list, then we can go back to auto detecting fields without n+1 queries

ami/main/api/views.py
mihow: **Claude says:** You're right — I had the framing wrong. Dropped `.only()` entirely in 04999abe80fad69842b557062219ce9119e28af8.

Empirically verified: with `select_related("event", "deployment", "deployment__data_source")` alone, the captures list query count stays flat at 5 / 25 rows (`TestSourceImageListQueryCount` still green). `.only()` was a row-width trim on top, not the N+1 fix.

The actual N+1 fix in this PR is now just one line: adding `"deployment__data_source"` to `select_related` so `SourceImage.public_url()` doesn't lazy-load the data_source chain per row. zen-queries + the existing query-count tests cover regressions from new big fields/methods.

ami/main/api/views.py
mihow: **Claude says:** Moot now — the whole `_PUBLIC_URL_FIELDS` block is gone in 04999abe80fad69842b557062219ce9119e28af8. Only `select_related("deployment__data_source")` remains, which is enough on its own.

ami/main/api/views.py
mihow: **Claude says:** Agreed — went one step further in 04999abe80fad69842b557062219ce9119e28af8 and dropped `.only()` entirely. select_related alone keeps the query count flat (`TestSourceImageListQueryCount` confirms). When the CDN/image-resize layer lands and `public_url()` goes away, we can also drop `deployment__data_source` from select_related.

mihow:
**Claude says:** PR body updated to reflect current state — `.only()` machinery dropped per @mihow's review, net diff is now a single `select_related("deployment__data_source")` line plus a regression test. Collection-count work moved to #1301 (stacked on this branch).

@coderabbitai full review please. Sister PR #1301 stacks on this branch and ports the collection-count denormalization out of the original draft.

Issue #1288, opened 2026-05-01
perf(jobs): defer aggregate count refresh in process_nats_pipeline_result (5-10x speedup expected)


mihow:
## Summary

`process_nats_pipeline_result` averages **20-33s per task in production** (p99 78s, occasional 300s SoftTimeLimit hits). About **98% of that time is in DB**, and Sentry trace data shows ~83% of all task wall time is consumed by **three denormalized DISTINCT COUNT queries** triggered by `Event.update_calculated_fields()` and `Deployment.update_calculated_fields()` running on **every result message**.

Each result message → save Detection / Classification / Occurrence rows → cascading saves trigger `update_calculated_fields_for_events` → that runs aggregate counts over the entire deployment / event. With thousands of result messages per job, the same expensive aggregate is recomputed thousands of times.

The actual writes (`UPDATE main_sourceimage` ~11ms, `INSERT jobs_joblog` ~7ms) are cheap. The latency is entirely the **count refresh cascade**.

## Observed (last 1h prod sample, n=11,662 task transactions)

| Query shape | avg | p95 | sum (% of all task time) |
|---|---|---|---|
| `SELECT COUNT(*) FROM (SELECT DISTINCT main_detection ... INNER JOIN sourceimage/occurrence/taxon WHERE deployment_id = ... + project filters)` | 13.7s | 31s | **~50%** |
| Same shape, `WHERE event_id = ...` | 7.0s | — | **~26%** |
| `SELECT COUNT(*) FROM (SELECT DISTINCT main_sourceimage ... WHERE event_id = ...)` | 1.9s | — | **~7%** |

p50 task latency 12.6s, p95 50.9s, p99 80.7s. The 300s tail = `SoftTimeLimitExceeded` on these queries → NATS redeliveries → cascade.

## Root cause (verified locally)

- `ami/main/models.py:1217` `Event.get_detections_count` and `:1233` `Event.get_occurrences_count` — both apply `build_occurrence_default_filters_q` and run `.distinct().count()` over Detection joined to SourceImage / Occurrence / Taxon with the project's include/exclude-taxa M2Ms
- `ami/main/models.py:3080-3081` — Deployment equivalent
- Both fire from `Event.update_calculated_fields()` (`models.py:1307-1309`) and the matching Deployment `update_calculated_fields()`, which are called from `update_calculated_fields_for_events` invoked **per result message** inside `ami/ml/models/pipeline.py:save_results`

### 1. Defer the aggregate refresh

[...]

PR #1301, opened 2026-05-13
Speed up the capture set list view. Introduce the CachedCountField.

mihow:
## Summary

Two changes, both targeting `SourceImageCollectionViewSet.list` query cost on large M2M tables:

1. **Denormalize the 3 source-image counts** as cached columns on `SourceImageCollection`. Replaces 3 per-row `Subquery(Count(...))` annotations with a plain column SELECT; O(1) read regardless of M2M size.
2. **Introduce `CachedCountField`** in `ami/base/models.py` — a marker subclass of `IntegerField` — and apply it to all 12 existing cached count columns project-wide. Discoverable via `Model._meta.get_fields()` + `isinstance(f, CachedCountField)` so future cross-cutting tasks (admin display, refresh jobs, list-endpoint defer) can enumerate them.

**Stacked on #1300** (`select_related` on captures list). Base branch is `perf/list-endpoint-subquery-rewrite`; will retarget to `main` once #1300 merges.

## Denormalization details

- 3 new columns on `SourceImageCollection`: `source_images_count`, `source_images_with_detections_count`, `source_images_processed_count`. Schema migration `0085_denormalize_sourceimagecollection_counts` creates them and backfills via a single GROUP BY + `FILTER` scan of the M2M (computes all three counts in one pass). `atomic=False` so the UPDATE can chunk on prod-sized data.
- `SourceImageCollection.get_source_image_counts()` returns the 3 counts as a dict from a single aggregate query (no writes — pure compute, testable in isolation). `update_calculated_fields(save=True)` is the side-effecting wrapper that also writes the columns.
- Signal handlers in `ami/main/signals.py`:
  - `m2m_changed` on `SourceImageCollection.images.through` for `post_add`, `post_remove`, `post_clear`.
  - `Detection` `post_save` / `post_delete` recompute for any collection containing the affected `source_image`.
- `pipeline.save_results()` explicitly recomputes for affected collections after its `bulk_create` batch, since `bulk_create` skips signals.
- Drops the 3 `with_*_count()` annotation methods on `SourceImageCollectionQuerySet` and removes them from the viewset chain.

## CachedCountField details

- New class in `ami/base/models.py` — `class CachedCountField(models.IntegerField)`. Pure marker; no field-level behavior change.
- Applied to: `Deployment.{events,occurrences,captures,detections,taxa}_count`, `Event.{captures,detections,occurrences}_count`, `SourceImage.detections_count`, `SourceImageCollection.{source_images,source_images_with_detections,source_images_processed}_count`.
- Migration `0086_use_cached_count_field` is recorded but the generated SQL is `(no-op)` for every `AlterField` — the column type is unchanged. The migration only updates Django's recorded model state so future operations see the right field class.
- Kept **separate** from the 0085 backfill migration so the schema-and-data migration is auditable on its own.

## Why denormalize rather than the subquery rewrite from the earlier draft of #1300

The collections list UI does **not** paginate. The paginator `COUNT(*)` win from rewriting the annotations as correlated subqueries (the original #1300 approach) is invisible to UX. The remaining cost was 3 per-row count subqueries against the M2M + `main_detection`; on a collection holding 500k+ images, each request still spent hundreds of ms per subquery.

Bench data from the earlier draft also showed the subquery rewrite was **slower** than the original `Count(images, filter=..., distinct=True)` on small projects and roughly equal on the heaviest project. Cached columns are an unambiguous win across all project sizes.

Precedent: `Deployment` already had 5 cached counts (`events_count`, `occurrences_count`, `captures_count`, `detections_count`, `taxa_count`). The `CachedCountField` marker formalizes that pattern.

## Recompute granularity tradeoff

Per-Detection signal recompute fires on every individual `Detection.save()` / `Detection.delete()`. For ML pipeline writes this is a non-issue: `Detection.objects.bulk_create()` skips signals, and `pipeline.save_results()` batches one recompute per job. The per-row signal handles non-bulk paths (manual workflows, retroactive ML, admin actions).

## Tests

- **New** `TestSourceImageCollectionCountsDenormalize` (7 cases) — image add/remove, Detection create/delete, null-bbox processed-but-not-with-detections case, `update_calculated_fields()` recovery from drift, pure-compute `get_source_image_counts()` returns dict without writes. A tautological `test_initial_counts_zero` was dropped during review — it asserted Django's `default=0`, not our code.
- **New** `TestSourceImageCollectionListQueryCount` (3 cases) — page-size scaling, `with_counts=true`, `ordering=-source_images_count` (sort hits the cached column directly, no extra subquery).
- Full `ami.main` + `ami.ml` + `ami.jobs` suites pass on local.

## Rollout

- 0085 creates columns with `default=0` and backfills in one deploy. Backfill is the slow step on prod-sized data — measure on staging arbutus-2026 first.
- 0086 is no-op SQL.
- Signals fire on new writes immediately after deploy. The viewset switches from annotation to column reads in the same commit, so the deploy must include both the migration and the code.

## Test plan

- [x] Local: full `ami.main` + `ami.ml` + `ami.jobs` test suites
- [ ] Staging arbutus-2026: measure 0085 backfill duration on production-shape M2M
- [ ] Staging: NR check that `/api/v2/captures/collections/` p99 drops vs production

## Release Notes

* **New Features**
  * Collection counts (total images, images with detections, processed images) are now denormalized for improved query performance.
  * Cached counts across collections, deployments, events, and source images automatically stay in sync when related data changes.
  * Added methods to manually refresh cached counts when needed.

* **Chores**
  * Database migrations applied to add denormalized count fields and backfill existing data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RolnickLab/antenna/pull/1301?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ami/main/signals.py
```diff
@@ -197,3 +197,55 @@ def exclude_taxa_updated(sender, instance: Project, action, **kwargs):
...
+
+@receiver(m2m_changed, sender=SourceImageCollection.images.through)
+def update_collection_counts_on_m2m(sender, instance, action, reverse, pk_set, **kwargs):
+    """Recompute denormalized counts when collection<->image links change.
+
+    Handles add/remove from either side. Forward post_clear knows the collection;
+    reverse post_clear cannot identify affected collections from signal payload,
+    so callers must invoke ``update_calculated_fields`` explicitly in that case.
+    """
+    if action not in ("post_add", "post_remove", "post_clear"):
```
mihow: this looks generic and familiar. review our m2m signal for when team members on a project change

ami/ml/models/pipeline.py
```diff
@@ -999,6 +999,17 @@ def save_results(
     for deployment in Deployment.objects.filter(pk__in=deployment_ids):
         deployment.update_calculated_fields(save=True)
 
+    # bulk_create above skips Detection signals; recompute affected collections explicitly.
```
mihow: Can we trigger the signal manually rather than rewrite the update code? Trigger all relevant signals manually after a bulk update? Otherwise make a clearly named method we can use like `update_calculated_fields(Model, pks/queryset)`

ami/main/signals.py
mihow: **Claude says:** Aligned in 39a8f02fd7d3993568614c45bfb762de3aa6e60c. Slimmed the signal to match the project members pattern at `signals.py:48`: drop the `reverse` and `post_clear` branching, drop the `_recompute_collection_counts` helper, just dispatch on `action` and call `instance.update_calculated_fields(save=True)`. The `reverse` direction was never used in the codebase (grep confirmed all callers use `collection.images.add/remove/set`).

ami/ml/models/pipeline.py
mihow: **Claude says:** Refactored in 39a8f02fd7d3993568614c45bfb762de3aa6e60c. Now matches the `Deployment.update_calculated_fields` pattern just above it at `pipeline.py:999`:

```python
deployment_ids = {img.deployment_id for img in source_images if img.deployment_id}
for deployment in Deployment.objects.filter(pk__in=deployment_ids):
    deployment.update_calculated_fields(save=True)

# bulk_create above skips Detection signals; refresh affected collections explicitly.
source_image_ids = [img.pk for img in source_images]
for collection in SourceImageCollection.objects.filter(images__id__in=source_image_ids).distinct():
    collection.update_calculated_fields(save=True)
```

Dropped the redundant local import + intermediate `collection_ids` step. The named method `update_calculated_fields(save=True)` is the single entry point everywhere — m2m signal, detection signals, and this bulk hook all call the same thing.

ami/main/models.py
```diff
@@ -4205,6 +4179,12 @@ class SourceImageCollection(BaseModel):
         default=dict,
     )
 
+    # Denormalized counts. Kept in sync via m2m_changed and pipeline-completion
+    # hooks. Reads are O(1).
+    source_images_count = models.IntegerField(default=0)
```
mihow: How can we indicate which fields are cached counts? field properties? prefix? I believe we have a cached_ prefix on other models, but then it makes the method replacements not 1:1. maybe thats okay. open to ideas.

ami/main/models.py
```diff
@@ -4240,6 +4207,28 @@ def taxa_count(self) -> int | None:
...
+        any_det = Detection.objects.filter(source_image=models.OuterRef("pk"))
+        counts = self.images.annotate(
+            _has_any_det=Exists(any_det),
+            _has_valid_det=Exists(valid_det),
+        ).aggregate(
+            total=models.Count("id"),
+            processed=models.Count("id", filter=models.Q(_has_any_det=True)),
+            with_det=models.Count("id", filter=models.Q(_has_valid_det=True)),
+        )
+        self.source_images_count = counts["total"]
```
mihow: move the get counts into it's own testable method (get counts without writing)

ami/main/models.py
mihow: **Claude says:** Checked — no `cached_` prefix exists on any model field in the codebase. All denormalized count fields use the bare `<thing>_count` pattern: `Deployment.events_count`, `Deployment.captures_count`, `Event.captures_count`, `SourceImage.detections_count`, etc. The new fields match this convention.

I'd lean against adding a `cached_` prefix just to this PR's fields — it would create a split where some counts are prefixed and most aren't, and method names (`occurrences_count`, `taxa_count` on the same model) wouldn't be 1:1 anymore as you noted. If we want the prefix project-wide, that's a separate refactor PR covering all the count fields.

Happy to do either — flag if you'd prefer the new fields prefixed and we adopt project-wide later, or leave matching the existing convention.

ami/main/models.py
mihow: **Claude says:** Done in 401a61350a70445ee0726c6ceabc2e210b20facc. Split into:

```python
def get_source_image_counts(self) -> dict[str, int]:
    """Compute the 3 source-image counts in a single query. No writes — testable."""
    ...
    return counts  # keyed by the field names

def update_calculated_fields(self, save: bool = False) -> None:
    counts = self.get_source_image_counts()
    self.source_images_count = counts["source_images_count"]
    ...
    if save:
        SourceImageCollection.objects.filter(pk=self.pk).update(**counts)
```

The return dict keys match the field names, so the `save` path is a clean `**counts` splat. Added `test_get_source_image_counts_returns_dict_without_writes` that asserts the dict shape AND that calling `get_source_image_counts()` doesn't mutate the DB row.

ami/main/models.py
mihow: Is there a simple way to mark/tag/categorize/group model fields? To indicate these are cached counts that could be potentially stale or empty at any time? declare in a list on the model? add a model field property, etc?

mihow:
**Claude says:** PR body updated. Two new pieces since the initial draft:
1. `CachedCountField` marker class added to `ami/base/models.py` and applied to all 12 existing cached-count columns project-wide (Deployment×5, Event×3, SourceImage×1, SourceImageCollection×3). Migration `0086_use_cached_count_field` is recorded but the generated SQL is no-op for every `AlterField` — column type unchanged. Kept separate from the 0085 backfill migration on purpose.
2. Dropped a tautological `test_initial_counts_zero` (asserted Django's `default=0`, not our code) and rewrote a handful of em-dash fragment comments added during earlier review rounds.

@coderabbitai full review please. Base branch is `perf/list-endpoint-subquery-rewrite` (sister PR #1300, which strips down to a single `select_related` line). Will retarget to `main` once #1300 merges.

ami/base/models.py
```diff
@@ -7,6 +7,18 @@
 from ami.users.models import User
 
 
+class CachedCountField(models.IntegerField):
```
mihow, line 10: @mohamedelabbas1996 I think we discussed a field type for cached counts last year. it finally happened!

ami/main/models.py
```diff
@@ -4205,6 +4179,12 @@ class SourceImageCollection(BaseModel):
         default=dict,
     )
 
+    # Denormalized counts. Kept in sync via m2m_changed and pipeline-completion
+    # hooks. Reads are O(1).
+    source_images_count = CachedCountField(default=0)
+    source_images_with_detections_count = CachedCountField(default=0)
+    source_images_processed_count = CachedCountField(default=0)
```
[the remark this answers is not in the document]
mihow: Great catch

mihow:
**Claude says:** Self-suggestion for follow-up PR (or fold into this one): simplify the cached-counts plumbing introduced here. Current shape works but has three per-model pieces (dedup state attr + flush helper + per-model Celery task). Next cached-count addition repeats all three.

## Proposed consolidation

One free helper + queryset/instance methods on `BaseModel` / `BaseQuerySet`. Fits the established `run_async=` flag pattern (`Deployment.save(regroup_async=)`, `Job.retry(async_task=)`, `process_single_source_image(run_async=)`).

```python
# ami/base/cached_counts.py
_PENDING_ATTR = "_pending_cached_count_recomputes"

def _schedule_recompute(model, pk):
    """Per-connection dedup + transaction.on_commit dispatch.

    Outside an atomic block, on_commit fires synchronously at registration
    time, so we set the pending set BEFORE registering the flush callback.
    """
    pending = getattr(connection, _PENDING_ATTR, None)
    is_new = pending is None
    if is_new:
        pending = set()
        setattr(connection, _PENDING_ATTR, pending)
    pending.add((model._meta.label, pk))
    if is_new:
        transaction.on_commit(_flush)

def _flush():
    pending = getattr(connection, _PENDING_ATTR, set())
    try: delattr(connection, _PENDING_ATTR)
    except AttributeError: pass
    for label, pk in pending:
        recompute_cached_counts_task.delay(label, pk)

@shared_task(ignore_result=True)
def recompute_cached_counts_task(model_label, pk):
    model = apps.get_model(model_label)
    try: instance = model.objects.get(pk=pk)
    except model.DoesNotExist: return
    instance.update_calculated_fields(save=True)
```

```python
# ami/base/models.py
class BaseModel(models.Model):
    def update_cached_counts(self, run_async: bool = True) -> None:
        if run_async:
            _schedule_recompute(type(self), self.pk)
            return
        self.update_calculated_fields(save=True)

class BaseQuerySet(QuerySet):
    def update_cached_counts(self, run_async: bool = True) -> None:
        for pk in self.values_list("pk", flat=True):
            if run_async:
                _schedule_recompute(self.model, pk)
            else:
                self.model.objects.get(pk=pk).update_calculated_fields(save=True)
```

## Call sites collapse

Pipeline `save_results`:
```python
SourceImageCollection.objects.filter(
    images__id__in=source_image_ids
).update_cached_counts(run_async=False)
```

Signals:
```python
@receiver(post_save, sender=Detection)
@receiver(post_delete, sender=Detection)
def update_collection_counts_on_detection_change(sender, instance, **kwargs):
    if not instance.source_image_id:
        return
    SourceImageCollection.objects.filter(
        images__id=instance.source_image_id
    ).update_cached_counts()  # async by default
```

Reconcile loop in `ami/main/checks/cached_counts.py` swaps `update_calculated_fields(save=True)` → `update_cached_counts(run_async=False)`.

## What goes away

- `_PENDING_SOURCE_IMAGE_IDS_ATTR` + `_flush_pending_collection_refreshes` + `_schedule_collection_refresh_for_source_image` in `signals.py` (per-model state attrs disappear)
- `refresh_collection_cached_counts(collection_id)` task in `ami/main/tasks.py` (replaced by generic `recompute_cached_counts_task(model_label, pk)`)
- Per-model signal handlers that need to know which Celery task to call

## What stays

- `CachedCountField` marker (still drives reconcile discovery)
- Per-model `update_calculated_fields(save=True)` (recompute logic lives where the model lives)
- Periodic `reconcile_cached_counts_task` (safety net for bulk-write paths that bypass signals)
- `.filter(pk=).update()` vs `.save()` shape divergence in `update_calculated_fields` — documented inline, leaving alone

## Next cached-count addition

Three things instead of six:
1. Field: `foo_count = CachedCountField()`
2. Recompute body in `update_calculated_fields(save)`
3. One signal handler per write source, each ending in `.update_cached_counts()`

Reconcile + drift detection are free.

## What I'd NOT do

- Declarative `invalidate_on=[...]` on the field — hides signal wiring, adds magic
- Drop signals entirely + rely on TTL reconciliation — breaks interactive freshness (just-uploaded image not showing in collection)
- Lazy recompute-on-read — read path doing writes is its own footgun

Happy to land as a follow-up PR (cleaner review scope) or fold into this branch if reviewers prefer the unified pattern in one place.

mihow:
**Claude says:** Landed in [9ac2ce17](https://github.com/RolnickLab/antenna/pull/1301/commits/9ac2ce17). `update_cached_counts(run_async=True)` lives on `BaseModel` and `BaseQuerySet`; backing module is `ami/base/cached_counts.py`. Detection signal handler is one line. `refresh_collection_cached_counts` task is gone; `refresh_project_cached_counts` body unchanged from this PR (kept the bulk-subquery UPDATE for SourceImage.detections_count).

Test isolation note worth flagging: the per-connection dedup set survives `TestCase` rollback because `connection` attribute state is not reset between tests, but the registered `on_commit` callbacks are discarded. Fix is to register on_commit on every `schedule_recompute` call and make `_flush_pending_recomputes` idempotent on the drain side (first call drains, subsequent calls see an empty set and no-op). That keeps the dedup benefit (10k Detection writes still collapse to one task per collection) without the rollback leak.

Reconcile loop in `ami/main/checks/cached_counts.py` switched to `update_cached_counts(run_async=False)` for stylistic unity — same behavior since reconcile is already in a worker.

Design doc + trade-offs in `docs/superpowers/specs/2026-05-14-cached-counts-update-method-design.md` (committed in 451c8be3).

mihow:
@coderabbitai @copilot+gpt-5.4 would this be enough to speed up the SourceImageCollection list view without caching the count? Or will 500k+ rows still be quite slow?

```
SourceImageCollectionViewset:
..
    def get_count(self, queryset):
        # The recent-activity orderings annotate correlated subqueries onto the
        # queryset. They don't change the row count, so strip them (and ordering)
        # before counting to keep the pagination COUNT query cheap.
        return super().get_count(queryset.order_by().values("pk"))
 ```

mihow:
**Claude says:** Recording this PR as the fix for the capture-set list DB contention now tracked in #1346. On 2026-06-20 a `/api/v2/captures/collections/` request reached p99 ~80s with the database host near 100% CPU (concurrent with job-result ingestion writing detections). This PR's denormalization is the right fix for that — the approach is sound, it just never landed.

**Coverage caveat** (found while triaging #1346, for whoever rebases): this PR denormalizes the **three base source-image counts** — `source_images_count`, `source_images_with_detections_count`, `source_images_processed_count` — which run on **every** capture-set list request, and drops their `with_*_count()` annotations from the viewset. It does **not** denormalize `with_occurrences_count` / `with_taxa_count`, which remain live `COUNT(DISTINCT)` aggregates over the deeper `images → detections → occurrence → determination` join, gated behind `?with_counts` and the detail view (`ami/main/models.py` `SourceImageCollectionQuerySet.with_occurrences_count` / `with_taxa_count`). The base counts are the always-on driver, so this PR covers the bulk of the recurring cost; but if `?with_counts` traffic is significant, that tail stays live. Worth either extending this PR to cover those two or filing a follow-up.

**Revival notes** (we'll rebase in a fresh session): the branch is currently `CONFLICTING` with `main`. The migrations need renumbering — `0085` / `0086` here collide with `0085_project_activity_sort_indexes` and `0086_sourceimage_recent_capture_index` already on `main` (main is at `0092`), so bump to ~`0093`/`0094`. Re-run the staging backfill-duration measurement before merge — the chunked `UPDATE` over a production-sized M2M is the one risky step.

mihow:
**Claude says:** Verdict from the #928 measurement pass: **land it, with a few changes.** This is the only thing that fixes the genuinely expensive case — the per-capture-set counts on the collection list. Measured on a production-snapshot dataset, that page runs ~9.5 s for a 10-collection page and ~11 s for the largest single collection (three `COUNT(DISTINCT)` over a 4-table join with on-disk merge sorts). Capping can't touch that — those are GROUP BY counts on the returned rows, not a pagination `COUNT(*)` — so a denormalized cached count is the right tool here, and the `CachedCountField` is a clean, reusable mechanism.

**Good news on the migration risk.** The backfill (`0086`) was the one un-run step worth worrying about. Measured it on a production-shape dataset inside a rolled-back transaction: the UPDATE took ~41 s but writes only ~168 rows, and it holds a `RowExclusiveLock` on the collection table plus `AccessShareLock` on the join tables — **no table-level `AccessExclusive`, reads are not blocked.** The `AccessExclusive` belongs to the `0085` ALTER, which is metadata-only and clocked at 22 ms (no table rewrite). So this is not a maintenance-window-with-downtime change — it's a "run it off-peak / not during heavy ingestion" change, because those ~168 collection rows are row-locked for the duration and concurrent writers to *those* collections would wait.

**Changes to make before merge:**

1. **Rebase + renumber the migrations.** The branch is ~36 commits behind `main`, and its `0085/0086/0087` now collide with three different migrations that have since landed on `main`. They'll need to move to the next free numbers. Heads-up: #1351 (the occurrence-list index) is also queued at `0093`, so coordinate the final numbering.
2. **Move the collection-count refresh to the very end of `save_results`,** after the null-detection sentinel `create_detections()` from the #1310/#1312 line of work. If the refresh runs before the sentinel rows are written, the processed-capture count undercounts. This is the main diff-invisible bug.
3. **Route the ML-path refresh through the async `update_cached_counts(run_async=...)`** rather than recomputing inline. It keeps the recompute off the ingestion transaction and — importantly — gives us a single refresh path to reconcile with #1288, which is trying to *remove* synchronous count refresh from the same `save_results`/`process_nats_pipeline_result` area. These two should be co-designed as one invalidation contract rather than landing one on top of the other.
4. **Add a test that exercises the `bulk_create` path.** The ML save writes detections in bulk, and `bulk_create` does not fire `post_save` signals — so a test that only uses `Detection.objects.create()` won't catch a cache that goes stale on the real write path.
5. **Treat the drift reconciler as a hard dependency, not a follow-up.** There's no staleness sentinel, so a single missed refresh shows a confidently-wrong `0` rather than degrading visibly. The periodic reconcile is what makes the denormalization safe.

Still wants confirming after the changes: before/after p99 on the capture-set list from staging APM, and the cached-vs-live read decision in (3)/#1288 (whether in-progress views read the cached column or a live annotation — that sets the staleness window).

PR #1306, opened 2026-05-14
Thumbnails API for SourceImages

loppear:
## Summary

Provide thumbnails for SourceImages, stored in `default_storage`, generated on image request and paths "cached" in new table `SourceImageThumbnail`.

### List of Changes

* Settings `THUMBNAILS` to define storage prefix and pre-defined sizes.
* New model `SourceImageThumbnail`
* New API read-only Viewset at `/api/v2/captures/thumbnails/<source_image_id>` (that only implements detail) for proxying image requests and perform thumbnailing
* * If a cached record for the requested size (requested by label, queried by dimension) does not exist, retrieves SourceImage from deployment datasource and resizes to save to default storage, returning a redirect to the media URL.
* Modifies `captures` API responses to include a `thumbnails` property with urls keyed by thumbnail size label, e.g. `.thumbnails.small`.

**Update 2026-06-10** (follow-up commits, several promoted from the #1331 proposal so this PR is deployable on its own):

* Thumbnail rows now record the requested spec width instead of the encoder's output, so Pillow's aspect-ratio rounding (239px for a 240px spec) no longer causes affected thumbnails to regenerate on every request.
* The freshness check trusts the cached row without issuing a storage existence check (an S3 HEAD request) per warm request.
* Thumbnails are encoded as progressive JPEGs (optimize pass, quality 82) so browsers can paint a coarse preview while bytes arrive.
* The thumbnail redirect carries `Cache-Control: private, max-age=300`, letting browsers reuse it across page views instead of re-paying the round trip per thumbnail.
* Rows with a blank stored path (failed or interrupted generation) regenerate instead of redirecting to the storage root.

**Also landed during review** (earlier follow-up commits, not in the original description):

* First pass of frontend wiring: the captures table and gallery use the small thumbnail, the session detail viewer and session-list example captures use the medium one, all falling back to the full-size image URL when no thumbnail is present.
* Deleting a thumbnail row (directly or via cascade from its capture) also deletes the blob from storage, so removed captures don't leave dead thumbnail files behind.
* New Django admin panel for thumbnails; the capture admin marks `size`, `last_modified`, `checksum`, and `checksum_algorithm` read-only since they are populated from the source file during sync/upload.
* Image fetching during generation now has connect/read timeouts (5s/30s) so a stalled upstream storage can't tie up web workers (`fetch_image_content` moved to `ami/utils/media.py`).

### Related Issues

Implements #236 

## Detailed Description

Thumbnails are not generated during capture requests, but delayed until frontend requests image, at which point we check if there is a valid thumbnail already or need to generate.

*(Update 2026-06-10, design context worth recording: lazy generation is a deliberate fit for how captures are accessed, not just an implementation convenience. A project can hold on the order of a million captures, but users review **occurrence crops**, not raw captures — only a few hundred captures are ever opened, typically while troubleshooting or checking the intermediate context around a detection. Generating thumbnails eagerly (at upload/sync) would spend storage and compute on images that are never viewed; generating on first request means cost scales with what people actually look at.)*

This does NOT implement front-end changes to use this `thumbnails` property of captures. Happy to include a first stab at that for the captures list gallery before this is accepted.

*(Update 2026-06-10: the first stab at the frontend is now included — captures list/gallery and session views use the new thumbnails with a fallback to the original image URL.)*

Other considerations implicated:
* Monitoring default media storage usage, as this is the first larger use of it?
* Monitoring response time and django worker availability in performing thumbnailing in-request.

### How to Test the Changes

Adds tests (`manage.py test -k thumbnail`) for the new API view and modifications to serializing captures.

## Deployment Notes

Adds database migration for new `SourceImageThumbnail` table *(now migrations `0091` and `0092` after a renumber following the merge of main)*

## Checklist

- [ x ] I have tested these changes appropriately.
- [ x ] I have added and/or modified relevant tests.
- [ x ] I updated relevant documentation or comments.
- [ x ] I have verified that this PR follows the project's coding standards.
- [ x ] Any dependent changes have already been merged to main.

## Summary by CodeRabbit

* **New Features**
  * Added thumbnail support with configurable sizes and storage prefix; capture list/detail include thumbnail URLs and a new thumbnail retrieval endpoint that redirects to stored images.
  * Admin UI: thumbnails and related source/deployment attributes are shown and filterable in the admin list view.

* **Behavior**
  * Thumbnails are cached and automatically regenerated when sources or size settings change.

* **Tests**
  * Added tests for generation, retrieval, size selection, caching, and regeneration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RolnickLab/antenna/pull/1306)


ami/main/api/serializers.py
```diff
@@ -1107,7 +1120,6 @@ class Meta:
             "event",
             "url",
             "path",
-            # "thumbnail",
```
mihow, line 1114: Funny that this `thumbnail` field has been hanging out empty all along! Glad to retire it

ami/main/api/views.py
```diff
@@ -706,6 +711,81 @@ def unstar(self, _request, pk=None) -> Response:
...
+                    img.thumbnail(new_size)
+
+                    buffer = io.BytesIO()
+                    img.save(buffer, format="JPEG")
+                    contents = buffer.getvalue()
+                    file_size = len(contents)
+
+                    # Write to storage
+                    thumbnail_key = f"{self._prefix}capture_{obj.id}/{label}.jpg"
+                    thumbnail_path = default_storage.save(thumbnail_key, buffer)
```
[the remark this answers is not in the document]
loppear: This isn't true, `getvalue` does not move the stream position.

ami/main/api/views.py
```diff
@@ -709,6 +714,77 @@ def unstar(self, _request, pk=None) -> Response:
...
+        if size is None:
+            raise api_exceptions.ValidationError(
+                detail=f"Invalid thumbnail size label provided: {label} not in {', '.join(_sizes.keys())}"
+            )
+        obj: SourceImage = self.get_object()
+        try:
+            thumb = obj.thumbnails.get(**size)
+        except SourceImageThumbnail.DoesNotExist:
+            thumb = None
+        if not thumb or thumb.last_modified < obj.last_modified or not default_storage.exists(thumb.path):
```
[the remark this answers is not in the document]
loppear: Fair.

config/settings/base.py
```diff
@@ -587,3 +587,7 @@ def _celery_result_backend_url(redis_url):
 # preserves existing behavior; deployments seeing contention can set to False
 # until the append-only ``JobLog`` child table (PR #1259) is in place.
 JOB_LOG_PERSIST_ENABLED = env.bool("JOB_LOG_PERSIST_ENABLED", default=True)  # type: ignore[no-untyped-call]
+
+
+# Sizes for Source Image Thumbnails
+THUMBNAILS = {"STORAGE_PREFIX": "thumbnails/", "SIZES": {"small": {"width": 240}, "medium": {"width": 1024}}}
```
mihow, line 593: My instinct is to namespace this, `SOURCE_IMAGE_THUMBNAILS`  or CAPTURE_THUMBNAILS, but I actually we could use these sizes for other images, like the deployment images or device type thumbnails, which use the same image thumb UI component in the list views.

mihow: Oh if you are fighting with the python formatter, we should make sure that vscode is picking up the right black settings for you, so you don't have to wait for the github tests to fail. Or are you using another IDE?

ami/main/api/views.py
```diff
@@ -709,6 +714,82 @@ def unstar(self, _request, pk=None) -> Response:
...
+        _prefix = settings.THUMBNAILS["STORAGE_PREFIX"]
+
+        label = self.request.query_params.get("label", next(iter(_sizes)))
+        size = _sizes.get(label, None)
+        if size is None:
+            raise api_exceptions.ValidationError(
+                detail=f"Invalid thumbnail size label provided: {label} not in {', '.join(_sizes.keys())}"
+            )
+        obj: SourceImage = self.get_object()
+        try:
```
mihow, line 885: I'm tempted to say we don't need the SourceImageThumbnail model and can rely on the exists() check in  the storage. But then I suppose we have to parse the height/width from the file name, and make sure we have a way to create those consistently. So I could go either way! This will be a big table eventually.

ami/main/api/views.py
```diff
@@ -709,6 +714,82 @@ def unstar(self, _request, pk=None) -> Response:
...
+            not thumb
+            or thumb.width != size["width"]
+            or thumb.last_modified < obj.last_modified
+            or not default_storage.exists(thumb.path)
+        ):
+            if obj.path and obj.deployment and obj.deployment.data_source:
+                config = obj.deployment.data_source.config
+                # Get the file
+                try:
+                    img = ami.utils.s3.read_image(config=config, key=obj.path)
```
mihow: Could we move the core thumbnailing logic to it's own home? And then import in the view? That would be helpful if we want to trigger it from a management command or a test, or another method.

ami/main/api/serializers.py
```diff
@@ -1118,6 +1130,7 @@ class Meta:
             "taxa_count",
             "detections",
             "project",
+            "thumbnails",
```
mihow: I believe this will trigger the thumbnail creation for all captures in the list view response, before responding. You proposed an alternative method in one of our chats that I liked:

- Return the thumbnail URLs in the list view before they exist, so the list view endpoint returns immediately
- The thumbnail URLs look like images (/api/captures/21/thumbnail/small/project-18-capture-21-small.jpg) etc. and they have an image mimetype BUT they are actually dynamic Django routes. 
- When the frontend is ready to load each image, each image URL gets hit, that image is generated and redirected to the media URL in the storage (or bytes served directly with some passthrough server trick)

Then each image generation happens in it's own request, and only when it's truly demanded. 

If you ignore my other comments, I think this is the most important one!

config/settings/base.py
loppear: Could be used for other thumbnail sizes, or could be namespaced when we add others, agreed.

I think the `black` struggles were because I didn't have the pre-commit setup at first, took a few rounds to go back and get all the files that would have been found in the initial changes, hopefully won't bite me going forward. (But yes, using emacs not VSCode.)

ami/main/api/views.py
loppear: Assuming files are on S3, I don't think we'll want to do filesystem name checks. Table size will be commensurate to SourceImage, could be pruned for age if that becomes an issue, though would lose the "we still have thumbnails if source files are removed" benefit. We don't join to the table just per-image lookups - should probably add a (source_image, label) index.

ami/main/api/views.py
loppear: Moving to `SourceImage.find_or_generate_thumbnail_for_label`

ami/main/api/serializers.py
loppear: Hmm. Perhaps I'm missing something about serializers, `thumbnails` is overridden here by the `SourceImageListSerializer.get_thumbnails` to work as you describe, not loading or generating thumbnails but returning the URLs to e.g. `/api/v2/captures/thumbnails/21/` which are (redirected to) image responses.

If including `thumbnails` in the `Meta.fields` is querying that relation, that's unintended and unnecessary (but still wouldn't generate thumbnails, just unneeded db query), but it seems I need this in here to get that field into the list results of the serializer.

ami/main/api/views.py
mihow: thanks!

ami/main/api/serializers.py
loppear: Yeah, thumbnails aren't in the queryset or iterated here, this is just here to include the field to use the serializer method field as noted.

ami/main/models.py
```diff
@@ -2206,6 +2207,63 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+            thumb = None
+        size = settings.THUMBNAILS["SIZES"].get(label)
+        prefix = settings.THUMBNAILS["STORAGE_PREFIX"]
+
+        if (
+            not thumb
+            or thumb.width != size["width"]
+            or thumb.last_modified < self.last_modified
+            or not default_storage.exists(thumb.path)
+        ):
```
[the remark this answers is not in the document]
loppear: Nope, `auto_now_add``.

ami/main/api/serializers.py
mihow: Ah I missed that `get_thumbnails` only calls `reverse_with_params` to return the URL routes but doesn't actually call the thumbnailer. That's great. Thanks for double checking. Exciting!

ami/main/models.py
```diff
@@ -2206,6 +2207,63 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+            height = size.get("height", None)
+            if not height:
+                height = int(orig_height * (width / float(orig_width)))
+            new_size = (width, height)
+            img.thumbnail(new_size)
+
+            buffer = BytesIO()
+            img.save(buffer, format="JPEG")
+            contents = buffer.getvalue()
+            file_size = len(contents)
```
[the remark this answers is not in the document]
loppear: @mihow My guess is this an assumption throughout (that source images do not include alpha layer PNGs) and I'm not going to address/guard that here if so.

loppear:
Now adds:
 * Basic use of thumbnails for captures in captures UI (table, gallery, playback using `medium`)
 * Thumbnail URLs to api responses for `example_captures` and `first_capture`.

Not sure why the frontend lint/tests are cancelling for this PR, but I did get them running locally to check.

mihow:
Excellent work! Everything is working in my brief testing, and the session detail view is loading way faster!

One funny quirk, I think the bounding box overlay uses the size of the original image to calculate the placement of boxes, so now they end up outside the image! Probably one good reason for us to offer relative bounding box coordinates in the API instead of absolute ones. 

image

annavik:
I pushed a small fix for the thing you pointed out @mihow . We are actually inspecting the natural size of the image after load. The reason is we had problems with the the API not always returning the size, see this old PR: https://github.com/RolnickLab/antenna/pull/459. After my fix here, we will prioritize the size returned by the API and only use the natural size as a fallback.

Since bounding boxes are currently specified in pixels, not in percent, we need the size to render them. If we can trust the API to always return a size for the image, we could simplify the FE code and skip the natural size check. We could also do this if bounding boxes were specified in percent, instead of pixels.

mihow:
> I pushed a small fix for the thing you pointed out @mihow . We are actually inspecting the natural size of the image after load. The reason is we had problems with the the API not always returning the size, see this old PR: #459. After my fix here, we will prioritize the size returned by the API and only use the natural size as a fallback.
> 
> Since bounding boxes are currently specified in pixels, not in percent, we need the size to render them. If we can trust the API to always return a size for the image, we could simplify the FE code and skip the natural size check. We could also do this if bounding boxes were specified in percent, instead of pixels.

I remembered the reason while we can't guarantee that image sizes will always be there! We only know the filename, URL, file size and MD5 hash when images are synced. We don't know the image size until the image is processed. Only then is the original image downloaded & read. But! If an image has detections, that means it has been processed. 

Right now if the image hasn't been processed, we infer the image size based on the first image in the session.

mihow:
@loppear this is the 500 error I see on arctia

image

https://arctia.dev.antenna.insectai.org/api/v2/captures/thumbnails/4820504/?label=small


`source_image.path` is the relative path to the image in the Project's storage backend, which shouldn't ever be the Django "default storage".

There are only a couple places where Antenna loads a source image, here is one. This will work no matter if the image is stored in S3 or another storage backend. This will work for both public and private s3 buckets too.

```
def load_source_image(source_image: SourceImage) -> np.ndarray:
    url = source_image.public_url(raise_errors=True)
    assert url
    image_content = fetch_image_content(url)
    image = Image.open(io.BytesIO(image_content))
    return np.array(image)
```

example usage:
```
def create_detection_images_from_source_image(source_image: SourceImage) -> list[str]:
    # Check if the source image has detections without images before loading the image
    if not source_image.detections.filter(path__isnull=True).exists():
        return []

    image_np = load_source_image(source_image)
    processed_paths = []

    for detection in source_image.detections.filter(path__isnull=True):
        if detection.bbox:
            cropped_image = crop_detection(image_np, detection.bbox)
            path = save_crop(cropped_image, detection, source_image)
            detection.path = path
            detection.save()
            processed_paths.append(path)

    return processed_paths
```

There is also `read_image`, but that is specific to S3 backends onlye
https://github.com/RolnickLab/antenna/blob/f585ddc94ff9036f46b05033d0b1821a780feadb/ami/utils/s3.py#L615-L629

ami/main/models.py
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in ecfec9bf — convert to RGB when `img.mode not in ("L", "RGB", "CMYK")` before `img.save(format="JPEG")`, and added a regression test (`test_thumbnail_non_rgb_source_converts`) that feeds an RGBA PNG through and asserts the redirect + thumbnail row. RGBA/P uploads no longer 500.

ami/main/models.py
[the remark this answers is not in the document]
mihow: **Claude says:** Re-opening to confirm: the thread was closed but the bug was real. `SourceImage.last_modified` is nullable, never set by `create_source_image_from_upload`, and after the `not thumb` short-circuit no longer holds, `thumb.last_modified < self.last_modified` raised `TypeError` 500 on every subsequent request for any uploaded capture's thumbnail. Fixed in 0c13f731:

1. `create_source_image_from_upload` now sets `last_modified=timezone.now()`, matching the parity of the S3-sync path (which copies the object's `LastModified` header). Uploaded captures now have a real source-bytes mtime, so future re-uploads will invalidate cached thumbs correctly.
2. `find_or_generate_thumbnail_for_label` guards the comparison: None on either side means "no signal of source change, trust the cached thumb" instead of raising. Covers legacy rows synced before this backfill.

Both paths covered by new tests (`test_thumbnail_source_last_modified_none_does_not_crash`, `test_upload_handler_backfills_last_modified`).

ami/main/tests.py
```diff
@@ -210,6 +218,102 @@ def test_existing_processing_service_new_project(self):
...
+    def setUp(self) -> None:
+        self.project, self.deployment = setup_test_project()
+
+        self.captures = create_captures_from_files(deployment=self.deployment)
+        self.first_capture = self.captures[0][0]
+
+        return super().setUp()
+
+    def test_thumbnail_no_list(self):
+        response = self.client.get(f"/api/v2/captures/thumbnails/")
```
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in 138cfb5b — dropped the f-prefix on both this test and on `views.py` (the `list` action's NotFound detail had the same F541). Also switched the `list` action from a fake 404 to a proper 405 with the Allow header; the route only exists for `/captures/thumbnails/<pk>/?label=...` and listing has no defined semantics.

ami/main/migrations/0091_sourceimagethumbnail_and_more.py
```diff
@@ -0,0 +1,39 @@
...
+                ("last_modified", models.DateTimeField(blank=True, null=True, auto_now_add=True)),
+                (
+                    "source_image",
+                    models.ForeignKey(
+                        null=True,
+                        on_delete=django.db.models.deletion.SET_NULL,
+                        related_name="thumbnails",
+                        to="main.sourceimage",
+                    ),
+                ),
```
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in 84f70ea2. `source_image` FK switched to `on_delete=CASCADE` (no longer nullable). New `pre_delete` signal in `ami.main.signals` does best-effort `default_storage.delete(instance.path)` so the blob doesn't outlive the row — fires on both explicit deletes and CASCADE deletes from the parent SourceImage. Migration 0090 is two-step: reap any pre-existing `source_image_id IS NULL` rows through the same signal first, then AlterField. Tests cover both the cascade and the blob cleanup.

mihow:
@loppear I pushed a few small fixes related to data integrity checks, clarification and formatting. The CI is passing now. I made one bigger change, but pushed it to a separate branch/PR to see what you think. This avoids n+1 queries and hitting the Django server and S3 API for every thumbnail URL. Which is mostly important for any of the list views: https://github.com/RolnickLab/antenna/pull/1331 

I re-deployed the arctia environment with the latest `feat/thumbnail-source-images` branch

loppear:
Cleanup looks fine to me, CASCADE and update_or_create changes seem right.

ui/src/data-services/models/capture.ts
```diff
@@ -167,6 +167,20 @@ export class Capture {
     return this._capture.event?.name ?? ''
   }
 
+  get thumbnail_small(): string {
```
mihow, line 170: seems like we are using both thumbnails.small and thumbnail_small in the UI. is there a reason not to mirror the backend and just use a `thumbnails` object (e.g. thumbnails.small)?

ami/main/models.py
```diff
@@ -2375,6 +2399,77 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+
+            # Atomic upsert: concurrent generation races on the (source_image, label)
+            # unique constraint. ``width`` stores the requested spec width, not the
+            # encoder output — PIL's rounding (e.g. 239 for a 240 spec) would otherwise
+            # fail the strict regen gate above on every request. ``height`` is informational.
+            thumb, _created = self.thumbnails.update_or_create(
+                label=label,
+                defaults={
+                    "path": thumbnail_path,
+                    "width": size["width"],
```
mihow, line 2454: hi @loppear, I went ahead with this approach for fixing the issue where some thumbs are generated with plus or minus one pixel width off from the spec. this changes the saved width on the thumbnail record to use the width from the spec instead of the actual generated width. that way the thumbnail isn't re-generated every time due to a cold hit.

mihow (approved):
Tested multiple times in the arctia environment. I like the enhancement in https://github.com/RolnickLab/antenna/pull/1331 to keep traffic off of the main Django web server. But I'm okay to deploy that as a follow up.

ui/src/data-services/models/capture.ts
mihow: **Claude says:** `thumbnail_small` / `thumbnail_medium` are model accessors (added in 36c14d6a) that read the raw `thumbnails.small` / `thumbnails.medium` API fields and fall back to `url` when the thumbnail is still cold. The raw `thumbnails` object is only touched inside those getters; call sites use the accessors. Keeping the wrapper centralizes the cold-fallback in one place — using `thumbnails.small` directly would push a `?? url` into every call site. Leaving as-is.

ami/main/models.py
loppear: Yep, that makes sense.

mihow:
**Claude says:** Flagging a bug this change introduces for **non-public projects** (private / draft / membership-restricted), found while debugging missing capture images on a restricted project.

## Symptom

Captures in restricted projects render as broken images. The captures-list request itself succeeds (it's an authenticated `fetch`, so it carries the `Authorization` header), but the individual thumbnail URLs in that response point at the new auth-gated thumbnail endpoint, and the browser loads them via ``, which is **always anonymous** — an `` tag cannot send an `Authorization` header, and the API lives on a different subdomain than the UI, so a `Lax` session cookie isn't sent on the cross-site subresource either. With no credentials, `ObjectPermission` → `SourceImage.check_permission(AnonymousUser, "retrieve")` denies the request and the endpoint returns 401.

Public projects are unaffected because anonymous read passes their permission check, so the same `` request resolves (302 → storage). The regression is specific to projects whose captures are not anonymously readable.

## Why it worked before

Previously these images were displayed via the storage presigned URL directly in ``. A presigned URL is self-authenticating — the signature in the query string is the grant, and the browser fetches storage directly, bypassing any Django permission check. The thumbnail layer routes display through a Django endpoint that re-imposes `check_permission`, which a bare `` cannot satisfy.

## Confirmed on a live deployment

| Request (anonymous, as an `` would send) | Result |
|---|---|
| `GET /api/v2/captures/thumbnails/<pk>/?label=small` — capture in a **public** project | 302 → storage blob |
| same endpoint — capture in a **restricted** project | 401 |
| `check_permission(AnonymousUser, "retrieve")` — restricted capture | `False` |
| `check_permission(AnonymousUser, "retrieve")` — public capture | `True` |

Server-side generation and blob serving are both fine — the only failure is the 401 on the auth-gated endpoint for an anonymous `` on a restricted project.

## Relevant code

- `ami/main/api/views.py` — `SourceImageThumbnailViewSet`, `permission_classes = [ObjectPermission]`; `retrieve()` redirects to `default_storage.url(thumb.path)`.
- `ami/base/permissions.py` — `ObjectPermission.has_object_permission` → `obj.check_permission(request.user, view.action)`.
- `ami/main/api/serializers.py` — `get_thumbnails()` returns the endpoint route URL for every project (public and restricted alike).
- `ami/main/models.py` — `find_or_generate_thumbnail_for_label` (lazy generation; its only caller is the view above).
- UI: thumbnails are rendered through `` (e.g. `capture.thumbnail_small/_medium`, session-details, capture gallery/columns).

## Note on the stacked PR #1331

#1331 (serve direct storage URLs for warm thumbnails) does not fully resolve this for restricted projects:

1. **Cold thumbnails can't warm through the browser.** Generation is lazy and happens only on a permitted request to the endpoint. For a restricted project the anonymous `` request that would trigger generation is the request that 401s — so the thumbnail never generates, stays cold, and #1331 keeps it on the redirect (still 401). Warming a restricted project's thumbnails would need a server-side path (backfill, or generation on import/sync) rather than relying on an anonymous browser hit.
2. **Privacy of the direct URLs.** The direct storage URLs are unsigned, non-expiring, and path-predictable (derived from capture pk + label). For restricted projects that makes warm thumbnails effectively world-readable to anyone with or guessing the URL — a looser posture than the previous presigned URLs, which expired and weren't guessable. Worth a deliberate decision before relying on direct URLs for non-public captures.

## Fix directions (for discussion)

Since `` can't authenticate via header, the options are roughly:

- **Self-authenticating thumbnail URL** — have the endpoint/serializer return a signed URL (presigned storage URL, or a short-lived signed token on the route that the view validates without a session). Most faithful to the prior presigned-URL behavior and works in a bare ``.
- **Authenticated image loads in the frontend** — `fetch` the thumbnail with the header → `blob()` → object URL (or a service worker that injects the header). Heavier change; affects every image.
- **Cookie-based auth** for the image host — only viable if the cross-site cookie situation is addressed.

Any fix should cover restricted projects without making genuinely private captures more accessible than presigned URLs already made them.

loppear:
Agreed that the behavior I would expect as a consumer of the API (and the expectation prior to this) is that the URLs for images are pre-authenticated. (Actually, didn't occur to me that cross-site auth was in play here and that the frontend is using Authorization headers not session cookies.) This is partly because the thumbnail view is within the API rest_framework router, and that only knows `TokenAuthentication`.

Beneath that thumbnails are being written to default storage (arbutus ami-media?), the underlying media isn't really under permission checks and even if the thumbnail view is per-object authenticated, there is no presigning to do for default storage? So the underlying thumbnail media will be public/crawlable after generation?

Perhaps thumbnails in fact do need to be written back to project storage for private projects then, or thumbnailing skipped for private projects until that's sorted?

Issue #1316, opened 2026-05-21
Add guidance and stats about which taxa need verification


mihow:
## Motivation

To get a meaningful project-wide picture of model accuracy and data quality, users should verify at least one occurrence of every unique taxon their pipelines have apparently observed. Today there is no surface that tells them *which* taxa still need attention or how many they have already verified — they have to drill into the occurrence list per taxon and check by hand.

This ticket adds per-taxon verification and agreement data to the existing taxa list endpoint, plus matching UI controls, so users can sort and filter to find the taxa that most need attention. Part of this year's proactive-surfacing goal; natural next step after #1296 (project summary) and #1307 (dataset-wide model agreement endpoint).

## Scope

Backend annotations on `GET /api/v2/taxa/` plus a new filter, and a new column + filter in the taxa list table. The dataset-wide `/occurrences/stats/model-agreement/` endpoint from #1307 is unchanged — it stays as the aggregate view; this ticket adds the per-taxon breakdown.

### Out of scope (queued separately)

- "Needs verification" badge / status pill on rows.
- Project-summary widget with `X of Y unique taxa verified`.
- Dedicated unverified-taxa queue page.
- Backfilling counts onto a denormalized `Taxon` field.
- Macro-average rollup (occurrence-weighted descendant sum is the only rollup in this ticket).

### Filter

- `verified=true|false` on `TaxonViewSet` — matches taxa with at least one non-withdrawn `Identification` on an occurrence whose `determination` is the taxon itself **or any descendant** (via `parents_json__contains`). Implemented as an `EXISTS` subquery, project-scoped, respects `apply_default_filters`.

### Always-on annotations (cheap)

Both reuse the existing hierarchical-match pattern (`parents_json__contains [{id: OuterRef("id")}] OR determination_id = OuterRef("id")`) used by the `taxon=<id>` filter in the occurrence list, so a Family row aggregates all its descendant species' occurrences — occurrence-weighted by construction.

- `verified_count` — count of occurrences under the taxon (incl. descendants) with at least one non-withdrawn `Identification`. Sortable. Single correlated subquery per row.
- `agreed_with_prediction_count` — count of verified occurrences whose chosen `Identification.agreed_with_prediction` is non-null. No join through `Classification` — just a non-null FK check. Different signal from `agreed_exact_count` below: measures the *agree-with-model workflow* (user clicked "agree" on a prediction), not independent-match accuracy.

### Gated annotation (heavier)

[...]

PR #1317, opened 2026-05-22
Add verification status to Taxa views

mihow:
## Summary

Adds a new filter for filtering by verification status in the taxa list view. This is the beginning of a workflow that allows identifiers to quickly verify the presence of at least one occurrence of every predicted taxa. Other filters can be combined to verify the presence in different scenarios, e.g. one occurrence per taxa, per station. 

Also adds the number of verified occurrences per taxon, which is recursive (shows all taxa verified within a genus, subfamily, etc).

Also refactors how subqueries are performed for calculating stats about taxa observations within a project. Most of the notes below are analysis and comments about the performance of that aspect. I am trying to go as far as we can before we start needing introduce a `TaxonObserved` model (proposed in https://github.com/RolnickLab/antenna/pull/490) and start caching numbers, so that figures stay fresh and we don't have to introduce more complex invalidation logic, hooks and tests.

The taxa view should be overall faster for all queries after merging this, except when filtering by a Capture Set (`SourceImageCollection`) on smaller projects. Filtering by capture sets on medium & large projects currently times out on the live site now, so still considering it a win! 

## Screenshots

image

## Performance

Measured against the deployed branch on an internal dev box copy of production data (six projects, 58k–234k occurrences each), cold queryset evaluation, cachalot flushed between every request, query count = 5 on main / 7 on this PR (two extra annotations: `verified_count`, `id__in` membership).

### Default path (`/api/v2/taxa/?project_id=<id>&limit=25&ordering=-occurrences_count`)

| project | occurrences | verified-rollup taxa | main | this PR | delta |
|---|---|---|---|---|---|
| A | 234,625 | 236 | 2.4s | **0.7s** | **−71%** |
| B | 179,468 | **707** | 2.3s | 2.6s | +13% |
| C | 144,156 | 67 | 0.5s | 0.7s | +40% |
| D | 108,370 | 188 | 0.5s | 1.0s | +100% |
| E | 84,295 | 71 | 0.7s | 1.0s | +43% |
| F | 58,590 | 261 | 0.3s | 0.5s | +67% |

Mixed. Biggest project wins big (main's per-taxon `EXISTS` membership scales with total occurrences; the PR's materialised `id__in` does not). Mid-size projects pay the verified-count precompute + CASE annotation. Default-path cost on this PR scales with **verified-rollup cardinality** (size of the CASE map) — project B's 707-taxa rollup map is 3× bigger than project A's 236, so B is slower than A despite having fewer occurrences.

### Collection path (`/api/v2/taxa/?project_id=<id>&collection=<id>&limit=25`)

| project | occurrences | main | this PR | delta |
|---|---|---|---|---|
| A | 234,625 | **TIMEOUT (>30s)** | 6.2s | **fixed** |
| B | 179,468 | **TIMEOUT (>30s)** | 12.9s | **fixed** |
| C | 144,156 | 10.7s | 6.2s | −42% |
| D | 108,370 | **TIMEOUT (>30s)** | 13.4s | **fixed** |
| E | 84,295 | **TIMEOUT (>30s)** | 14.0s | **fixed** |
| F | 58,590 | 0.8s | 6.9s | **+760% (regression)** |

PR rescues four of six projects from the 30s statement-timeout. Project F regresses hard — small project where main's correlated-subquery + EXISTS shape was already cheap, but the PR's conditional-aggregate shape pays full Taxon→Occurrence→Detection join cost regardless of selectivity. Net positive: main has 30s timeouts in production today.

Collection path on this PR is 6–14s on the four big projects. UX-wise the filter is opt-in via the capture-set dropdown, so the slow path doesn't hit the default taxa view. Tracking under a follow-up perf issue (size-based hybrid dispatch, or push verified-count out of the aggregate).

### Backend

- **`ami/main/models.py` — new `TaxonQuerySet` helpers.** Count logic moved off the viewset:
  - `with_observation_counts_subqueries(project, request, ...)` — correlated `Subquery` for `occurrences_count`, `best_determination_score`, `last_detected`. Index-served by the composite `(determination_id, project_id, event_id, determination_score)` index on Occurrence. Used by all non-collection paths.
  - `with_observation_counts_aggregated(project, request, ...)` — conditional `Count(distinct=True)` / `Max(...)` on the Taxon→occurrences reverse relation, one `GROUP BY`. Used by the `?collection=` path.
  - `observed_in_project_subqueries(project, request, ...)` — `id__in` materialised membership. Replaces main's per-taxon `EXISTS`.
  - `with_verification_counts(project, request, ...)` — Python pass over verified occurrences building `{taxon_id: count}` (including ancestors via `parents_json`), applied as a sparse `CASE` annotation. Filter resolves from the same map via `id__in`.
- **`ami/main/api/views.py` — `TaxonViewSet.get_taxa_observed` is now an orchestrator.** Dispatches between `with_observation_counts_subqueries` and `with_observation_counts_aggregated` based on `"collection" in request.query_params`. `TaxonCollectionFilter` filter backend removed — redundant JOIN was a planner blow-up; the collection filter is enforced inside the aggregate now.
- **`verified_count` added to `ordering_fields`.** `verified=true|false` registered on the filterset.
- **Migration `0087_taxon_parents_json_gin_index`** — `CREATE INDEX CONCURRENTLY IF NOT EXISTS` on `Taxon.parents_json` (`jsonb_path_ops`). Serves literal-RHS containment (the occurrence-list `?taxon=<id>` filter, project default-taxa filter). Does **not** serve the rollup — see "Why not `OuterRef`" below.

### Frontend

- Sortable **Verified** column on the taxa list, linking to that taxon's verified occurrences in the occurrence-list.
- **Verification status** filter pill (All / Verified / Unverified) → `verified=` query param. Reuses the existing occurrence-list filter component.
- **Verification** panel on the taxon detail page showing the verified count.

## Why not a per-taxon `OuterRef` subquery for the rollup

First implementation in this branch used `parents_json @> [{"id": OuterRef("pk")}]` to walk the ancestor chain per row. With the GIN `jsonb_path_ops` index in place this still timed out at 30s on a 1k-taxa / 17k-occurrence project — the GIN containment index only serves a **constant** right-hand side. An `OuterRef` RHS degrades to a per-row scan, the index is not used. The timeout hit even with the Verified column hidden, because the annotation is always-on and the `verified=false` `COUNT` evaluates the subquery for every taxon in the project.

The single-pass precompute + sparse CASE map is what unblocks it. The rollup runs over a sparse subset (verified occurrences are bounded by human review effort) and the `CASE` annotation is select-only — stripped from `COUNT` queries by the ORM.

The GIN index still ships because it serves the pre-existing literal-RHS consumers (the occurrence-list `?taxon=<id>` filter and the project default-taxa filter via `build_occurrence_default_filters_q`). Not load-bearing for this PR's data path, but it closes the #1307 follow-up.

## Count semantics

- `occurrences_count` — direct match (unchanged).
- `verified_count` — rolls up descendants. For species rows the two are equivalent. For Family / Order rows `verified_count` can exceed the direct `occurrences_count`.

## Tests

`ami.main.tests.TestTaxaVerification` covers:
- Hierarchical rollup to ancestors (species verifies → genus / family / order rows all increment).
- `verified=true|false` filter, strict-complement property, `apply_defaults` interaction.
- `ordering=verified_count`.
- Detection fan-out dedup under `?collection=<id>` — regression for the rollup `.values()` query: a verified occurrence with multiple detections must not double-count.

Full backend taxa test suite (`TestTaxaVerification`, `TestTaxonomyViews`, `TestTaxonListQueryCount`, `TestProjectDefaultTaxaFilter`, `TestCachedCountsDefaultFilters`) green, 40/40.

## Dev-environment caveat

Projects with a verified-rollup map of ~700+ distinct taxa (like project B in the table above) trip `sqlparse`'s 10,000-token limit inside `django-debug-toolbar`'s `reformat_sql` step. The underlying query has already succeeded — the toolbar's pretty-printer chokes formatting the SQL for the panel and returns a 500 after-the-fact. Production doesn't load the toolbar; endpoint serves normally. Worth a trivial follow-up to wrap `reformat_sql` in try/except inside `config/settings/local.py`. Unrelated to the data path.

## Reference

Pattern, gotchas, and dispatch matrix captured in `docs/claude/reference/hierarchical-rollup-query-performance.md` for future rollup work on this endpoint or others.

## Test plan

- [x] Backend tests pass in CI (full taxa suite, 40/40).
- [x] Performance: main vs this PR benched against six 58k–234k-occurrence project copies on internal dev box (tables above).
- [x] Browser check of the column / filter pill / detail panel on the deployed branch.
- [ ] Optional follow-up: address the small-project collection-path regression (project F: 0.8s → 6.9s) — likely a size-based dispatch heuristic or pushing verified-count out of the aggregate.

## Deferred to follow-up

- **#1319** — model-agreement metrics (`agreed_with_prediction_count`, `agreed_exact_count`, `with_agreement`). Originally bundled here, deferred because no FE consumer at merge time.
- **#1320** — verification workflow from the Taxa view (single-click identification modal per predicted taxon, plus optional links from the `last_seen` / `best_score` cells to their source occurrences).
- Collection-path size-based dispatch heuristic to address the small-project regression in the table above.
- Wrap `debug_toolbar.panels.sql.utils.reformat_sql` in try/except for dev-box stability with large CASE maps.

## Release Notes

* **New Features**
  * Species table displays a new sortable "Verified" column showing verification counts per species.
  * Species details page includes a "Verification" section with verification metrics and links to verified occurrences.
  * Species can now be filtered by verification status.

* **Chores**
  * Database performance improvements for verification queries.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RolnickLab/antenna/pull/1317?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ami/main/api/views.py
```diff
@@ -1654,6 +1670,113 @@ def get_taxa_observed(
             # Efficient EXISTS check that uses the composite index
             qs = qs.filter(models.Exists(Occurrence.objects.filter(base_filter)))
 
+        qs = self.add_verification_data(qs, occurrence_filters, default_filters_q)
+
```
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in commit 16b14686. The always-on annotations no longer use correlated `parents_json` subqueries — the rollup is computed in a single Python pass over the sparse *verified*-occurrence set (those with a non-withdrawn Identification) and applied as constant-time `CASE` annotations. On a ~1k-taxa / ~17k-occurrence project the default list went from a 30s statement timeout to ~0.3s. See the PR description (Performance) and `docs/claude/reference/hierarchical-rollup-query-performance.md`.

ami/main/api/views.py
```diff
@@ -1428,6 +1443,7 @@ class TaxonViewSet(DefaultViewSet, ProjectMixin):
         "created_at",
         "updated_at",
         "occurrences_count",
+        "verified_count",
```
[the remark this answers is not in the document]
mihow: **Claude says:** Addressed in 16b14686. `verified_count` is now a constant `CASE` annotation built from a precomputed map, so `ordering=verified_count` no longer evaluates a per-taxon subquery before the LIMIT — measured ~0.24s on the same large project. Kept in `ordering_fields`.

ami/main/api/views.py
```diff
@@ -1654,6 +1657,114 @@ def get_taxa_observed(
...
+            Identification.objects.filter(occurrence=models.OuterRef("pk"), withdrawn=False)
+            .order_by(*BEST_IDENTIFICATION_ORDER)
+            .values("agreed_with_prediction_id")[:1]
+        )
+        verified_occurrences = (
+            Occurrence.objects.filter(occurrence_filters)
+            .filter(default_filters_q)
+            .filter(models.Exists(Identification.objects.filter(occurrence=models.OuterRef("pk"), withdrawn=False)))
+            .annotate(_agreed_prediction_id=best_identification_agreed_prediction)
+        )
```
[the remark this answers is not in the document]
mihow: **Claude says:** Good catch — fixed in commit 10c72cbd. When `?collection=<id>` joins Occurrence to detections, a single occurrence yielded one row per matching detection and inflated the counts. Added `pk` to the `values()` projection and `.distinct()` so each occurrence is counted once, plus a regression test (`test_verified_count_not_inflated_by_collection_join`) using a verified occurrence with two detections in the collection.

ami/main/api/serializers.py
```diff
@@ -588,13 +588,27 @@ def get_taxa(self, obj):
         return [{"id": taxon.id, "name": taxon.name} for taxon in obj.taxa.all()]
 
 
+def agreement_requested(request: Request | None) -> bool:
+    """Whether ``with_agreement=true`` is set, gating the heavier agreed_exact_count."""
+    if request is None:
+        return False
+    value = request.query_params.get("with_agreement", "")
```
mihow: Use our "single param field" for this or boolean field rather than recreating this

ami/main/api/serializers.py
```diff
@@ -886,6 +903,9 @@ class Meta:
             "parents",
             "details",
             "occurrences_count",
+            "verified_count",
+            "agreed_with_prediction_count",
```
mihow: perhaps `model_agreed_*` would be more clear than than `agreed_*` to distinguish from human verifications.

ami/main/api/views.py
```diff
@@ -1634,74 +1645,199 @@ def get_taxa_observed(
...
 
-        # Combine base occurrence filters with default filters
-        base_filter = models.Q(
-            occurrence_filters,
-            determination_id=models.OuterRef("id"),
+    def _include_agreement(self) -> bool:
+        """Whether the heavier ``agreed_exact_count`` annotation should be computed."""
+        if self.action == "retrieve":
+            return True
+        return bool(BooleanField(required=False).clean(self.request.query_params.get("with_agreement")))
```
mihow: Repetition of `agreement_requested`?

ami/main/api/serializers.py
mihow: **Claude says:** Moot after the trim in 20683ee2 — the entire `agreement_requested` helper plus the `__init__` field-pop logic was removed from `TaxonListSerializer`. The PR now ships only the verified count (always-on) and the verified filter (parsed once via `BooleanField().clean()` inside the view). Model-agreement counts and their gating are deferred to #1319, where the boolean parser will be wired through serializer context (single source of truth) per your guidance.

ami/main/api/serializers.py
mihow: **Claude says:** Agreed — and trimmed entirely from this PR (commit 20683ee2). `agreed_with_prediction_count` and `agreed_exact_count` are deferred to #1319, where the rename to `model_agreed_*` is captured in the implementation guidance. Keeps this PR scoped to issue #1316 (human verification UX) and isolates the model-evaluation surface for when a FE consumer actually lands.

ami/main/api/views.py
mihow: **Claude says:** Removed in 20683ee2 — both `_include_agreement` on the view and `agreement_requested` in the serializer are gone, since the model-agreement annotation is no longer computed in this PR. Tracking the single-parser refactor in #1319's implementation notes so it doesn't get reintroduced when the counts come back.

mihow (commented):
Tested in dev environment with large projects with many taxa & verifications.

Issue #1321, opened 2026-05-27
Preparing jobs takes too long preparing large collections


mihow:
## Summary

On a production job against a large `SourceImageCollection` (~147k images), the Collect stage appears to hang inside `filter_processed_images()` (`ami/ml/models/pipeline.py:60`) for long enough that the celery worker stops sending AMQP heartbeats and the job reaper later flips the job from `STARTED` to `REVOKED`. The worker itself does not crash and memory stays flat — the job is making forward progress, just at a rate that exceeds both the broker heartbeat window and the reaper's stall threshold.

Most jobs against smaller collections complete fine, so this is collection-size–dependent rather than a general regression. Filing because the failure mode is silent (no error, no Collect-stage progress, no NATS stream ever created) and the job-side symptoms point in misleading directions.

## Observed behaviour (rough timeline from a single occurrence)

A production job on a ~147k-image collection, a regional-classifier pipeline, `dispatch_mode=async_api`:

```
T+0:00   Job STARTED
T+0:07   JobLog: "Filtering images that have already been processed by pipeline #N ..."
                 ← last application log line for ~19 minutes
T+17:35  RabbitMQ closes the celeryworker's AMQP connection
                 (broker log: "missed heartbeats from client, timeout: 30s")
T+19:18  Reaper fires: "no progress for 19.3 min (threshold 10 min).
                       celery_state=FAILURE, dispatch_mode=async_api,
                       stages: collect=0.0% CREATED, process=0.0% CREATED, results=0.0% CREATED"
                 → status STARTED → REVOKED
T+19:18  NATS consumer/stream deletion logs a 404 (stream was never created)
```

Application-host memory was flat at ~13% (~4 GB) for the whole window. No `TransactionError`. No OOM kill in `dmesg`. ML workers polled `/next` at T+0:00–0:03 and correctly got "non-active job (status=PENDING); returning empty" — the worker pool was idle and healthy.

So the only thing the worker was doing during those ~19 minutes was running inside `filter_processed_images()`.

## Inferred cause (code reading, not measurement)

`filter_processed_images()` at `ami/ml/models/pipeline.py:60` iterates one image at a time and issues per-image ORM lookups against `Detection` / `Classification`:

- `existing_detections = image.detections.filter(detection_algorithm__in=pipeline_algorithms)`
- `existing_detections.exists()`
- `existing_detections.exclude(NULL_DETECTIONS_FILTER).exists()`
[...]

Issue #1282, opened 2026-04-30
CANCELLED jobs leak through /next filter, starve newer async_api jobs


mihow:
## Bug

`/next` (claim-next-job, currently `list(ids_only=1)` in `ami/jobs/views.py:296 get_queryset()`) hands out **CANCELLED** async_api jobs forever. The exclude filter does:

```python
jobs.exclude(
    status=JobState.failed_states(),
    updated_at__lt=cutoff_datetime,
)
```

Two problems:

1. **`JobState.failed_states()` does not include `CANCELLED`** — only `[FAILURE, REVOKED, UNKNOWN]`. CANCELLED jobs are treated as still-active and offered to workers indefinitely.
2. **`status=` is the wrong lookup** for a list-valued return — should be `status__in=`. As written, Django compares the column to the list as a single value; the exclude effectively never matches.

The two combine: any user-cancelled async_api job becomes a permanent zombie that workers serially "claim" every poll, then immediately bail because the corresponding NATS stream is empty/exhausted. Newer async_api jobs created after the zombies are starved out — workers never see them.

## Symptom

A real async_api job sits with `collect` stage = `SUCCESS`, `process` stage = `CREATED 0%` indefinitely. NATS stream for that job ID has `num_pending = total_tasks`, `delivered = 0`. Worker stdout cycles:

```
Processing job <old_cancelled_id> with pipeline ...
DataLoader subprocess 0/1 starting iteration for job <old_cancelled_id>
No jobs found, sleeping for 5 seconds
```

…on the same handful of stale CANCELLED job IDs every 5s, never picking up the new job.

## Manual unblock

Bump CANCELLED zombies to REVOKED so they fall in `failed_states()`. `.update()` bypasses `auto_now`, so `updated_at` stays at the original old value (already past the 72h `FAILED_JOBS_DISPLAY_MAX_HOURS` cutoff):

```python
from django.utils import timezone
Job.objects.filter(pk__in=[<zombie ids>]).update(
    status="REVOKED", finished_at=timezone.now()
)
```

Workers stop cycling within ~30s.

## Suggested fix

Either:

- Add `CANCELLED` to `JobState.failed_states()`, **and** change `status=` to `status__in=` in the `exclude` so the comparison actually works.
- Or short-circuit in the `/next` handler with an explicit `exclude(status__in=[CANCELLED, FAILURE, REVOKED, UNKNOWN])`.

## Links / urgency

[...]

PR #1324, opened 2026-05-27
Improve celery task dispatch and cancellation to prevent stuck jobs

### LINKED ISSUE #1323: Stuck job blocks other jobs even when workers are free
## Summary

A long-running `run_job` celery task on the `jobs` queue can block other queued `run_job` tasks even when sibling worker processes (and a sibling worker host) are idle. The blocking job appears stuck mid-execution; the queued jobs sit in `RESERVED` state on the same worker as the blocker rather than getting dispatched to free children or to the other worker container.

This was visible during a recent incident where a `run_job` task was running inside `filter_processed_images()` against a very large collection (separately tracked in #1321). While that task held one prefork child on one of the two `jobs`-queue containers, a second `run_job` was queued and ended up `RESERVED` against the same container — even though 15 of 16 children on that container were idle and the entire other container was idle. Manually `SIGKILL`'ing the stuck task immediately let the queued job start.

Two reinforcing causes underneath that shape, both fixable with small config changes.

## Observed (from `celery inspect`)

Inspecting the two `celeryworker_jobs` containers while the symptom was reproducing:

```
celery@A (queues=['jobs'])  max-concurrency=16  prefetch=16
  ACTIVE:   1   (the stuck run_job, sitting inside a slow DB loop)
  RESERVED: 1   (the freshly-queued run_job, never starts)

celery@B (queues=['jobs'])  max-concurrency=16  prefetch=16
  ACTIVE:   0
  RESERVED: 0
```

Symptom: the second `run_job` sat `RESERVED` on container A for ~9 minutes until I `SIGKILL`'d the blocker; it then transitioned to `ACTIVE` on the same container in the same second. The other 15 children on A and the entirety of B were never offered the task by the broker or by A's master process.

`SystemSample` memory on the worker host was flat throughout. Other queues (`antenna`, `ml_results`) were healthy. Nothing in `dmesg`. So the worker was alive and pulling control messages — it just couldn't shed the queued task.

## Inferred cause

I want to be explicit that this is from reading the config plus the symptom, not from instrumenting celery internals.

### 1. Default celery prefork pre-assignment

The celery worker start command (`compose/production/django/celery/worker/start`) does not pass `-O fair` (or set `worker_pool_optimization = 'fair'` in settings). Under the default scheduler, the master process **pre-assigns** prefetched messages to specific prefork children at delivery time. Once message M is pre-assigned to child C, M waits on C even if other children become free first.

With `-O fair`, the master holds prefetched messages in a shared buffer and hands one to a child only when that child is genuinely idle. Long, heterogeneous-duration tasks (which `run_job` very much is) is the exact use case the fair scheduler is recommended for.

### 2. No `acks_late` on `run_job`

`ami/jobs/tasks.py:139` decorates `run_job` with:

```python
@celery_app.task(bind=True, soft_time_limit=default_soft_time_limit, time_limit=default_time_limit)
def run_job(self, job_id: int) -> None:
    ...
```

No `acks_late=True`, no `reject_on_worker_lost=True`. With the celery default (early ack), the broker considers the message delivered the moment the worker receives it. If the worker process is `SIGKILL`'d (e.g. cancel path, OOM, supervisor restart) the message is gone — the broker will not redeliver it. The job row is left in `STARTED`, no celery task is running, no NATS stream was ever created.

This is consistent with what we observed on the prior `STARTED → REVOKED` reaper path: the celery task disappeared silently and the job sat in `STARTED` until the reaper noticed.

### 3. Cross-container distribution

Even after fixing (1), the broker still distributes messages by consumer-prefetch availability. With `prefetch=16` per container, a container with 1 unacked task still has 15 slots of headroom, so newly queued tasks can still pile onto it instead of going to a 100%-idle sibling. That's separate from the within-container head-of-line block but worth flagging — see direction (C) below.

## Possible directions

Ordered by effort/risk; these are options to discuss, not a chosen plan.

- **(A) Add `-O fair` to the worker start command.** One line in `compose/production/django/celery/worker/start`. Eliminates within-container head-of-line blocking. Safe to apply across all queues, but the value is mostly on the `jobs` queue where individual tasks are long. *What to verify before applying:* run a local reproducer (queue two long `run_job`s back-to-back, sleep one, confirm the other runs in parallel under `-O fair` and queues behind under default).

- **(B) Add `acks_late=True, reject_on_worker_lost=True` to `run_job`.** Broker redelivers if the worker dies mid-task. *What to verify before applying:* confirm `run_job` is safe to redeliver — currently the body at `ami/jobs/tasks.py:140–164` does `job.run()` without an explicit status gate. If a `STARTED` job gets a redelivered `run_job`, will it start over cleanly, no-op, or double-up on side effects? This is a correctness check, not a performance question.

- **(C) Lower `CELERY_WORKER_CONCURRENCY` for the `jobs` queue specifically.** Today it inherits the same `CELERY_WORKER_CONCURRENCY=16` as the `antenna` and `ml_results` queues, which gives `prefetch=16` per `jobs`-queue container. Dropping the `jobs`-queue containers to e.g. concurrency=4 (with the antenna and ml queues unchanged) would shrink the per-container reservation window. The broker would more aggressively spill to the sibling container when the first one has unacked tasks. Tradeoff is lower theoretical max-concurrent run_jobs per host, which is fine in practice because run_jobs spend most of their time waiting on NATS rather than CPU.

- **(D) Push the slow Collect work out of `run_job` itself** — covered by #1321. If `run_job` finishes quickly and hands off to a sub-task, none of the above matters much; the slot frees fast either way.

(A) + (B) are the immediate wins and don't conflict with (D) landing later.

1. Re-running the symptom against a real `-O fair` worker (in staging or dev) to confirm `RESERVED → ACTIVE` happens immediately when a sibling child is idle.
2. Whether `run_job` is safe to redeliver under `acks_late=True`. Specifically, what happens when a redelivered `run_job` finds the job already in `STARTED` (because the previous attempt got partway). Code reading suggests `job.run()` will do whatever it does without checking — worth confirming in a test.
3. Whether `-O fair` interacts poorly with `max-tasks-per-child=100`. The fair scheduler can leave more idle children, but the recycle-per-child counter should still tick down normally.

## Related

- #1321 — `filter_processed_images` is the upstream slowness that exposed this scheduling problem. Fixing #1321 reduces the *frequency* of long `run_job` tasks; fixing this issue reduces the *blast radius* when they do happen.

mihow:
Closes #1323.

Last week we hit the symptom Issue #1323 describes: a `run_job` was sitting inside `filter_processed_images()` for ~9 minutes against a huge collection, and a fresh `run_job` queued behind it sat in `RESERVED` on the same worker container the entire time — even though 15 of 16 children on that container were idle and the entire sibling container was idle. `SIGKILL`'ing the blocker let the queued job start in the same second. The job that finally ran was almost certainly fine, but it wasted nine minutes of wall clock for no good reason, and any user clicking "run" during that window saw nothing happen.

This PR fixes the three reinforcing causes the issue identified, plus an orthogonal cancel-path bug I noticed while reading the code paths. Each one is a small config or decorator change; the value is in stacking them.

## What changes for users

- **Stuck `run_job` no longer blocks other jobs in the queue.** A slow first job releases the worker slot for the next one as soon as a sibling child is idle, rather than holding 15 idle slots hostage.
- **A worker crash mid-job no longer silently drops the job.** Before: the celery message was already acked at delivery, so a SIGKILL/OOM/deploy roll meant the job stayed in `STARTED` forever until the reaper found it. After: broker holds the message, redelivers it when a worker comes back. The job either resumes or — if it had already settled — exits cleanly.
- **Cancelling an async ML job actually cancels.** Today's `Job.cancel` calls `revoke(terminate=True)` on the local `run_job` task. For ASYNC_API that task has almost always already finished (`queue_images_to_nats` returns fast) — terminating it does nothing about the actual work running on the remote ADC worker, and on the rare occasion the bootstrap is still running, the SIGTERM kills it without redelivery. The real cancel mechanism for async is tearing down the NATS stream + Redis state, which `cleanup_async_job_if_needed` already does. We now skip terminate for ASYNC_API.

## Plain-language summary

| Behavior | Before | After |
|---|---|---|
| Long `run_job` blocking sibling jobs on same container | Yes — pre-assignment pins messages to busy child | No — fair scheduler hands messages only to idle children |
| Long `run_job` blocking sibling jobs on sibling container | Sometimes — broker spills late | Reduced — pairs naturally with smaller per-container reservation window (see below) |
| Worker SIGKILL/OOM mid-`run_job` | Message lost; job stranded in STARTED | Broker redelivers; early-guard handles redelivery cleanly |
| Cancel of ASYNC_API job | Terminates local bootstrap (mostly a no-op, occasionally a SIGTERM mid-flight); remote workers keep going | Tears down NATS stream + Redis state; remote work stops naturally |
| Cancel of SYNC_API / INTERNAL job | Terminates celery task | Unchanged — terminate is still the only way |

## What's in this PR

1. **`config/settings/base.py`** — `CELERY_WORKER_POOL_OPTIMIZATION = "fair"`. One line. Applies to all queues; the value is largely on `jobs`.
2. **`ami/jobs/tasks.py`** — `acks_late=True, reject_on_worker_lost=True` on `run_job`, plus an early-guard at the top of the task body that returns cleanly when `job.status` is in `final_states()` or `CANCELING`. The guard is what makes redelivery and cancel-race safe.
3. **`ami/jobs/models.py`** — `Job.cancel` no longer passes `terminate=True` when `dispatch_mode == ASYNC_API`. For other dispatch modes, behavior is unchanged.
4. **`docker-compose.worker.yml`** — added a commented-out `CELERY_WORKER_CONCURRENCY: "4"` on `celeryworker_jobs` with a TODO referencing this issue, in case we want to enable cause C later.

## A note on the counter-intuitive concurrency knob

The issue suggests *lowering* per-container concurrency on the jobs queue as a third fix. I've left this commented out for now (just a discoverability hint in `docker-compose.worker.yml`) because it reads as backwards and I'd rather watch the first two fixes in production before pulling this lever too.

The reasoning, briefly: celery's prefetch reserves `concurrency × prefetch_multiplier(=1)` messages per container at the broker level. With concurrency=16, that's 16 messages held in the container's local buffer. When one of those messages is a stuck task, the container still tells the broker "I have 15 free slots" — and the broker keeps offering new messages to that container instead of spilling to a fully-idle sibling. Lowering concurrency to 4 shrinks the reservation window so the broker spills sooner.

The reason this isn't a meaningful capacity cut: `run_job` spends nearly all its time waiting on NATS results to come back, not burning CPU. The 16 was originally raised (#1228) for `ml_results` and `antenna`, which are DB/Redis-bound and benefit from oversubscription. The `jobs` queue inherited the high number incidentally.

## What the three fixes are actually doing

There are three different head-of-line problems, with three different mechanisms:

- **Inside a child** (one slow task blocks its own pipe): can't be fixed, that's just how prefork works.
- **Inside a container, across children** (default pre-assignment scheduler pins messages to specific child pipes regardless of which child becomes idle first): fixed by `-O fair`.
- **Across containers** (broker prefetch reservation: a 1-task container still looks like it has 15 free slots): mitigated by lower per-container concurrency (the deferred knob).

`acks_late` is orthogonal to all three — it's about surviving worker death, not about scheduling. But it's a precondition for both the cancel fix to be safe (cancellation can now terminate a worker without losing the redelivered message that the early-guard will short-circuit) and for the deferred concurrency change to be safe (smaller pools mean each child handles more tasks, and a single SIGKILL hurts more).

## What's verified

- New unit tests on `Job.cancel` for ASYNC_API / SYNC_API / no-task-id paths.
- New tests on the `run_job` early-guard covering `REVOKED`, `CANCELING`, `SUCCESS`, and the contract pair (`PENDING` still runs).
- Full `ami.jobs.tests` (118 tests) and `ami.ml.tests` + `ami.ml.orchestration.tests` (81 tests) green.

## What still needs verifying in staging/prod

From the issue's "what we still need to verify" section, two of three are now testable:

1. **Re-running the symptom against a real `-O fair` worker** — confirm `RESERVED → ACTIVE` happens immediately when a sibling child is idle. I'd want to do this on dev box (queue two long `run_job`s back-to-back, sleep one).
2. **`-O fair` interaction with `max-tasks-per-child=100`** — should be fine but worth watching for the first day after deploy.
3. **`acks_late` redelivery in practice** — the early-guard makes a redelivered `run_job` a no-op when the job is already settled, so this is covered by the tests, but worth eyeballing the celery logs after deploy for unexpected `Skipping run_job` messages.

The cancel fix is the one I'd most like a second pair of eyes on — the docstring is the long version, but the gist is "for ASYNC_API the celery task isn't where the work is, so terminating it doesn't cancel; cleanup does."

## Related

- #1323 — this issue.
- #1321 — `filter_processed_images` slowness, the upstream cause of the long `run_job` that exposed this. Fixing #1321 reduces the *frequency* of stuck jobs; this PR reduces the *blast radius* when they happen.
- #1228 — the PR that raised `CELERY_WORKER_CONCURRENCY` to 16 for ml/antenna queues. Context for why the jobs queue inherited the same number.
- #1234 / #1235 — the prior async-job fix pair (ACK/SREM ordering, Bug C guard, stale-job cutoff). The cancel fix here is in similar territory.

Co-Authored-By: Claude <noreply@anthropic.com>

## Summary by CodeRabbit

* **Bug Fixes**
  * Job cancellation now handles different job types more reliably, ensuring consistent state transitions and proper cleanup.
  * Task execution is more resilient when workers are unexpectedly lost, with automatic re-delivery of in-flight tasks to maintain reliability.

* **Chores**
  * Optimized worker pool scheduling configuration to enable fair task distribution and reduce processing delays.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RolnickLab/antenna/pull/1324?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ami/jobs/models.py
+++ b/ami/jobs/models.py
@@ -1067,25 +1067,39 @@ class Job(BaseModel):
 
     def cancel(self):
         """
-        Cancel a job. For async_api jobs, clean up NATS/Redis resources
-        and transition through CANCELING → REVOKED. For other jobs,
-        revoke the Celery task.
+        Cancel a job.
+
+        For ASYNC_API jobs the long-running work is on remote ADC workers via
+        NATS, not in the local ``run_job`` celery task — by the time the user
+        clicks cancel, ``run_job`` has usually already finished
+        ``queue_images_to_nats`` and returned. Tearing down the NATS stream +
+        Redis state (``cleanup_async_job_if_needed``) is what actually stops
+        further work: ADC stops being delivered tasks, and any in-flight
+        result handlers see no Redis state and fast-fail. Calling
+        ``revoke(terminate=True)`` on the (likely-done) run_job would SIGTERM
+        the worker child if it happens to still be inside the bootstrap (e.g.
+        a slow ``filter_processed_images`` for a huge collection), which
+        prior to ``acks_late`` was an unrecoverable message loss. We revoke
+        without terminate so a not-yet-started copy is dropped without
+        killing in-flight bootstrap; the in-flight copy then notices
+        ``status == CANCELING`` via the early-guard in ``run_job`` next time
+        it's invoked (e.g. on redelivery) and bails out cleanly.
+
+        For INTERNAL / SYNC_API jobs the celery task body owns the entire
+        job lifecycle, so terminating it remains the only way to stop
+        active work.
         """
         self.status = JobState.CANCELING
         self.save()
 
+        is_async_api = self.dispatch_mode == JobDispatchMode.ASYNC_API
         if self.task_id:
             task = run_job.AsyncResult(self.task_id)
             if task:
-                task.revoke(terminate=True)
-            if self.dispatch_mode == JobDispatchMode.ASYNC_API:
-                # For async jobs we need to set the status to revoked here since the task already
-                # finished (it only queues the images).
-                self.status = JobState.REVOKED
-                self.save()
-        else:
-            self.status = JobState.REVOKED
-            self.save()
+                task.revoke(terminate=not is_async_api)
+
+        self.status = JobState.REVOKED
+        self.save()
 
         cleanup_async_job_if_needed(self)
 
--- a/ami/jobs/models.py
+++ b/ami/jobs/models.py
@@ -529,6 +529,28 @@ class MLJob(JobType):
             progress=1,
         )
 
+        # Mid-bootstrap cancel guard. ``collect_images`` above can run for many
+        # minutes on large collections (S3 list + DB joins), and the user may
+        # cancel during that window. ``Job.cancel()`` for ASYNC_API does
+        # ``revoke(terminate=False)`` to avoid SIGKILL'ing this worker, then
+        # writes REVOKED + tears down the NATS stream / Redis state. Without
+        # this check we would (a) clobber the cancel's REVOKED via the next
+        # full ``job.save()`` and (b) proceed to ``queue_images_to_nats``,
+        # recreating the stream the cancel just deleted and dispatching real
+        # GPU work to ADC for a revoked job. Refresh is read-only against the
+        # ``status`` column; the in-memory ``progress`` mutations from the
+        # collect stage are intentionally dropped on the bail path because the
+        # job is settled — no further progress writes make sense. Covers
+        # ASYNC_API (NATS dispatch) and SYNC paths (Celery sub-tasks in
+        # ``process_images``); INTERNAL jobs benefit too. See
+        # RolnickLab/antenna#1323.
+        db_status = Job.objects.values_list("status", flat=True).get(pk=job.pk)
+        if db_status in JobState.final_states() or db_status == JobState.CANCELING:
+            job.logger.info(
+                f"Job {job.pk} settled to {db_status} during bootstrap; " f"skipping dispatch of {len(images)} images"
+            )
+            return
+
         # End image collection stage
         job.save()
 
@@ -1040,6 +1062,21 @@ class Job(BaseModel):
 
         if save:
             self.save()
+
+    def is_settled(self) -> bool:
+        """Return True when the job is in a terminal state or being cancelled.
+
+        Used by every code path that must not start (or continue) work for a
+        job whose lifecycle has effectively ended: the ``run_job`` early-guard
+        (after acks_late redelivery), the ``MLJob.run`` mid-bootstrap cancel
+        check (before ``queue_images_to_nats`` dispatches GPU work to ADC),
+        the prerun signal handler (so a redelivered or canceled message does
+        not get its status reset to PENDING), and ``_fail_job``. Centralized
+        so the predicate stays in one place — adding a new "do not resume"
+        status only requires touching :meth:`JobState.final_states` or this
+        method.
+        """
+        return self.status in JobState.final_states() or self.status == JobState.CANCELING
 
     def run(self):
         """
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -150,7 +150,7 @@ def update_async_services_seen_for_project(project_id: int) -> None:
     reject_on_worker_lost=True,
 )
 def run_job(self, job_id: int) -> None:
-    from ami.jobs.models import Job, JobState
+    from ami.jobs.models import Job
 
     try:
         job = Job.objects.get(pk=job_id)
@@ -161,8 +161,10 @@ def run_job(self, job_id: int) -> None:
     # Early-guard: under acks_late, the broker may redeliver this message after a
     # worker SIGKILL/OOM, and Job.cancel() may also flip status to CANCELING /
     # REVOKED while the message sits in the prefetch buffer. Don't re-run a job
-    # that's already settled or being torn down.
-    if job.status in JobState.final_states() or job.status == JobState.CANCELING:
+    # that's already settled or being torn down. The companion guard in
+    # pre_update_job_status above prevents the task_prerun signal from
+    # overwriting that status with PENDING before we get here.
+    if job.is_settled():
         job.logger.info(
             f"Skipping run_job for job {job.pk}: already in status {job.status} "
             f"(redelivery or cancellation in flight)"
@@ -444,7 +446,7 @@ def _fail_job(job_id: int, reason: str) -> None:
     try:
         with transaction.atomic():
             job = Job.objects.select_for_update().get(pk=job_id)
-            if job.status in (JobState.CANCELING, *JobState.final_states()):
+            if job.is_settled():
                 return
             job.update_status(JobState.FAILURE, save=False)
             job.finished_at = datetime.datetime.now()
@@ -1327,7 +1329,34 @@ def cleanup_async_job_if_needed(job) -> None:
 
 @task_prerun.connect(sender=run_job)
 def pre_update_job_status(sender, task_id, task, **kwargs):
-    # in the prerun signal, set the job status to PENDING
+    """Bump the job to PENDING when a worker picks the message up.
+
+    Skipped when the job is already settled (terminal state) or being
+    cancelled. Without that guard, a broker redelivery (acks_late + worker
+    crash) or a cancel that arrived while the message was still in the
+    prefetch buffer would have its REVOKED/CANCELING status silently
+    overwritten with PENDING here, and the ``run_job`` early-guard
+    (which reads ``Job.status`` after this signal fires) would then fail
+    to short-circuit and re-run the job. See RolnickLab/antenna#1323.
+    """
+    from ami.jobs.models import Job
+
+    job_id = task.request.kwargs.get("job_id") if task.request.kwargs else None
+    if job_id is None and task.request.args:
+        job_id = task.request.args[0]
+    if job_id is not None:
+        try:
+            job = Job.objects.only("status").get(pk=job_id)
+        except Job.DoesNotExist:
+            pass
+        else:
+            if job.is_settled():
+                logger.info(
+                    "task_prerun: skipping PENDING write for job %s in status %s " "(redelivery or cancel in flight)",
+                    job_id,
+                    job.status,
+                )
+                return
     update_job_status(sender, task_id, task, "PENDING", **kwargs)
 
 
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -136,30 +136,53 @@ def update_async_services_seen_for_project(project_id: int) -> None:
     )
 
 
-@celery_app.task(bind=True, soft_time_limit=default_soft_time_limit, time_limit=default_time_limit)
+# acks_late + reject_on_worker_lost so a worker SIGKILL/OOM mid-task does not
+# silently drop the job: the broker holds the message until the task body
+# either completes successfully or raises, and redelivers if the worker dies.
+# Pairs with the early-guard below — a redelivered run_job that finds the job
+# already in a terminal state (or mid-cancellation) returns cleanly instead of
+# re-running side effects. See RolnickLab/antenna#1323.
+@celery_app.task(
+    bind=True,
+    soft_time_limit=default_soft_time_limit,
+    time_limit=default_time_limit,
+    acks_late=True,
+    reject_on_worker_lost=True,
+)
 def run_job(self, job_id: int) -> None:
-    from ami.jobs.models import Job
+    from ami.jobs.models import Job, JobState
 
     try:
         job = Job.objects.get(pk=job_id)
     except Job.DoesNotExist as e:
         raise e
         # self.retry(exc=e, countdown=1, max_retries=1)
-    else:
-        job.logger.info(f"Running job {job}")
-        try:
-            job.run()
-        except Exception as e:
-            job.logger.error(f'Job #{job.pk} "{job.name}" failed: {e}')
-            raise
-        else:
-            from ami.jobs.models import JobDispatchMode
 
-            job.refresh_from_db()
-            if job.dispatch_mode == JobDispatchMode.ASYNC_API and not job.progress.is_complete():
-                _log_worker_availability(job)
-            else:
-                job.logger.info(f"Finished job {job}")
+    # Early-guard: under acks_late, the broker may redeliver this message after a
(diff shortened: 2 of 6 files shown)


mihow:
**Claude says:** (posted on Michael's behalf — this is part of a structural review pass over the four open PSv2-readiness PRs: #1276, #1279, #1312, #1324, toward the umbrella #910.)

First, the parts of this PR that look solid: the `Job.cancel()` rewrite is a genuine root-cause fix — for `async_api` jobs the celery task is not where the work lives, so cleanup-then-REVOKED is the right cancel semantics, and it also fixes a real pre-existing bug where sync jobs with a `task_id` could sit in CANCELING forever. The new tests exercise the signals via `run_job.apply` rather than mocking them away, and the mid-bootstrap cancellation test writes a real status change mid-collect. That's the strongest part of the diff and we'd love to see it merge soon.

A few things seem worth discussing before the rest merges:

1. **`CELERY_WORKER_POOL_OPTIMIZATION` appears to be a no-op.** We could not find this setting anywhere in the celery package (checked against the pinned 5.4.x line and current releases): fair scheduling seems to be CLI-only (`celery worker -O fair`), wired through `WorkController.setup_defaults` as a kwarg from the CLI parser, with no `app.conf` lookup. Since the worker start scripts under `compose/` aren't modified here, the head-of-line blocking incident this PR targets (#1323) would likely recur after merge. Suggestion: add `-O fair` to the worker start command(s) and drop the settings entry. Easy to verify on staging by watching task distribution across prefork children.

2. **`acks_late` interacts badly with RabbitMQ's `consumer_timeout` in some of our environments.** `run_job` has a multi-day time limit, and with `acks_late=True` the message stays unacked for the task's whole runtime. RabbitMQ's default `consumer_timeout` is 30 minutes — past that it closes the channel and redelivers the message **while the original task is still running**. We checked our deployments: the production broker already has `consumer_timeout` raised well above the task limit, but the demo and staging rabbitmq containers (from `docker-compose.staging.yml`) run the 30-minute default — which is also exactly where the concurrent-load e2e validation for PSv2 runs. Two options: raise `consumer_timeout` in the compose file before this lands anywhere, or scope `acks_late` to the `async_api` path where the task body is short.

3. **Redelivery while a job is STARTED isn't guarded.** `is_settled()` covers final states + CANCELING, so a redelivered message for a STARTED job (consumer timeout above, broker hiccup, worker loss) passes both guards and re-runs `MLJob.run` concurrently with in-flight work — double `collect_images`, double NATS enqueue for not-yet-processed images. And since `reject_on_worker_lost=True` requeues without any delivery cap on the celery/RabbitMQ side, a job that reliably OOMs its worker becomes a redeliver loop. A test for the redelivered-while-STARTED case (or an explicit idempotency argument in the PR body) would settle this.

4. **Bookkeeping:** this PR closes #1323 — worth stating explicitly that #1282 (cancelled jobs served by `/next`) and #1283 (revoke cleanup) remain open and aren't covered here, so they don't get lost. The temporary-hack markers in `ami/jobs/views.py` that reference them are untouched by this diff.

Smaller cleanups, low priority: the `is_settled()` docstring says the mid-bootstrap check uses it, but that check re-implements the predicate inline — worth unifying since centralizing the predicate is the point; and Copilot's note about `update_job_status` (task_postrun) lacking the same guard symmetry still looks open.

Given how much of this is good, one path worth considering: split the `cancel()` fix into its own PR and merge it now, and let the dispatch/acks_late half follow once items 1–2 are resolved. Happy to help verify the `-O fair` behavior on staging.

mihow:
**Claude says:** Recommendation now that the #1337 work has landed — this PR is mostly superseded and worth closing once its one unique win is salvaged. Flagging rather than closing since it's your PR.

- **`cancel()` rewrite (skip `revoke(terminate=True)` for ASYNC_API)** — the genuinely useful part, and it isn't anywhere else. I've folded it into #1342 (which now owns `cancel()` after #1338/#1342), so the two PRs no longer edit the same method in conflicting ways. Your reasoning carried over verbatim: the local `run_job` only queues images and has usually finished, so terminate does nothing about the remote ADC work — that's stopped by the NATS/Redis teardown.
- **`CELERY_WORKER_POOL_OPTIMIZATION = "fair"`** — this is a no-op as a Django setting; fair scheduling is the worker CLI flag `-O fair` (in the worker `start` script), not a settings value. So it doesn't deliver the queue-unblocking the PR describes.
- **`acks_late` / `reject_on_worker_lost` crash-resume** — the one piece left, and it needs gating before it's safe: demo/staging RabbitMQ `consumer_timeout` is 30 min vs prod's 7 days, so a `run_job` longer than 30 min on those boxes gets channel-killed and redelivered = duplicate concurrent jobs (and it overlaps the 10-min reaper). The stranded-STARTED recovery it targets is already handled by the reaper today. Suggest splitting `acks_late` into its own PR gated behind a `consumer_timeout` fix + an idempotent `initialize_job` (#1337 follow-up), or dropping it.

Net: once the skip-terminate fold in #1342 merges, this can be closed. Happy to open the small `acks_late`-gated PR separately if you want to keep that piece.

mihow:
**Claude says:** Closing — superseded by #1342, which merged. The one genuinely useful change here (skip `revoke(terminate=True)` for ASYNC_API in `Job.cancel()`) was folded into #1342 with the same reasoning. The other parts didn't carry their weight: `CELERY_WORKER_POOL_OPTIMIZATION="fair"` is a no-op as a Django setting (fair scheduling is the `-O fair` worker CLI flag), and the `acks_late`/`reject_on_worker_lost` change is risky against the 30-min `consumer_timeout` on demo/staging while the stale-job reaper already recovers stranded jobs. If we want the crash-resume behaviour later, it should be its own PR gated behind a `consumer_timeout` fix and an idempotent `initialize_job`.

mihow:
**Claude says:** *Replacing an earlier version of this comment. One half of it was outright wrong and is retracted; the other half needed one word softened.*

### Retracted: my claim about `-O fair`

I wrote that `CELERY_WORKER_POOL_OPTIMIZATION = "fair"` was correctly rejected as a no-op **but that the real fix, the `-O fair` CLI flag, was never applied anywhere**, leaving the fairness problem unaddressed. The first half stands. **The second half is wrong**, and it is wrong in a way that would waste someone's time.

On the pinned celery version, `celery/concurrency/asynpool.py:80-86`:

```python
SCHED_STRATEGIES = {
    None: SCHED_STRATEGY_FAIR,
    'default': SCHED_STRATEGY_FAIR,
    'fast': SCHED_STRATEGY_FCFS,
    'fcfs': SCHED_STRATEGY_FCFS,
    'fair': SCHED_STRATEGY_FAIR,
}
```

and the CLI option defaults to `'default'` with choices limited to `('default', 'fair')`. Both resolve to the **same** constant. A worker started with no `-O` flag is already using the fair strategy, and FCFS is not reachable from the command line at all.

So adding `-O fair` would change nothing. My literal observation — that no start script or compose file passes `-O` — was true, but the inference drawn from it was not. There is no shelved fix here.

That has a consequence beyond this PR, which I have noted on #1323: if the scheduler was already fair, then the unfair-pre-assignment explanation for that issue's symptom cannot be right, and its root cause needs revisiting rather than its fix reapplying.

### Softened: the `acks_late` interaction

The substance holds. `CELERY_WORKER_CANCEL_LONG_RUNNING_TASKS_ON_CONNECTION_LOSS = True` is set but currently inert, because the cancel loop is guarded on `request.task.acks_late` and nothing in this codebase sets it — no per-task flag, no global. Ordering is also favourable: that loop runs before `on_close()` and `blueprint.restart()`. And `request.cancel(self.pool)` is not merely bookkeeping; it reaches a real `SIGTERM` to the child.

But I wrote that with `acks_late` enabled it "would have cancelled the blocked in-flight tasks at connection loss." That is one word too strong — **"would have attempted to terminate"** is accurate. It sends exactly one `SIGTERM` with no escalation to `SIGKILL`, only covers requests that have actually started, and relies on a Python-level signal handler in the child. A child blocked in a C-level call that retries internally on `EINTR` may never run that handler, and a socket read on a dead connection is exactly that shape.

So the interaction remains a real argument for revisiting `acks_late` that this PR's close rationale did not weigh — but it is a *possible* mitigation, not a certain one, and it should not be adopted on that basis alone given the duplicate-and-loss semantics of cancelling in-flight result tasks.

PR #1331, opened 2026-06-04
Serve direct thumbnail URLs to reduce load on main app server

mihow:
**Stacked on #1306.** Base branch: `feat/thumbnail-source-images`. Several pieces of the original proposal were promoted to the parent PR on 2026-06-10 so it deploys safely on its own; what remains here is only the warm/cold URL classification (+223/−20 over four files).

## Summary

If a thumbnail has already been generated, and the source image hasn't changed, list responses return the direct file-storage URL instead of the Django redirect route. Cold or stale thumbnails keep the existing redirect route, which generates them on first request.

Here you can see the small thumbnail has already been fetched, but the medium one has not.
image

## Two separate costs — what this fixes and what it must not introduce

Earlier drafts of this PR (and some review discussion) conflated these, so spelling them out:

1. **Browser HTTP round-trips — what this PR removes.** With the parent alone, every thumbnail `` is a request through `SourceImageThumbnailViewSet`: DRF auth, `get_object()`, a thumbnail-row check, then a 302 to storage. This is not a database N+1 — it is per-image HTTP fan-out, paid by every client. A 40-card gallery page is 80 of those requests. The parent has since reduced the per-request cost (no storage existence check, and the 302 is now browser-cached for five minutes), but each client still pays the full fan-out on first view and again whenever the cache expires. This PR removes it for warm rows: a [measurement on a dev deployment](https://github.com/RolnickLab/antenna/pull/1331#issuecomment-4653895903) showed Django requests per warm gallery view going from 80 to 0.

2. **Database N+1 — what this PR must avoid introducing.** The parent's serializer never touches the thumbnail table: it builds route URLs from `obj.pk` plus settings, so there is no pre-existing N+1 to fix (an earlier draft wrongly claimed there was — see [this comment](https://github.com/RolnickLab/antenna/pull/1331#issuecomment-4652469570) and Luke's [review thread](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373654234)). This PR makes the serializer read `obj.thumbnails.all()` per row to decide warm vs cold, which *would* be a new N+1 on its own. `SourceImageQuerySet.with_thumbnails()` prefetches it as one extra query per page, and the tests pin that contract.

In short: the win is fewer HTTP requests to Django; the prefetch is self-defense against a regression this PR would otherwise create, not a query-count improvement over the parent.

## Why generation stays lazy (and why the cached row matters)

Captures have a sparse access pattern: a project can hold on the order of a million captures, but users review occurrence crops — captures themselves are opened only when troubleshooting or checking the context around a detection, so only a few hundred per project are ever viewed. That rules out pre-warming thumbnails at upload/sync time (it would spend storage and compute on images nobody opens) and with it the simpler "deterministic URL, no lookup" design, which only works when the blob is guaranteed to exist.

With generation lazy, something must answer "has this thumbnail been generated, at the current spec, and is it still fresh?" per capture and label — that is exactly what the `SourceImageThumbnail` row records, and why this PR reads it (via the prefetch below) instead of constructing URLs blindly.

### `ami/main/models.py`

- **New `SourceImageQuerySet.with_thumbnails()`** — a plain `prefetch_related("thumbnails")` so `thumbnail_urls` can decide warm/cold in memory instead of firing a SELECT per row. (An earlier revision pushed an `is_source_changed` annotation into this prefetch; it was dropped in favour of the in-memory `thumbnail_is_valid` predicate below, so the same check is shared with the generator.)
- **New `SourceImage.thumbnail_urls(request)`** — per-label `{label: url}` builder, the single source of truth for the warm/cold decision. Warm = a cached row exists with a non-blank path, a width equal to the requested spec width, and no source change → `default_storage.url(row.path)`. Anything else → route URL via `reverse_with_params`, exactly as the parent serializer built it. The width check is strict equality because the generator (since the spec-width fix in #1306) records the requested width on the row; legacy rows that stored the encoder's output read as stale here and self-heal through one regeneration.
- **New `SourceImage.thumbnail_is_valid(spec, thumb)`** — the in-memory validity predicate (non-blank path, width equal to the spec, source not changed) shared by `thumbnail_urls` and the lazy generator gate, so the warm/cold rule lives in one place (per Luke's [review](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3414271551)).
- **Un-prefetched fallback** — if `thumbnails` was not prefetched (a freshly created instance in a write response, or a caller that skipped `with_thumbnails`), every label falls back to the route URL **without** querying. This keeps single-object serialization working and guarantees no per-row N+1; it also means the warm-cache win is enforced by the list query-count test rather than at runtime (until Django zen-queries lands).

### `ami/main/api/serializers.py`

`SourceImageThumbnailSerializer.get_thumbnails` becomes a one-line passthrough:

```python
def get_thumbnails(self, obj):
    return obj.thumbnail_urls(request=self.context.get("request"))
```

### `ami/main/api/views.py`

- `SourceImageViewSet.get_queryset` → `.with_thumbnails()`.
- `EventViewSet.get_queryset`'s inner `example_captures` Prefetch → same.

### List of changes

| Change (user/operator effect) | How (implementation) | Notes |
|---|---|---|
| Warm-cache gallery pages no longer route thumbnail requests through Django | Serializer emits `default_storage.url(row.path)` when the cached row matches the configured width and isn't source-changed | Removes up to 80 browser→Django round-trips per page view. Frontend treats the URL as opaque (``); no FE change. |
| Cold and stale thumbnails still regenerate via the redirect viewset | Same `SourceImageThumbnailViewSet` + concurrency-safe upsert | No behavior change on this path. |
| Re-uploaded source images don't serve stale thumbnails from the warm path | In-memory `thumbnail_is_valid` compares row vs source `last_modified`; stale rows fall back to the route URL | Mirrors the predicate the generator already uses. Guards [r3373715397](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373715397). |
| Thumbnails generated before the #1306 spec-width fix self-heal | Rows whose stored width is the encoder output (e.g. 239 for a 240 spec) read as stale and get the route URL, regenerating once | One extra regeneration per legacy row, then warm forever. |
| The serializer's new per-row thumbnail read can't become an N+1 | `with_thumbnails()` prefetch is the contract; viewsets wire it; un-prefetched callers fall back to route URLs without querying; a multi-row test asserts a warm list serves storage URLs with one thumbnail query | One extra query per page, not per row. |
| Per-label URL contract tested at the model surface | `test_thumbnail_urls_model_method_emits_warm_and_route` exercises `SourceImage.thumbnail_urls` directly | Future template tags / management commands pick up the contract for free. |

## Contract change

`thumbnails.{small,medium}` URL shape changes for cached rows from `http://<host>/api/v2/captures/thumbnails/<id>/?label=small` to e.g. `https://<bucket>.s3.amazonaws.com/thumbnails/capture_<id>/small.jpg`.

Frontend treats the value as opaque (see `ui/src/data-services/models/capture.ts:170-182`, `ui/src/pages/captures/capture-columns.tsx`). No frontend change required.

On deployments using `FileSystemStorage` (e.g. local dev), `default_storage.url()` returns a relative `/media/...` URL that Django still serves — the "bypass Django" win applies fully only where storage URLs are external (S3/MinIO). Compose with a CDN later via `AWS_S3_CUSTOM_DOMAIN` and these URLs front-load automatically.

## Promoted to #1306 (no longer in this diff)

These shipped on the parent so it is solo-deployable: removal of the per-request `default_storage.exists()` check, progressive JPEG encoding, `Cache-Control: private, max-age=300` on the redirect, the blank-path regen guard, and the spec-width fix (replacing the 2px width tolerance an earlier revision of this branch carried).

## Luke's review threads addressed

- [r3373388308](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373388308) `reverse_with_params` already exists — adopted.
- [r3373625304](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373625304) `lru_cache` helper is obfuscation — dropped; warm path just calls `default_storage.url(row.path)`.
- [r3373654234](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373654234) N+1 framing misleading — rewritten (see the "Two separate costs" section above).
- [r3373715397](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373715397) source-changed bug — fixed via the `is_source_changed` annotation and warm-path check.
- [r3373735406](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373735406), [r3373777016](https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373777016) comment hedges — addressed; the code they referred to now lives in #1306.

## What this does NOT change

- The redirect viewset (`SourceImageThumbnailViewSet`) and its concurrency-safe upsert path are unchanged.
- Browser cache semantics for storage URLs are governed by the storage backend's headers, unchanged here.
- Frontend code is unchanged.

## Known follow-ups

- **Orphan-thumbnail reconciler** — a row whose storage blob was deleted out of band now produces a broken `` until the row is removed (consequence of the `exists()` removal, which lives in #1306). Planned as a periodic integrity-check task; separate ticket.
- `Event.first_capture()` at `ami/main/models.py:1277` is a per-row method that runs a fresh query; pre-existing, unchanged.

## Verification

- Browser network panel on a warm gallery page: `` requests go direct to storage; zero requests to `/api/v2/captures/thumbnails/...`.
- Captures-list endpoint query count: +1 thumbnail prefetch query vs the parent, not +N (asserted by `test_warm_list_serves_storage_urls_with_single_thumbnail_query`).
- First measurement on a dev deployment is in [this comment](https://github.com/RolnickLab/antenna/pull/1331#issuecomment-4653895903): Django round-trips per gallery view went from 80 to 0 once the cache was warm.

## Summary by CodeRabbit

* **Performance**
  * Improved thumbnail loading by prefetching cached thumbnail metadata, reducing repeated database access while browsing image captures and galleries.
  * Updated thumbnail URL behavior: valid cached thumbnails now return direct storage URLs (“warm”); stale/missing thumbnails fall back to the thumbnail detail route (“cold”).

* **Tests**
  * Added regression and contract tests to verify warm/cold thumbnail URL selection, cache invalidation when source images change, and query-efficiency guarantees during listing.


mihow:
@coderabbitai review

mihow:
**Claude says:** Adjacent perf question that came up while landing the progressive-JPEG change in `46d5fe53`: is it worth flipping the thumbnail format from JPEG to WebP (or AVIF) in a follow-up? Summarizing the tradeoffs here so we can either decide now or punt to a separate ticket.

## WebP

**Pros**
- ~25–35% smaller than JPEG at equivalent perceptual quality. Real bandwidth win on dense gallery views (40-card pages) and on mobile.
- Browser support is **~96%** as of mid-2026: Safari 14+ (Sep 2020), all evergreen Chromium/Firefox, all current mobile. The legacy long tail is small enough that a `<picture>` fallback isn't worth the markup complexity for thumbnails.
- Pillow handles it natively (`format="WEBP"`, no extra dependency).
- Plays nicely with the future content-addressed-paths + `immutable` cache-control plan — `_<hash>.webp` is the same shape as `_<hash>.jpg`.

**Cons**
- Encode is ~3–5× slower than JPEG. Doesn't matter here because this is a write-once cache and the encode only fires on cold-path requests — never on the user-facing warm path that the rest of this PR was about.
- Format change in the storage path: today `capture_<pk>/<label>.jpg`, post-flip `capture_<pk>/<label>.webp`. Existing `.jpg` rows phase out lazily as users browse and regen triggers fire (width change, source mtime bump). No bulk migration needed; the orphan-thumbnail reconciler (the planned follow-up #1331 names) eventually cleans the leftover `.jpg` blobs.
- Two settings keys to plumb (`FORMAT`, `QUALITY`), which is small but real wiring.

## AVIF

**Pros**
- 10–20% smaller again than WebP at the same perceptual quality.
- Safari 16+ (Sep 2022), all evergreen Chromium/Firefox — ~94% support. Marginally less than WebP, still well within "don't bother with a fallback for thumbnails."

**Cons**
- Encode is **5–10× slower than WebP** — cold-path-only here, but the cold path matters more than it sounds when you fire `create_detection_images` over a 700k-capture session and every thumbnail goes through the same encoder. WebP is bounded; AVIF starts being noticeable.
- Pillow's AVIF support requires the `pillow-avif-plugin` package (third-party, GPL-compatible but extra dependency surface).
- The size-vs-encode-time math is worse than WebP for the gallery use case. AVIF wins more on photographic detail and high-resolution display assets, less on 1024px screen-sized thumbnails where WebP is already very close to lossless-perceptual.

Verdict: **AVIF probably isn't worth the encoder-time cost for this PR's use case.** Revisit in ~2027 if encoder libraries get faster.

## Suggested shape: pluggable format

If we do this, the smallest reasonable surface is two new keys on `settings.THUMBNAILS`:

```python
THUMBNAILS = {
    "STORAGE_PREFIX": "thumbnails/",
    "FORMAT": "webp",      # or "jpeg", default "jpeg" for dev/test, "webp" in production
    "QUALITY": 80,         # PIL accepts 0-100; WebP sweet spot ~80, JPEG ~82
    "METHOD": 4,           # WebP-only encode effort 0-6 (default 4); ignored for JPEG
    "SIZES": {"small": {"width": 240}, "medium": {"width": 1024}},
}
```

`find_or_generate_thumbnail_for_label` reads `FORMAT`, derives the extension and PIL save kwargs, and the storage path becomes `capture_<pk>/<label>.{ext}`. The model's `_thumbnail_warm_url` already builds URLs from `row.path` so no serializer change needed.

Open questions worth pinning down before opening the follow-up PR:

1. **Dev/test parity** — should the test suite encode WebP or stay on JPEG? WebP makes some PIL test fixtures that hard-code JPEG markers / quality assumptions slightly more fragile. JPEG-in-tests + WebP-in-prod is operationally fine but means CI doesn't catch WebP-specific bugs (e.g. mode conversion edge cases that differ from JPEG). Lean toward WebP-in-tests with the `THUMBNAILS["FORMAT"]` flag respected throughout, so the behavior matches prod.
2. **Migration during the lazy phase-out** — for a busy deployment (production-scale), do we want to actively prime regen for cached `.jpg` rows, or trust browsing traffic to drive it? Lean toward "trust browsing"; a one-shot management command (`manage.py prime_thumbnails --format webp`) is the escape hatch if it's slower than expected.
3. **Whether `THUMBNAILS["FORMAT"]` should default to `"jpeg"` or `"webp"`** in `config/settings/base.py`. Production override is fine either way. Defaulting to `"webp"` says "this is the recommended format now"; defaulting to `"jpeg"` keeps the historical default and explicit opt-in for new behavior. Lean toward `"webp"` since the contract is the same path-with-different-extension.

Happy to open a separate ticket for this if you'd rather keep #1331 focused on the queryset/model rework + the progressive-JPEG win that's already in. Otherwise I can land the WebP flip as a third commit on this branch.

ami/main/models.py
```diff
@@ -1930,11 +1931,65 @@ def with_was_processed(self):
...
+    """
+    return default_storage.url("")
+
+
+def _thumbnail_warm_url(path: str) -> str:
+    """Direct storage URL for a cached thumbnail blob (warm path)."""
+    return _thumbnail_storage_url_base() + urllib.parse.quote(path.lstrip("/"), safe="/")
+
+
+def _thumbnail_route_url(pk: int, label: str, request: Request | None) -> str:
```
loppear: This already exists as `ami.base.serializers.reverse_with_params` as used in the code this PR refactors?

ami/main/models.py
```diff
@@ -1930,11 +1931,65 @@ def with_was_processed(self):
...
+        """
+        return self.prefetch_related("thumbnails")
+
 
 class SourceImageManager(models.Manager.from_queryset(SourceImageQuerySet)):
     pass
 
 
+@functools.lru_cache(maxsize=1)
+def _thumbnail_storage_url_base() -> str:
```
loppear: I'm not sure any of this obfuscation to avoid calling `storage.url` is worthwhile?

ami/main/tests.py
```diff
@@ -476,6 +505,144 @@ def test_thumbnail_delete_removes_storage_blob(self):
...
+                path=f"thumbnails/preexisting_{cap.pk}.jpg",
+                label="small",
+                width=configured_width,
+                height=180,
+                size=0,
+            )
+        self.assertGreater(len(captures), 1, "Test fixture must produce >1 capture to exercise N+1.")
+
+        with CaptureQueriesContext(connection) as ctx:
+            response = self.client.get(f"/api/v2/captures/?project_id={self.project.pk}")
```
loppear: This test couldn't fail before, right? The N queries this is all predicated on is over the retrieval of every image url redirect, not in just the `captures` list response.

ami/main/models.py
```diff
@@ -2240,6 +2295,45 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+        Returns:
+            Ordered ``{label: url}`` matching ``settings.THUMBNAILS['SIZES']``.
+        """
+        sizes = settings.THUMBNAILS["SIZES"]
+        existing_by_label: dict[str, "SourceImageThumbnail"] = {t.label: t for t in self.thumbnails.all()}
+
+        out: dict[str, str] = {}
+        for label, spec in sizes.items():
+            row = existing_by_label.get(label)
+            if row and row.path and row.width == spec["width"]:
```
loppear: In order to determine whether to skip the redirect, this check needs to reproduce the `find_or_generate_thumbnail_for_label` check, looking at `last_modified` (`source_changed`).

ami/main/models.py
```diff
@@ -2272,7 +2375,22 @@ def find_or_generate_thumbnail_for_label(self, label):
...
+            #
+            # ``optimize`` runs an extra Huffman pass for ~2–5% smaller files at
+            # ~10–15% slower encode. Cold-path only (one encode per cache miss),
+            # so the encode-time cost never lands on the user-facing path.
+            #
+            # ``quality=82`` is the mozjpeg-studies sweet spot — visually
+            # identical to the Pillow default of 75 on a screen-sized thumbnail
+            # but noticeably crisper on saturated edges. Bytes grow ~10% vs 75;
+            # still well under any uncompressed equivalent.
+            img.save(buffer, format="JPEG", progressive=True, optimize=True, quality=82)
```
loppear, line 2522: I mean, maybe?

ami/main/models.py
```diff
@@ -2255,7 +2349,16 @@ def find_or_generate_thumbnail_for_label(self, label):
         source_changed = (
             self.last_modified is not None and thumb is not None and thumb.last_modified < self.last_modified
         )
-        if not thumb or thumb.width != size["width"] or source_changed or not default_storage.exists(thumb.path):
+        # We deliberately do NOT call ``default_storage.exists(thumb.path)`` here: a
+        # HEAD per request is expensive on remote storage, and the cached row's
+        # presence is the authoritative warm signal. The narrow case of an orphan
+        # row (DB row points at a storage path whose blob was deleted out-of-band)
+        # is handled by the orphan-thumbnail reconciler (separate follow-up
```
loppear: Probably worth it (I don't think there's a path for orphaning now), but maybe a comment that we're not handling orphans rather than that we're implementing this other thing.

loppear (changes requested):
I'm probably good with the intention here to split the cost differently between list & image retrieval - lookup thumbnails during API response by join, and skip serving through django redirect if we've already generated it. A lot of this is incidental to making that tradeoff, however.

mihow:
**Claude says:** On a re-read prompted by Luke's review, my initial analysis behind this PR was wrong on a fundamental point. Flagging it here before going any further on rework, because it changes the merge calculus.

## What the PR body claimed

> Captures-list endpoint adds **one** SELECT against `main_sourceimagethumbnail` regardless of card count, not N.

The `test_captures_list_prefetches_thumbnails_no_n_plus_one` test docstring reinforced this: "must hit `main_sourceimagethumbnail` at most once via prefetch, not once per row."

I framed this as fixing a pre-existing N+1 in the captures-list response.

## What the code actually does

Pre-PR (parent branch, `feat/thumbnail-source-images` at `22045e07`), `SourceImageThumbnailSerializer.get_thumbnails` reads `obj.pk` and settings only:

```python
def get_thumbnails(self, obj):
    return {
        label: reverse_with_params(
            "sourceimagethumbnail-detail",
            args=(obj.pk,),
            request=self.context.get("request"),
            params={"label": label},
        )
        for label in settings.THUMBNAILS["SIZES"]
    }
```

No `obj.thumbnails` access. Zero DB queries per row. No prefetch needed in `SourceImageViewSet.get_queryset` — the parent branch's queryset doesn't include a `thumbnails` prefetch and never did.

**There is no pre-existing N+1 against `main_sourceimagethumbnail` in the captures-list response.** The N+1 the test guards is one this PR's warm-URL feature would introduce — without `with_thumbnails()`, the serializer collapses to per-row queries because we made it read `obj.thumbnails.all()` for the cached-row check. The test still has value as a regression guard for the prefetch, but it's pinning a contract this PR creates, not one it repairs.

Luke called this out at https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373654234 and was right.

## What the PR actually wins

The real cost the PR removes is **80 HTTP round-trips per gallery page view** through the `SourceImageThumbnailViewSet` redirect endpoint. Per page view (40 cards × 2 thumbnails):

| Step | Pre-PR | Post-PR |
|---|---|---|
| `/api/v2/captures/` request | 1 | 1 |
| `/api/v2/captures/thumbnails/...` redirect requests | 80 | 0 |
| S3 GET for blob bytes | 80 | 80 |

Each of the 80 redirect requests goes through DRF auth + `get_object()` + `find_or_generate_thumbnail_for_label` (DB + a `default_storage.exists()` HEAD) before 302-ing to the storage URL. The 302 has no `Cache-Control` so the browser doesn't cache it; every page view repeats the fan-out.

That's a real cost. It's an **HTTP-level fan-out problem, not a DB-level N+1.** Different shape, different remedy space. The PR body framed it as the latter.

## Correctness regression I missed

The original architecture has the redirect viewset acting as an implicit "is this thumbnail still fresh?" check on every browser request — `find_or_generate_thumbnail_for_label` examines `source_changed = self.last_modified > thumb.last_modified` and regens if true.

This PR's warm path skips that check entirely. Luke flagged this at https://github.com/RolnickLab/antenna/pull/1331#discussion_r3373715397. Concrete failure mode:

1. SourceImage bytes are re-synced or re-uploaded → `SourceImage.last_modified` bumps
2. Thumbnail row's `last_modified` is now stale; row's `path` still points at the OLD blob
3. Captures-list response emits the warm URL → browser GETs the OLD blob from storage
4. No code path detects the staleness until someone hits the redirect viewset on a cold path (which may never happen if everyone reaches thumbnails via the gallery list)

User sees a stale thumbnail indefinitely. Pre-PR, this self-healed on the next page view.

## Alternative approaches worth weighing before pushing rework

1. **Do nothing.** A 5ms gunicorn round-trip × 80 per page view = 400ms of worker time per page view, distributed across the pool. Whether that's a bottleneck is a measurement question, not a code question. The PR was opened on assumption, not on a measured baseline. Worth running pass-1 of the NR measurement on the current parent branch deploy before deciding the rework is justified.
2. **Cache the 302.** The redirect URL is content-addressable by `(pk, label)`. Set `Cache-Control: public, max-age=N` on the redirect response and put nginx / Cloudflare in front. CDN serves the 302 from cache; Django hit only on cache miss. Browser still gets a normal 302 → S3 flow; auth and source-changed checks survive on cold path.
3. **Separate Django worker pool for thumbnails.** Run a second gunicorn pool (or a separate service) that only handles `/api/v2/captures/thumbnails/...`. Main app's pool stays free. No code change. Independent scaling.
4. **Signed S3 URLs from the captures-list endpoint.** Direct-to-storage like this PR, but with short-TTL signed URLs maintained via `AWS_QUERYSTRING_AUTH=True` (off in production today). Keeps auth context. Still has the source-changed problem.
5. **Continue this PR.** Drop the over-engineered helpers per Luke 1+2, add the source-changed check per Luke 4, reframe the PR body per the above. Net: a correct but architecturally invasive change that solves something which may not have been measured as a problem.

My read is that **option 1 (measure first) followed by option 2 if needed** is probably the right path, and this PR may not be worth landing. The structural cost (lost source-changed check, frontend coupling to direct S3 URL shape, complexity of warm/cold branching in the serializer) is high for a perf gain we never validated.

Up to you on whether to keep iterating here or close in favor of measurement-first. I'd lean toward close; the architecture trade isn't favorable without a real bottleneck measurement.

ami/main/models.py
mihow: **Claude says:** Adopted in 545873af. Dropped the bespoke route helper, using `reverse_with_params` directly. Kept the explicit `api:` namespace so the model method works request-less (mgmt commands, template tags) as well as via DRF context.

ami/main/models.py
mihow: **Claude says:** Fair point — the `lru_cache` + base-prefix helpers were chasing a speculative boto3 overhead we never measured. Dropped in 545873af; warm path is now just `default_storage.url(row.path)` inline. The string-concat-vs-storage-backend distinction is below noise on the gallery hot path.

ami/main/tests.py
mihow: **Claude says:** You're right — there was no pre-existing N+1 in the parent branch's captures-list serializer; the N+1 that test guarded was one this PR introduced by reading `obj.thumbnails.all()` from the warm-row check. Dropped that test and the queryset-method prefetch test in 545873af. Rewrote PR body to lead with what this actually wins (80 HTTP round-trips per gallery page, not a DB query count change). Full reasoning in https://github.com/RolnickLab/antenna/pull/1331#issuecomment-4652469570.

ami/main/models.py
mihow: **Claude says:** Correctness fix in 545873af. `thumbnail_urls` now mirrors the generator's `source_changed` predicate (`row.last_modified < source.last_modified`) and falls back to the route URL when the source has been re-uploaded, so the next browser fetch regenerates. Added `test_serializer_falls_back_to_route_url_when_source_changed` to pin it.

ami/main/models.py
mihow: **Claude says:** Tightened in 545873af — dropped the speculative percentages and the mozjpeg-studies justification. Just names what each flag does and that the encode cost is cold-path only.

ami/main/models.py
mihow: **Claude says:** Reframed in 545873af. Comment now says we explicitly do NOT detect or repair the orphan case here — missing blob shows broken `` until the row is removed and the next request regenerates — instead of claiming there's a reconciler that doesn't exist yet.

ami/main/models.py
```diff
@@ -2240,6 +2288,45 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+
+        For each label configured in ``settings.THUMBNAILS['SIZES']``:
+
+        * If a cached :class:`SourceImageThumbnail` row exists and its width
+          still matches the configured size, emit a direct storage URL — the
+          warm path, which bypasses Django entirely for the thumbnail fetch.
+        * Otherwise (no row, missing path, or stale width), emit the redirect
+          viewset's route URL. The next browser fetch flows through Django
+          and triggers lazy (re)generation via
+          :meth:`find_or_generate_thumbnail_for_label`.
```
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in 545873af. Added `not thumb.path` to the regen gate in `find_or_generate_thumbnail_for_label`, so a row with an empty `path` triggers a real generate instead of being treated as warm.

ami/main/tests.py
```diff
@@ -319,25 +319,24 @@ def test_thumbnail_regen_reuses_row_via_upsert(self):
...
-        # ``not default_storage.exists(thumb.path)``.
-        default_storage.delete(original_path)
-        self.assertFalse(default_storage.exists(original_path))
+        # Bump the source mtime so the freshness check (``thumb.last_modified <
+        # self.last_modified``) flips to True and forces the regen branch. The
+        # warm-path code no longer HEADs the storage blob — orphan blobs are
+        # the reconciler's job, not the generator's — so a missing-blob test
+        # would no longer trigger regen.
+        self.first_capture.last_modified = datetime.datetime.now() + datetime.timedelta(minutes=1)
+        self.first_capture.save()
```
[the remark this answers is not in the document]
mihow: **Claude says:** Fixed in 545873af. Switched to `timezone.now()` to match `USE_TZ` settings.

mihow:
**Claude says:** Arctia two-pass measurement landed. Revising my own framing — earlier write-up overclaimed.

**Important caveat about arctia (didn't catch this initially):** arctia uses `FileSystemStorage` with `MEDIA_URL=/media/`, so `default_storage.url(row.path)` returns a relative `/media/...` URL. That URL still hits Django — `config/urls.py` includes `static(settings.MEDIA_URL, …)`, which routes every `/media/` request through the full Django middleware stack (auth, sessions, CSRF, content negotiation — response headers confirm `vary: Accept-Language`, `x-frame-options: DENY`, etc., applied by Django middleware). The thumbnail bytes get served by Django's `static()` view reading from disk.

So the on-arctia win isn't "skip Django for warm thumbnails" — it's **skip the redirect viewset's DRF auth + `get_object` + `find_or_generate_thumbnail_for_label` (DB lookup + `default_storage.exists()` HEAD) + 302 hop**. The middleware-stack hit per `/media/` request remains. Still meaningful (~270 ms → much less per thumbnail) but not the "Django bypassed" framing my earlier comment implied.

Where the "Django bypassed" framing **does** hold is on prod-like deployments using S3 storage with public bucket URLs: `default_storage.url(...)` then returns `https://<bucket>.s3.amazonaws.com/...` and browsers hit S3 directly with no Django involvement at all. That's the headline gain this PR is designed for — but it's not what arctia measures.

## Numbers (with corrected framing)

Test data: UKCEH project 1, event 823, 40 captures × 2 labels pre-warmed.

| Metric | Pass-1 (parent `feat/thumbnail-source-images`) | Pass-2 (this PR `aeb46c01`) | Delta |
|---|---|---|---|
| Per-thumbnail path: DRF auth + `get_object` + `find_or_generate` + `default_storage.exists()` HEAD + 302 | yes (80 / page) | no (0 / page) | **eliminated** |
| Per-thumbnail path: Django middleware + `static()` view + file read (on arctia FileSystemStorage) | yes (80 / page, follow-up of 302) | yes (80 / page, direct) | unchanged on arctia |
| Per-thumbnail path: external storage GET (on prod S3 setup) | yes (80 / page, follow-up of 302) | yes (80 / page, direct) | **Django eliminated in prod** |
| Captures-list response p50 (n=10, single client over public internet) | 1.109 s | 0.846 s | −24% |
| Captures-list response p95 (n=10) | 1.569 s | 0.923 s | −41% |
| Per single Django thumbnail-redirect call (pre-PR path) | 0.267 s | n/a — bypassed | — |
| 20 simulated gallery views wall-clock (40 capture cards × 2 labels each = 80 thumbnail follow-ups per view) | 49 s | 10 s | ~5× |

## On the wall-clock 5× number

The 5× is a single concurrency point, not a fixed ratio. My test fanned out 80 parallel `curl` requests per gallery view via `xargs -P 20` × 4 background views. Pass-1 each follow-up cost ~270 ms going through the redirect viewset; pass-2 each follow-up cost much less going through the `static()` view. With only 2 gunicorn workers on arctia those 80 parallel requests queue heavily in pass-1 — that queueing is most of the 49 s wall-clock. Cranking concurrency higher widens the gap (pass-1 tail-latency tanks while pass-2 scales). Running serial shrinks it. So 5× is "at this concurrency, on this 2-worker dev box" — directionally correct but don't treat it as a fixed multiplier.

## Bug surfaced during measurement → fix [`aeb46c01`](https://github.com/RolnickLab/antenna/commit/aeb46c01)

PIL.Image.thumbnail() preserves aspect ratio by shrinking, so a `spec=240` request often gets a `width=239` blob. The pre-existing strict `width != size["width"]` regen-gate then re-triggered on every cold hit, and the warm-path emitter on this PR would never fire for those rows. Fixed with a 2-px tolerance applied symmetrically in both the warm-check and the regen gate, plus a regression test (`test_serializer_treats_pil_rounded_width_as_warm`). Pre-existing latent bug on parent — exposed by the new code path.

## Caveats

- Latency numbers are n=10 from a single client over the public internet — suggestive, not conclusive.
- The 80 → 0 redirect-viewset elimination is mechanical (URL classification, not noise-dependent).
- The "Django middleware bypassed" win is **prod-S3-only** — does not apply on FileSystemStorage deployments like arctia.
- Arctia doesn't have New Relic enabled so the protocol's planned `SourceImageThumbnailViewSet.retrieve` span count + `HeadObject` rate metrics aren't measurable; for those a prod-monitored deploy is needed.

ami/main/models.py
```diff
@@ -2240,6 +2257,61 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+        rows: dict[str, "SourceImageThumbnail"] = {t.label: t for t in self.thumbnails.all()}
+
+        out: dict[str, str] = {}
+        for label, spec in sizes.items():
+            row = rows.get(label)
+            source_changed = (
+                row is not None
+                and self.last_modified is not None
+                and row.last_modified is not None
+                and row.last_modified < self.last_modified
```
mihow: Can source_changed be an annotation in the queryset method instead of looping through the thumbnails here? If not, rename `row`/`rows` to `thumb`/`thumbs` for clarity.

ami/main/api/views.py
```diff
@@ -593,11 +594,11 @@ def get_queryset(self) -> QuerySet:
...
         project = self.get_active_project()
 
-        queryset = queryset.select_related(
-            "event",
-            "deployment",
-            "deployment__data_source",
-        ).order_by("timestamp")
+        queryset = (
+            queryset.select_related("event", "deployment", "deployment__data_source")
+            # Required by SourceImage.thumbnail_urls — drops N+1 on SourceImageThumbnailSerializer.
```
mihow: Are they actually N+1 queries or N+1 requests? N+1 is a misleading statement that seems to be persisting.

ami/main/models.py
```diff
@@ -2255,7 +2327,13 @@ def find_or_generate_thumbnail_for_label(self, label):
         source_changed = (
             self.last_modified is not None and thumb is not None and thumb.last_modified < self.last_modified
         )
-        if not thumb or thumb.width != size["width"] or source_changed or not default_storage.exists(thumb.path):
+        # No ``default_storage.exists(thumb.path)`` here: HEAD-per-request on remote storage
+        # is expensive, and the cached row is the warm signal. We do NOT currently detect or
```
mihow: Remove reactive/corrective language, the all caps "NOT"

ami/main/models.py
mihow: @loppear @claude the point of this particular change was `progressive=True` to make the images load in a nicer looking way (blurry-crisp rather than top pixels-> bottom pixels). Please remove that inline comment.

ami/main/models.py
```diff
@@ -2240,6 +2257,61 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+                row is not None
+                and self.last_modified is not None
+                and row.last_modified is not None
+                and row.last_modified < self.last_modified
+            )
+            # PIL.Image.thumbnail preserves aspect ratio and may round the output
+            # dimensions down by a pixel or two vs the spec (e.g. spec=240 → actual
+            # 239). Use a small tolerance so existing rows aren't treated as stale
+            # by every warm-check; the same tolerance is applied in the regen gate.
+            width_match = row is not None and abs(row.width - spec["width"]) <= _THUMBNAIL_WIDTH_TOLERANCE
```
mihow: Discuss some alternatives to this approach of width matching with tolerance.

mihow:
@loppear thanks for reviewing this one. the agent went a bit overboard on my requests. But there are a few changes I like. I think we can consider these all optimizations and merge your main thumbs PR soon.

ami/main/api/views.py
mihow: **Claude says:** Scrubbed in `1b54cc8a`. Inline `views.py` comments + `serializers.py` docstring + `models.py` `with_thumbnails`/`thumbnail_urls`/`get_was_processed` docstrings all now say "each row triggers a separate SELECT on `main_sourceimagethumbnail`" (or "fires its own DB query" for the `get_was_processed` case). N+1 phrasing dropped.

ami/main/models.py
mihow: **Claude says:** Rewritten in `1b54cc8a`. New text:

```
# Cached row is the warm signal — no HEAD against storage on the regen check.
# Orphan rows (DB row whose blob was deleted out of band) show a broken ``<img>``
# until something removes the row and the next request regenerates.
```

No "NOT", flat behaviour description.

ami/main/models.py
mihow: **Claude says:** Done in `1b54cc8a`. Annotation route worked cleanly.

`SourceImageQuerySet.with_thumbnails()` now builds the prefetch queryset and annotates `is_source_changed` per thumbnail row:

```python
thumbnail_model = apps.get_model("main", "SourceImageThumbnail")
thumbnails_qs = thumbnail_model.objects.annotate(
    is_source_changed=models.Case(
        models.When(
            last_modified__lt=models.F("source_image__last_modified"),
            then=models.Value(True),
        ),
        default=models.Value(False),
        output_field=models.BooleanField(),
    )
)
return self.prefetch_related(models.Prefetch("thumbnails", queryset=thumbnails_qs))
```

The reverse FK `source_image__last_modified` is reachable from the prefetched queryset, so no `OuterRef`/`Subquery` was needed. `NULL < x` is `NULL` in SQL, which the `CASE` resolves as the `False` default — same behaviour the inline Python check had (it returned `False` when either side was `None`).

`thumbnail_urls()` is now pure URL-formatting: reads `thumb.is_source_changed` straight off the prefetched row. No more staleness loop. Renamed `row`/`rows` → `thumb`/`thumbs` while I was in there.

Contract is now strict: callers MUST go through `with_thumbnails()` — without it, `thumbnail_urls()` raises `AttributeError` on `thumb.is_source_changed`. Production callers (the serializer) already do; the direct-instance test was updated to re-fetch through `with_thumbnails()`.

ami/main/models.py
mihow: **Claude says:** Five options to weigh; current tolerance is the cheapest interim.

**a. Store spec width, not PIL's output.** Generator writes `row.width = spec["width"]` instead of `img.size[0]`. `row.width == spec["width"]` then works without tolerance. Trade: `row.width` becomes "what we asked PIL to produce" rather than "what PIL actually produced." Anything that reads `row.width` for layout (e.g. a hypothetical `` server-side) would be off by 1–2px.

**b. Store the spec key as a denorm field.** Add `generated_for_spec` (CharField, e.g. `"small@240"`) or full spec-dict on `SourceImageThumbnail` at generation time. Predicate becomes `row.generated_for_spec == settings.THUMBNAILS["SIZES"][label].canonical_key`. Migration cost; simplest semantic; future-proof if `THUMBNAILS["SIZES"]` ever grows fields beyond width/height.

**c. Generation-version field.** `THUMBNAILS["VERSION"]` bump in settings, stored on row, regen when `row.version != current`. Decouples regen from any per-row attribute. Adds a versioning burden — bump-on-edit discipline lives in settings, easy to forget.

**d. One-direction predicate (`row.width >= spec["width"]`).** PIL only shrinks, so anything ≥ spec is fresh-enough by definition. No tolerance constant. Doesn't handle the case where spec _shrinks_ to a smaller size — e.g. old row at 1024, new spec 800, would falsely be "fresh" and visually too big.

**e. Current tolerance (`abs(row.width - spec["width"]) <= 2`).** No schema change, single module constant, symmetric in both directions. Arbitrary 2-px threshold, opaque without context.

My read: **(a)** is cleanest semantically and matches how the row should be labelled ("this thumbnail represents what we wanted to generate, not what PIL emitted bit-for-bit"). It's a one-line generator change + no migration. **(b)** is the right long-term answer if `THUMBNAILS["SIZES"]` ever grows beyond width/height. The tolerance is fine as an interim.

If you want me to swap to (a) in this PR I'm happy to — it's a small follow-up commit. Or land the current tolerance as the unblocking fix and queue (a)/(b) as a follow-up issue. Your call.

mihow: @loppear what do you think about options a) or d) here?

ami/main/tests.py
```diff
@@ -388,6 +386,36 @@ def test_thumbnail_non_rgb_source_converts(self):
         thumb = self.first_capture.thumbnails.get(label="small")
         self.assertEqual(thumb.width, 240)
 
+    def test_thumbnail_jpeg_is_progressive(self):
```
mihow: this test feels overkill to merge after we are done with the TDD dev

ami/main/models.py
mihow: **Claude says:** Went with option (a). The generator now records the requested spec width on the row (parent branch commit 012efbc7, with a regression test that simulates Pillow's rounding), and the 2px tolerance is gone from both the warm-check and the regen gate (56bc46ad on this branch — strict equality everywhere). Legacy rows that stored encoder output read as stale, fall back to the route URL, and self-heal with a single regeneration. Option (b) (a `generated_for_spec` field) remains the long-term answer if thumbnail specs ever grow beyond width; queued as a follow-up rather than another migration in this pair of PRs.

ami/main/models.py
mihow: **Claude says:** Inline comment removed. The encode flags (progressive/optimize/quality=82) moved to the parent branch in cb83de32 without the comment, and this branch picked that up in the merge commit 4292ee42.

ami/main/tests.py
mihow: **Claude says:** Removed in ea4f0bd6 — agreed it pinned a Pillow encoder flag rather than behavior we own.

ami/main/models.py
```diff
@@ -2399,6 +2424,43 @@ def get_custom_user_permissions(self, user) -> list[str]:
...
+        sizes = settings.THUMBNAILS["SIZES"]
+        thumbs: dict[str, "SourceImageThumbnail"] = {t.label: t for t in self.thumbnails.all()}
+
+        out: dict[str, str] = {}
+        for label, spec in sizes.items():
+            thumb = thumbs.get(label)
+            # ``thumb.width`` stores the requested spec width (see the generator), so
+            # strict equality works; legacy encoder-width rows read stale and self-heal.
+            width_match = thumb is not None and thumb.width == spec["width"]
+            warm = thumb is not None and thumb.path and width_match and not thumb.is_source_changed
```
loppear: So this "warm" (thumb exists and is valid) check is implemented here (depending on a new annotation to the queryset) and in `SourceImage.find_or_generate_thumbnail_for_label`. In both cases we have already queried/loaded a source image and thumb (or None) objects, my inclination would be to extract the non-annotated version as a helper method e.g. `SourceImage.thumbnail_is_valid(self, spec, thumb)`.

loppear (approved):
Minor simplification/DRY suggested, but overall I agree this change is good and worth including with #1306

mihow:
@coderabbitai full review

ami/main/models.py
mihow: thanks for reviewing again @loppear! I implemented the singular helper as you suggested. I did add an extra guard that should ensure the thumbnails are prefetched and the thumbnail_is_valid check does not trigger n+1 queries. I'd like to integrate Django zen-queries into the project sometime this year to catch introductions of n+1 queries automatically, since that has been the main source of slow load times as the database grows

loppear (approved):
👍

Issue #1337, opened 2026-06-16
Saving job progress concurrently is the root of multiple issues related to incorrect job statuses


mihow:
## Summary

Several open and closed issues about jobs showing the wrong status, finished jobs being marked failed, and the task queue starving all trace back to one mechanism: concurrent result handlers update a job's `progress`/`status` with an unlocked read-modify-write of the whole JSONB blob, so two handlers that read the same snapshot silently lose one of the two updates. This issue is meant to be the single place that names that root cause, separates the bugs it produces into "wrong appearance" vs "actually fails work" (they overlap — see below), and decides on a fix that does not reintroduce the row-lock contention that #1261 removed.

Part of the PSv2 effort (#515).

## Root cause

`_update_job_progress` (`ami/jobs/tasks.py`) does, per result batch:

1. read the `Job` row including the `progress` JSONB (stages, counts, status),
2. mutate it in Python (increment processed/failed counts, recompute stage percentages, possibly set a terminal status),
3. write the whole blob back with `job.save(update_fields=["progress", "status", ...])`.

Under `async_api`, many workers POST results concurrently, so step 1→3 interleaves across processes. Classic lost-update: two handlers read the same snapshot, both write their own version, one increment (or one status transition) is lost. The same unlocked full-object save also happens for the `logs` field and from the task signals.

There was a lock here — `_update_job_progress` previously used `select_for_update`. It was removed in #1261 because under concurrent load every result handler serialized on the single job row, causing slow POSTs, timeouts, and NATS redeliveries (contention analysis in #1256). So **the fix must make these writes atomic without reintroducing a transaction-length row lock** — re-adding `select_for_update` would trade this race straight back for the contention problem.

## Two failure shapes — appearance vs. lost work

The point of this split is to triage: some of these only make the UI lie, some actually destroy or stall work, and a few start as the former and *become* the latter.

### A. Appearance only (work is fine, the status shown is wrong)

- **#721** — job has finished (or errored), but overall status and/or per-stage status still reads `STARTED`; the status bar can't show the real state. This is the most-reported user-facing face of the bug.
- **#1155** — log says "Finished job" when only the first stage finished.
- Progress bar stuck at ~99.9% on a job that actually completed (the lost final-increment interleaving).
- Counts that visibly go up and down during a run (older sighting: #370, closed 2023 — same mechanism).

### B. Actually causes lost / stalled / failed work

[...]

PR #1338, opened 2026-06-16
Stop a finished job from being pulled back to running by a slower worker

mihow:
## Summary

When many ML workers process an async job at once, a job that has actually finished could end up still showing as "running", and a job that was cancelled or revoked could flip back to "succeeded". This happened because every result a worker posted re-saved the job's overall status together with its progress, so a slower worker working from an out-of-date snapshot could overwrite a newer status. Beyond the wrong label, this has a real cost: a job left wrongly in a running state stays claimable by workers through the task endpoint, so workers keep picking it up and newer jobs wait behind it.

This change makes the job's final status transition safe under that concurrency. A worker can no longer regress a job that another worker, the stale-job checker, or a user cancellation has already finished. It is the first, smaller half of the fix described in #1337 (the lost-update race on the job's progress/status); the counter-accumulation half is deferred to a separate design discussion.

### List of Changes

1. **A finished job no longer gets pulled back to "running" by a slow worker.** The progress update now writes the progress detail and the overall status separately, and the status only moves to a terminal value (SUCCESS/FAILURE) from a still-running state — as a single database statement that can't be overwritten by a stale in-memory copy. *Technical:* in `_update_job_progress` (`ami/jobs/tasks.py`), `status`/`finished_at` are dropped from the progress-blob `save(update_fields=...)`; the terminal transition is a guarded `Job.objects.filter(pk=..., status__in=JobState.finalizable_states()).update(status=..., finished_at=...)`. This is statement-scope and holds no transaction-length lock, so it does not reintroduce the row-lock contention removed in #1261.
2. **A cancelled or revoked job is never resurrected into "succeeded".** Late-arriving results that complete the work leave a terminal/CANCELING job untouched. *Technical:* the guarded UPDATE only fires from `finalizable_states()`, which excludes terminal, CANCELING, and UNKNOWN states, so completion can't override them. This is a small, intentional behaviour change versus the previous code, which could flip CANCELING → SUCCESS. It aligns with #1282/#1283.
3. **The set of states a job may finalize from now lives in one place.** *Technical:* extracted `JobState.finalizable_states()` (`[CREATED, PENDING, STARTED, RETRY]` — `running_states()` minus CANCELING and UNKNOWN), with a short note on why those two are excluded, so the guard rule has a single source of truth for the remaining terminal-status writers to reuse.
4. **Regression tests.** `TestConditionalTerminalTransition` covers the guard logic directly: a completing worker does not resurrect a REVOKED job; a normal job still transitions STARTED → SUCCESS with `finished_at` set; a non-terminal progress update never writes the overall status; and a completion does not flip the JSONB `summary.status` of a REVOKED job. `TestConcurrentStatusRace` reproduces the actual lost-update interleave with real threads and database transactions — a slow worker reads the job while STARTED, another writer commits REVOKED, and only then does the slow worker save. This test **fails on the pre-fix code** (the stale save resurrects the job to SUCCESS) and passes with the guarded transition, so it is a genuine regression guard rather than a unit test of the guard in isolation.

## Detailed Description

Since #1261 dropped the `select_for_update` that serialized progress writes (it was a contention bottleneck — see #1256), concurrent result handlers do an unlocked read-modify-write of the `Job` row. Two handlers reading the same snapshot and both saving the whole blob means one update is lost. The most damaging case is the status field: a stale `STARTED` overwriting a freshly-written terminal status. That both misreports the job and — because status drives claimability at the worker task endpoint — keeps a finished job in rotation, starving newer async jobs (the same shape as #1282).

This PR is **Layer 1** of #1337: it makes the *status* transition atomic and non-regressing without re-adding a transaction-length lock. The *counter* accumulation (processed/failed/detection/classification counts in the progress blob) can still drift by a batch under race — that is **Layer 2**, deferred because the right shape (counter columns vs. a per-task table vs. Redis-as-source vs. an event log) depends on other consumers of per-image job history and needs a dedicated design pass. See #1337 for the plan.

Scope note: this hardens the result-handler status writer. The other terminal-status writers (the cancel path and the Celery task signal handlers) still do unguarded full-row saves; routing them through the same `finalizable_states()` guard is a focused fast-follow on top of this PR. #1324 should also be gated behind this change: its `acks_late` redelivery guard reads `job.status`, which is the field this PR makes reliable.

### How to Test the Changes

- `pytest ami/jobs/tests/test_tasks.py::TestConditionalTerminalTransition` — the four guard cases.
- `pytest ami/jobs/tests/test_tasks.py::TestConcurrentStatusRace` — the real-concurrency race test. Check out the pre-fix commit to see it fail (the job is resurrected to SUCCESS), then this branch to see it pass.
- Full `ami/jobs/tests/test_tasks.py` and `ami/jobs/tests/test_jobs.py` pass locally.

## Checklist

Refs #1337, #1282, #1283

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented completed jobs from being incorrectly “resurrected” or moved to a non-terminal result due to concurrent or late progress updates (including revoked jobs).
  * Improved job progress handling so stage updates only affect completion indicators when the job is successfully transitioned to its final state.

* **Tests**
  * Added coverage for conditional terminal-state transitions.
  * Added a concurrency race test to ensure late worker updates never overwrite an already-terminal job status.

### DIFF
--- a/ami/jobs/models.py
+++ b/ami/jobs/models.py
@@ -94,6 +94,12 @@ class JobState(str, OrderedEnum):
         """States where a job is actively processing and should serve tasks to workers."""
         return [cls.STARTED, cls.RETRY]
 
+    @classmethod
+    def finalizable_states(cls):
+        # running_states() minus CANCELING (don't resurrect a cancel in progress)
+        # and UNKNOWN (never served; shouldn't auto-finalize). See #1337.
+        return [cls.CREATED, cls.PENDING, cls.STARTED, cls.RETRY]
+
 
 def get_status_label(status: JobState, progress: float) -> str:
     """
--- a/ami/jobs/tasks.py
+++ b/ami/jobs/tasks.py
@@ -632,10 +632,9 @@ def _update_job_progress(
             progress=progress_percentage,
             **state_params,
         )
-        if job.progress.is_complete():
-            job.status = complete_state
+        became_complete = job.progress.is_complete()
+        if became_complete:
             job.progress.summary.status = complete_state
-            job.finished_at = datetime.datetime.now()  # Use naive datetime in local time
         job.logger.info(f"Updated job {job_id} progress in stage '{stage}' to {progress_percentage*100}%")
         # Narrow the write to the fields we actually mutated. Without this, a full
         # save() would overwrite `logs` and any other field on the instance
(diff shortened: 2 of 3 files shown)


mihow:
@coderabitai full review

mihow:
Tested on arctia

PR #1339, opened 2026-06-17
Add zoom and pan support in session detail view

annavik:
## Summary

In this PR we make it possible to zoom and pan the capture in the session detail view. This makes it possible to see more details in high res captures. Users can zoom using the mouse wheel or using button controls in UI.

The solution using the library [react-zoom-pan-pinch](https://github.com/BetterTyped/react-zoom-pan-pinch).

## Notes

We now load the original image instead of the medium thumbnail image in the session detail view. Before merge, maybe we should load both, so we can still show something quicker, but still make it possible to see all the details?

## Screenshots

Screenshot 2026-06-17 at 14 43 07
Screenshot 2026-06-17 at 14 43 32

## Summary by CodeRabbit

* **New Features**
  * Added zoom and pan capabilities to capture image viewer, allowing users to magnify and navigate across detailed images.
  * Introduced zoom controls (zoom in, zoom out, reset) with keyboard-accessible buttons and tooltips in the session details view.

* **Style**
  * Updated button styling for settings interface.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -38,6 +38,7 @@
     "react-markdown": "^9.0.1",
     "react-plotly.js": "^2.6.0",
     "react-router-dom": "^6.8.2",
+    "react-zoom-pan-pinch": "^4.0.3",
     "tailwind-merge": "^3.6.0"
   },
   "scripts": {
--- a/ui/src/data-services/models/capture.ts
+++ b/ui/src/data-services/models/capture.ts
@@ -167,17 +167,19 @@ export class Capture {
     return this._capture.event?.name ?? ''
   }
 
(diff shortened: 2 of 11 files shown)


mihow (commented):
This is so cool! It's very smooth and everything works for me

Is it possible to make it so the labels don't get zoomed too? Not a dealbreaker if it's a limitation of how elements are contained, but it was a little surprising! Ha
Image

Also I forgot its easy to test in the deploy preview envs again
https://deploy-preview-1339--antenna-preview.netlify.app/projects/20/sessions/32?capture=46310&occurrence=454140&occurrence=548678

PR #1356, opened 2026-06-30
Stop sending admin emails on every HTTP 500 in production

mihow:
## Summary

In production, every uncaught HTTP 500 triggers Django's default admin-error email, sent synchronously inside the request — one outbound email per error, each holding a worker thread for the duration of the send. During an error storm this becomes an email storm that ties up the worker pool and makes an incident worse. Sentry already captures every error independently, so these emails add no observability. This change disables that handler by default in deployed environments and puts it behind an environment variable, so it can be re-enabled per environment if a non-blocking (async/queued) handler is configured later.

### List of Changes

1. Admin error emails are no longer sent on production 500s by default. Operator-facing effect: an error spike can no longer snowball into a synchronous outbound-email storm that starves the worker pool. (Implementation: gate Django's default `mail_admins` `AdminEmailHandler` behind `DJANGO_ADMIN_ERROR_EMAILS`, default `False`; when off, the `django` logger routes to console only at `INFO` so 4xx/5xx request lines still reach stdout, and no longer propagates to the handler.)
2. Configurable per environment: set `DJANGO_ADMIN_ERROR_EMAILS=True` to re-enable admin error email where wanted. Covers production, demo, and staging (same settings module).

### Notes

- Sentry error capture is unaffected — it hooks the logging system via `LoggingIntegration`, independent of the `LOGGING` handlers list.
- Surfaced during a production incident where the synchronous admin-email path amplified a 500 storm.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated production error logging behavior to better control when Django errors are sent to admin email alerts.
  * Added a new production toggle to route Django logs to the console at info level when admin error emails are disabled, helping keep logging behavior consistent.

### DIFF
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -177,6 +177,17 @@ LOGGING = {
     },
 }
 
+# Django's default mail_admins handler sends email synchronously, blocking a worker
+# thread per HTTP 500; other (async/queued) handlers are recommended. Disabled by
+# default and gated behind an env var; errors are already captured by Sentry.
+ADMIN_ERROR_EMAILS = env.bool("DJANGO_ADMIN_ERROR_EMAILS", default=False)
+if not ADMIN_ERROR_EMAILS:
+    LOGGING["loggers"]["django"] = {
+        "level": "INFO",
+        "handlers": ["console"],
+        "propagate": False,
+    }
+
 # Sentry
 # ------------------------------------------------------------------------------
 SENTRY_DSN = env("SENTRY_DSN")
--- a/config/settings/production.py
+++ b/config/settings/production.py
@@ -187,6 +187,11 @@ if not ADMIN_ERROR_EMAILS:
         "handlers": ["console"],
         "propagate": False,
     }
+else:
+    # Nothing to wire here: Django's DEFAULT_LOGGING attaches the mail_admins handler
+    # to the "django" logger before this module's LOGGING is applied. Explicit branch
+    # so the "on" path is visible here instead of hidden in Django's defaults. See #1356.
+    pass
 
 # Sentry
 # ------------------------------------------------------------------------------


config/settings/production.py
```diff
@@ -177,6 +177,17 @@
...
+# Django's default mail_admins handler sends email synchronously, blocking a worker
+# thread per HTTP 500; other (async/queued) handlers are recommended. Disabled by
+# default and gated behind an env var; errors are already captured by Sentry.
+ADMIN_ERROR_EMAILS = env.bool("DJANGO_ADMIN_ERROR_EMAILS", default=False)
+if not ADMIN_ERROR_EMAILS:
+    LOGGING["loggers"]["django"] = {
+        "level": "INFO",
+        "handlers": ["console"],
+        "propagate": False,
+    }
```
mihow, line 189: The `else` (on) branch is implicit today: `DJANGO_ADMIN_ERROR_EMAILS=True` just falls through and inherits Django's `DEFAULT_LOGGING` `mail_admins` handler. That works, but it depends on Django internals and isn't visible at this line. Minimal fix — make the on-path explicit:

```suggestion
if not ADMIN_ERROR_EMAILS:
    LOGGING["loggers"]["django"] = {
        "level": "INFO",
        "handlers": ["console"],
        "propagate": False,
    }
else:
    # Admin error emails are on. Nothing here wires them: they come from Django's
    # DEFAULT_LOGGING, which attaches the mail_admins handler to the "django" logger
    # before this module's LOGGING is applied. Kept as an explicit branch so the
    # "on" behaviour is visible at this line rather than hidden in Django's defaults.
    pass
```

If you'd rather not depend on Django's defaults at all, the more robust version defines the `mail_admins` handler and the `django` logger explicitly in the `else` branch, so the on behaviour is fully self-contained in this file. Happy to draft that instead — this minimal version matches the intent without expanding scope.

mihow (commented):
**Claude says:** (takeaway review, drafted by Claude on Michael's behalf.)

The fix is correct and safe to ship. I traced it in a running container rather than from the diff alone, and the mechanism holds up:

- The admin error emails aren't wired anywhere in the repo — they come from Django's `DEFAULT_LOGGING`, which `configure_logging()` applies *before* this module's `LOGGING`. Both use `disable_existing_loggers: False`, so the `django` logger silently keeps its default `mail_admins` handler unless something redefines it. This PR redefines it, which is exactly the right lever, and explains why grepping for `mail_admins` finds nothing while the emails still send.
- No loss of error visibility. `django.request` (where unhandled 500s are logged) has no handlers of its own and propagates up into the `django` logger, so it still reaches the console. Sentry is entirely unaffected: `LoggingIntegration` patches `logging.Logger.callHandlers` upstream of handler dispatch, so it captures the error regardless of the handler list. Console logging is preserved, and there's no double-logging.

One structural point worth addressing before merge (inline suggestion below): the **"on" path is implicit**. When `DJANGO_ADMIN_ERROR_EMAILS=True`, the code does nothing and inherits `mail_admins` from Django's internals — so "on" works only as long as Django's `DEFAULT_LOGGING` keeps attaching that handler, and nobody reading `production.py` can see where the emails come from. Making it an explicit branch costs a few lines.

Nothing else needs changing — the fix does what it says.

config/settings/production.py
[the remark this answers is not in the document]
mihow: **Claude says:** Verified against this project's Django (4.2.10) rather than from memory — this claim describes pre-1.9 Django. Since Django 1.9, `DEFAULT_LOGGING` attaches the `mail_admins` handler to the `django` logger, not `django.request`:

```python
>>> from django.utils.log import DEFAULT_LOGGING
>>> DEFAULT_LOGGING['loggers']
{'django': {'handlers': ['console', 'mail_admins'], 'level': 'INFO'},
 'django.server': {'handlers': ['django.server'], 'level': 'INFO', 'propagate': False}}
```

`django.request` is not defined in `DEFAULT_LOGGING` at all — records logged to it propagate up to the `django` logger, which is exactly the logger this PR overrides. Because both configs use `disable_existing_loggers: False`, redefining `LOGGING["loggers"]["django"]` replaces its handler list, dropping `mail_admins` while keeping console output. The change is not a no-op; this was also traced in a running container (see the review summary on this PR).

mihow: **Claude says:** Applied in 8ccbc17f. Kept the explicit `else` branch, with the comment trimmed to three lines plus a `See #1356` link per the repo's comment-length convention — the full reasoning stays in this thread and the review summary.

Issue #1364, opened 2026-07-02
Let users build a project taxa list from a region, so class masking works out of the box


mihow:
**Related:** #999 (class masking), #1289 (post-processing framework), #1094 (configurable taxa lists), #939 (import taxa from external lists), #1020 (closed — GBIF taxon select in the UI), #1293 (taxa list CSV export)

## Summary

Class masking (#999) lets an operator keep only the classifier predictions that fall inside a chosen taxa list, which cuts a global classifier down to the species that actually occur at a site. The feature is nearly ready, and the taxa list editor is merged (#1094). The missing piece is the taxa list itself: today someone has to hand-curate one taxon at a time, which is too tedious to expect from project owners, and nothing connects a list to the place a project monitors.

This ticket proposes a way for a user to generate a regional species list automatically from an external biodiversity database (GBIF and/or iNaturalist) by giving a region code, so that curating a project's masking list becomes a single action instead of hundreds. It also proposes where region codes and taxa lists should live on the data model (on the `Site`, with a fall-back on the `Project`), and how the first class-masking pipeline can pick the right list automatically from the occurrence's site rather than making the operator choose one every time.

The core capability — "given a region, produce a taxa list" — is written once as a service function and surfaced through a management command, a Django admin action, an API endpoint, and the main UI, so the same logic can also backfill regional lists for every project already in Antenna.

We should ship the simplest useful version first. My recommendation is Proposal A (generate from a region code) over Proposal B (spreadsheet import with a matching UI); B is more flexible but much more work, and A covers the common case.

---

## Background — current state of the code

Grounded in the current tree so the proposals below reference real names.

**Models**
- There is a `Site` model (verbose name "Research Site", `ami/main/models.py:654`) that groups deployments. It has `name`, `description`, `project` only — **no coordinates, region, or country**. Its geographic extent is derived on the fly from its deployments via `boundary_rect()` (`models.py:670`).
- Coordinates live on `Deployment` (`models.py:763`): `latitude` (`:770`) and `longitude` (`:771`). Deployment has **no `country` or `region` field** either.
- `Occurrence` → site chain: `Occurrence.deployment` (`models.py:3598`) → `Deployment.research_site` (`models.py:803`) → `Site`. `Occurrence` also carries a direct `project` FK (`models.py:3599`).
[...]

PR #1366, opened 2026-07-02
Add the implementation plan for building project taxa lists from a region

mihow:
## Summary

Class masking (#999) can cut a global classifier down to the species that actually occur at a site, but only if someone first curates a taxa list for the project — and today that means adding taxa one at a time, which is too tedious to expect of project owners. This PR does not add any code; it adds the implementation plan for the approach proposed in #1364 ("Proposal A"): let a user generate a regional species list automatically from an external biodiversity database (GBIF and/or iNaturalist) by giving a region code, so building a masking list becomes a single action.

The point of opening this as a draft is to get agreement on the design before implementation starts — in particular the two decisions that shape everything downstream: (1) when multiple sources are used they are combined as a wide **union** with per-source provenance, never intersected against each other; and (2) the list is, by default, restricted to species the classifiers can actually predict, with a stored "model coverage" relationship so the UI can be honest about regional species that no model will ever predict (many valid species lack training data). The same core service is designed to be reused by a management command, a Django admin action, an API endpoint, the main UI, and unit tests — which also lets us backfill regional lists for every existing project.

No migrations, models, or endpoints are included yet. This is the plan; the code lands in the phased slices it describes, starting with a Phase 0 spike to verify the external APIs (none of the GBIF/iNaturalist endpoints have been exercised against the live services — they are flagged CANDIDATE/UNVERIFIED throughout).

Planning for #1364. Design writeup: `docs/claude/planning/2026-07-02-regional-taxa-lists-class-masking.md`.

### List of Changes

| # | Change (what it adds) | Detail |
|---|---|---|
| 1 | Adds the Proposal A implementation plan document | `docs/claude/planning/2026-07-02-proposal-a-regional-taxa-lists-impl-plan.md` — 14 sections, docs-only, no executable change |
| 2 | Specifies the reusable core service | `generate_regional_taxa_list(...)` with a `Result` breakdown, surfaced through five thin wrappers (command, admin action, API, UI, tests) |
| 3 | Specifies the multi-source design | A `RegionalSpeciesSource` protocol + a wide **union** merge that keeps per-source provenance; sources are never intersected with each other |
| 4 | Specifies model & DB awareness | Default list = regional species ∩ model-covered taxa; a stored `Taxon.covered_by_algorithms` relationship (+ a denormalized `has_model_coverage` flag) plus a refresh path, so the UI can flag regional species no model can predict |
| 5 | Specifies the data-model changes | New `region_source` / `region_code` / taxa-list fields on `Site` and `Project`, plus the coverage relationship — each with its migration, called out but not implemented here |
| 6 | Specifies auto-apply masking | A `taxa_list_mode="auto"` resolution ladder (occurrence → site → project) that no-ops until a region/list is configured; lands on the #999 branch since `class_masking.py` is not on `main` yet |
| 7 | Lays out a TDD test plan and phased rollout | Ten tests written first; internal-first phases (spike → core service → command/admin → masking wiring → API/UI), with an explicit "what to verify before building" list |

The branch is docs-only. CI is skipped on these commits (`[skip ci]`). Nothing outside `docs/claude/planning/` is touched.


mihow:
**Phase 0 (de-risk spike) ran — verdict: GO.**

Measured against the real 2497-label Quebec & Vermont classifier, a Vermont region list (GBIF ∪ iNat) covers **70.0% (1749/2497)** of the classifier's labels, so the default masking list would keep 1749 classes and mask ~748 (30%) that neither source records in Vermont. GBIF alone reaches 69.8%; iNat adds little to the intersection but feeds the 550-species `include_uncovered` bucket — so Phase 1 starts GBIF-first. The A3 reverse-geocode path resolved `(44.26,-72.58)` → GADM1 `USA.46_1` cleanly, so the `--all-projects` backfill is viable.

Top risk to carry into Phase 1: name-join fragility — 30% of labels are absent from the region union, a mix of true regional absences and likely name/synonym mismatches that needs a sample audit. Full numbers, caveats, and the reproducible script are in `docs/claude/analysis/` on this branch (commit 574f9e30).

PR #1367, opened 2026-07-03
Generate a project taxa list from a region, and track which species the models can predict

mihow:
## Summary

Class masking (#999) can cut a global classifier down to the species that actually occur where a project monitors, but only if someone first curates a taxa list — and doing that by hand, one taxon at a time, is too tedious to expect of project owners. This PR lets a project build that list automatically from a geographic region: give a region (or let it be derived from the project's deployments) and it pulls the species recorded there from an external biodiversity database, keeps the ones a classifier can actually predict, and saves them as the project's taxa list. Class masking can then resolve that list on its own.

The same core capability is reachable five ways — a management command (including an `--all-projects` backfill), Django admin actions, a REST endpoint, and unit tests all call one service — so we can also generate regional lists for every existing project in one pass. This is the backend and API; the in-app UI button is the remaining piece (see the checklist below).

Two design decisions are worth a reviewer's attention:

1. **Multiple sources union, they never intersect.** A species present in *any* source (GBIF today, iNaturalist next behind the same protocol) is a candidate; the merge is a wide union that preserves per-source provenance. Sources are never intersected against each other.
2. **The list is honest about what the models can predict.** By default it is restricted to species some classifier can output — tracked by a new, persisted `Taxon ↔ Algorithm` relationship. An opt-in mode also keeps regional species no model can predict, flagging each so the UI can say so.

Phase 0 measured this against the real 2497-label Quebec & Vermont classifier: a Vermont region list covers 70% of its labels, so the default masking list keeps ~1749 classes and drops ~748 that neither GBIF nor iNaturalist records in Vermont (full findings in `docs/claude/analysis/`, verdict: proceed).

Verification: the feature ships with its own tests (44 in `tests_regional_taxa.py` plus 12 for masking auto-mode), and the existing taxonomy/taxa-list/class-masking suites still pass. `makemigrations --check` is clean; black/isort/flake8 pass. No existing behavior changes until a project configures a region.

Planning: #1364. Plan/design PR: #1366.

### List of Changes

| # | Change (effect) | How (implementation) |
|---|---|---|
| 1 | A project can build a taxa list from a region | Core service `ami/main/services/regional_taxa.py`: `generate_regional_taxa_list()`, the wide-union `merge_source_species()`, `map_to_taxa()`, `apply_model_coverage()`, `Result` with per-bucket counts |
| 2 | Regional species come from GBIF (primary source, verified in Phase 0) | `services/gbif.py` — occurrence facet by GADM region + species-name resolution + reverse-geocode; iNaturalist later behind the same protocol |
| 3 | The list knows which species the models can predict | Persisted `Taxon.covered_by_algorithms` (M2M → Algorithm) + `has_model_coverage` flag; `services/taxon_coverage.py` refreshes it from category-map labels, hook keyed on `labels_hash`, targeted refresh for newly created taxa |
| 4 | Operators can generate lists and backfill every project | `generate_regional_taxa_list` management command with `--all-projects` (derives each region from deployments), `--dry-run`, `--include-uncovered`; `refresh_taxon_model_coverage` command |
| 5 | Admins can generate a list from the changelist | Project & Site admin actions that enqueue a background task (the external fetch is slow); region fields exposed in both admins |
| 6 | Class masking can pick the list automatically | `ClassMaskingConfig.taxa_list_mode="auto"` resolves the list from the occurrence's site, then the project's default; a no-op when nothing is configured, so masking is safe to enable by default |
| 7 | The app/API can trigger generation | `POST /projects/{id}/generate-regional-taxa-list/` enqueues the task and returns 202; requires project update permission; region derivable from deployments |
| 8 | A site or project records its region | `region_source` / `region_code` fields on `Site` and `Project`, plus `taxa_list` (Site) / `default_taxa_list` (Project) FKs; migration `0095` |
| 9 | Shared taxonomy building | Extracted the `create_taxon` hierarchy builder from `import_taxa` into `services/taxonomy.py` (no behavior change) |

### Still to do

- [ ] In-app UI button ("Create taxa list for region") in project settings, with screenshots — the remaining phase.
- [ ] iNaturalist source (second source behind the existing protocol).
- [ ] Name-mismatch audit of the ~30% of classifier labels not attested in-region (the top risk flagged in Phase 0), to decide how much fuzzy matching to add.


mihow:
Filed #1405: the name-only join between a classifier label and a `Taxon` (in `with_taxa()`, which this PR's default masking list ultimately feeds) never follows `synonym_of`, so some of the region-attested-but-not-in-classifier gap this PR's Phase 0 spike measured (the 30% uncovered bucket) is plausibly synonym mismatch rather than genuine absence. Worth resolving, or at least measuring the overlap, before leaning on the name join at scale for the `--all-projects` backfill.

PR #1368, opened 2026-07-09
Filter occurrences by the algorithms that actually ran, and make class masking safe to re-run

mihow:
## Summary

Follow-up to #999 (class masking). Two rough edges surfaced while running class masking on a project in a production deployment.

First, the occurrence list could not be filtered down to what a given algorithm actually produced. The filter's choices came from pipeline configuration, so a standalone post-processing algorithm such as a class-masked classifier — which has no pipeline — never appeared, and neither did a superseded model version whose determinations still exist. The filter itself also had a counting bug that inflated result counts, and filtering by a detector returned nothing at all.

Second, class masking was only safe to run once per capture set: if a source classification became terminal again after a later dedup or re-classification pass, the source could be re-scored and pick up a duplicate masked classification.

This PR addresses both. The occurrence algorithm filter now offers exactly the algorithms that produced results in the project — every classifier, detector, and post-processing algorithm whose output still exists — served from a new endpoint, `/occurrences/algorithms/`. The project's algorithms page is unchanged from `main`: it keeps showing the algorithms on the project's enabled pipelines, so a freshly configured project still sees what it can run before any job has completed. Class masking is now idempotent: a source is masked at most once per masking algorithm.

No schema change and no migration. Every fix is query-level.

### List of Changes

| Change (user-facing effect) | How (implementation) |
| --- | --- |
| The occurrence algorithm filter offers exactly the algorithms that produced results in the project — including older/superseded versions, detectors, and standalone post-processing algorithms — so it never lists a choice with zero results. | New `OccurrenceViewSet` sub-action at `/api/v2/occurrences/algorithms/`, backed by a reusable queryset method `Algorithm.objects.used_in_project()`: the algorithm ids owning classifications (classifiers, post-processing) or detections (detectors) whose capture belongs to the project. Requires `project_id` and applies the project visibility check. |
| Filtering occurrences by an algorithm now matches detectors too, and the result count is no longer inflated when an occurrence has several matching classifications. | New `OccurrenceQuerySet.processed_by_algorithm()` / `not_processed_by_algorithm()`, backed by an `EXISTS` subquery matching a detection's `detection_algorithm` or a classification's `algorithm`. `OccurrenceAlgorithmFilter` routes through them. The name covers post-processing algorithms (a size filter neither detects nor classifies, but its classifications match here); `processed_by_pipeline` and `processed_by_job` can follow the same family later. |
| The project's algorithms page behaves as on `main`: it lists the algorithms on the project's enabled pipelines, so a new project sees its available algorithms before anything has run. | `AlgorithmViewSet.get_queryset` keeps main's enabled-pipeline filter; this PR only adds a comment pointing to the new choices endpoint. |
| Clicking an algorithm name on an occurrence's classification still opens its detail page (and category map) even when that algorithm is not on an enabled pipeline. | The detail endpoint is unscoped by design; this PR pins that with a regression test. |
| Re-running class masking (to finish an interrupted run, or after a dedup / re-classification pass) no longer creates duplicate masked classifications. | `_scoped_classifications` excludes sources that already have a masked child for the same masking algorithm, keyed on the `applied_to` lineage rather than the mutable `terminal` flag. |

### Two endpoints, two purposes

The project algorithm list and the occurrence filter choices sound alike but answer different questions, so they are served separately:

- **`/ml/algorithms/?project_id=…`** (the algorithms page) answers *"what can this project run?"* — the algorithms on its enabled pipelines. Configuration-driven, non-empty for a freshly configured project.
- **`/occurrences/algorithms/?project_id=…`** (the filter choices) answers *"what has produced results here?"* — anything whose output still exists, regardless of pipeline configuration. Results-driven, so filtering by any choice returns at least one occurrence.

An earlier revision of this PR merged the two into one used-only list with historical entries grayed out. That left a new project's algorithms page empty until its first run, which is the wrong page to be empty — the user is there to decide what to run. Splitting the consumers resolved it and removed the extra `enabled_in_project` flag the merged design needed.

A follow-up (planned alongside a denormalised `project` column on classifications/detections) can make the algorithms page all-inclusive — configured ∪ used, with per-algorithm usage counts and links into filtered occurrences.

### The occurrence filter counting bug

`OccurrenceAlgorithmFilter` previously filtered with `detections__classifications__algorithm__in=…`. That is a multi-valued join, so an occurrence with three matching classifications became three rows, and with no `.distinct()` downstream the paginator's `COUNT` counted it three times. On the largest project, filtering by one classifier reported **111,272** occurrences where there are **55,214**. The same join matched only classifiers — a detector authors no classification, so filtering by a detector returned nothing at all. The `EXISTS` form counts each occurrence once and matches both roles.

## What was measured

All figures are from a local copy of production (~834k classifications platform-wide), on the largest project unless noted. They are indicative — a working database, not an isolated benchmark rig — so read them as order-of-magnitude, not precise. Timing loops ran with `cachalot_disabled()` and rebuilt the queryset each iteration (a reused queryset serves its result cache and reports a false ~0 ms).

**Filter-choices query shapes** (median wall-clock for the project-scoped used-only list):

| Shape | Median | Why it was kept or rejected |
| --- | --- | --- |
| Pipeline-only join (the algorithms page, unchanged) | ~8 ms | Index scan, but excludes superseded versions and post-processing algorithms — not "used". |
| Hybrid: pipeline join + restricted post-processing lookup | ~12 ms | An earlier revision here; fast, but still cannot express "used" (drops disabled-pipeline versions and pipeline-less detectors). |
| **True used-only (`used_in_project`, this PR)** | **~0.1–0.6 s cold** | Correct definition. Scans classification and detection because neither has a project column; cachalot-warm after the first load. Served from `/occurrences/algorithms/`, which the UI requests when the filter panel opens — not on page load. |
| Authorship-only, no candidate restriction | ~595 ms | An earlier wrong revision; also silently dropped every detector. |
| Per-algorithm `EXISTS` loop | ~941–1988 ms | Rejected. |
| Pipeline `OR` used, one query | ~2206 ms | Rejected — `COUNT(DISTINCT)` over the multi-valued join dominates. |

**EXPLAIN (ANALYZE) findings:**

- The two-step form (materialize candidate ids, then `algorithm_id IN (...)`) plans as an **index scan (~6 ms)**. Written as one join, the `pipelines__isnull=True` anti-join hides the ids at plan time and forces a **parallel sequential scan of the whole classification table (~48 ms)**, whose cost grows with the platform's total classification count. This is why `used_in_project` materializes and sorts its id lists.
- Neither `Classification` nor `Detection` has a `project` column — project is reachable only through `source_image` — so no index reaches the project directly. That is the structural reason the honest "used" query scans, and the reason its cost tracks total table size rather than a project's own volume.

**Occurrence filter correctness** (largest project, one classifier id):

| | Old join (`detections__classifications__algorithm__in`) | New `EXISTS` (`processed_by_algorithm`) |
| --- | --- | --- |
| COUNT reported | 111,272 (inflated) | 55,214 (correct) |
| Filter by a detector id | 0 results | matches the detector's occurrences |

The measurement discipline behind all of this is now recorded in `AGENTS.md`: run `EXPLAIN (ANALYZE)` before claiming a plan, and measure on the largest project rather than a small one. An earlier revision of this PR claimed a plan it had not EXPLAINed and a speedup measured against the wrong baseline; both were corrected, and the rule is written down so the next change does not repeat it.

## Tests

All green locally (CI-compose against the prod-copy database) and on the pushed head in GitHub CI.

- `TestAlgorithmViewSetProjectFilter` (5) — the project algorithms page lists exactly the enabled pipelines' algorithms, including one that has never run; a disabled pipeline's algorithm is not listed; other projects are isolated; the unscoped list is unchanged; the detail endpoint resolves an algorithm outside the project's enabled set (the occurrence → algorithm name → category map link).
- `TestOccurrenceAlgorithmChoices` (8) — `/occurrences/algorithms/` returns exactly the algorithms with results in the project; a configured-but-never-run algorithm is not a choice; a superseded version whose determinations survive is; a detector that only ever produced detections is; classifications in another project do not leak; the underlying lookup deduplicates in SQL; `project_id` is required (400 without it); the response is the standard paginated `{count, results}` shape the frontend's entity picker consumes.
- `TestOccurrenceAlgorithmFilterQuerySet` (5) — filtering by a classifier and by a detector each match the right occurrences; an occurrence with several matching classifications counts once (the counting bug above); exclusion is the strict complement of inclusion.
- Class-masking suite (14, including the scope-shape tests merged from #1376) — including `test_rerun_does_not_duplicate_masked_classifications`: running the same mask twice, including the case where the source became terminal again between runs, yields exactly one masked classification. The #1376 scope-shape tests were updated for the idempotency guard's extra argument during the rebase.
- Frontend: `tsc --noEmit`, ESLint, and Prettier pass on the changed files.

## Notes / follow-ups

- **Denormalised project column** on `Classification` / `Detection` is the durable fix for the cold cost of the used-only query, and the prerequisite for the all-inclusive algorithms page (configured ∪ used, with usage counts). Until then the choices endpoint pays a table scan on cache invalidation.
- An algorithm that produced neither classifications nor detections in the project is not a filter choice — correct under "used".
- The masking scope still filters on `scores__isnull=False`. As the `scores` field is retired in favour of logits-only masking, that filter should be revisited so logits-only classifications are not skipped.
- The idempotency guard is keyed on `(source classification, masking algorithm)`, and the masking algorithm's identity includes the taxa list by primary key rather than by contents. Editing a taxa list in place and re-running the same mask will skip already-masked sources and leave them scored against the old list; create a new taxa list to re-mask.
- Running a *different* taxa list over an already-masked collection applies only to the sources the first mask left terminal, and reports that smaller count as if it were the whole collection. This predates the PR but sits close to the re-run safety it claims.

Closes the filterability and re-run gaps identified after #999.

### DIFF
--- a/ami/ml/views.py
+++ b/ami/ml/views.py
@@ -15,7 +15,7 @@ from ami.base.permissions import ProjectPipelineConfigPermission
 from ami.base.views import ProjectMixin
 from ami.main.api.schemas import project_id_doc_param
 from ami.main.api.views import DefaultViewSet
-from ami.main.models import Project, SourceImage
+from ami.main.models import Classification, Project, SourceImage
 from ami.ml.schemas import PipelineRegistrationResponse
 
 from .models.algorithm import Algorithm, AlgorithmCategoryMap
@@ -64,44 +64,44 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
                 # Algorithms reach the list two ways. An enabled pipeline configures most of
                 # them, and that join alone covers detectors, which never author a
                 # Classification and so cannot be found by their results at all.
-                #
-                # Post-processing algorithms (e.g. class masking) are created standalone with
-                # no pipeline, so they are found by their classifications instead. Restricting
-                # that lookup to unpipelined algorithms bounds which rows are scanned:
-                # Classification has no project column and reaches one only through
-                # detection -> source_image, so an unrestricted version scans the whole table.
-                #
-                # An algorithm attached to a pipeline that is not enabled for this project is
-                # therefore absent even when it owns determinations here. That matches the
-                # behaviour on the pipeline join alone, and widening the second query to cover
-                # it costs roughly five times the runtime.
-                #
-                # The two are collected separately rather than OR'd into one filter. An OR
-                # across the pipeline join forces a SELECT DISTINCT whose COUNT costs more
-                # than both queries together.
                 configured_for_project = Algorithm.objects.filter(
                     pipelines__project_pipeline_configs__project=project,
                     pipelines__project_pipeline_configs__enabled=True,
                 ).values_list("pk", flat=True)
-                # Deduplicated in the database: the join emits one row per matching
-                # classification, so without this the result grows with a project's masked
-                # classification count rather than with its handful of algorithms.
+
+                # Post-processing algorithms (e.g. class masking) are created standalone with
+                # no pipeline, so they are found by their classifications instead. This is done
+                # in two steps on purpose. Collecting the unpipelined algorithm ids first, as an
+                # explicit list, lets the classification lookup filter on `algorithm_id IN (...)`,
+                # which the planner answers from the algorithm_id index. Expressing the same thing
+                # as a single join (`pipelines__isnull=True, classifications__...`) hides those ids
+                # behind an anti-join at plan time, so the planner sequentially scans the whole
+                # classification table instead — a cost that grows with the platform's total
+                # classification count rather than with this project. The first query is also
+                # cheap to cache: it only changes when algorithms or pipeline links change, not on
+                # every classification write.
+                #
+                # An algorithm attached to a pipeline that is not enabled for this project is
+                # absent even when it owns determinations here. That matches the behaviour on the
+                # pipeline join alone; widening the second lookup to cover it costs several times
+                # the runtime and is left as a follow-up.
+                unpipelined_algorithm_ids = list(
+                    Algorithm.objects.filter(pipelines__isnull=True).values_list("pk", flat=True)
+                )
+                # `.order_by()` clears Classification's default ordering before `.distinct()`.
+                # Without it the ordering columns join the SELECT and widen the DISTINCT, so it
+                # would return one row per classification rather than one per algorithm.
                 post_processing_used_in_project = (
-                    Algorithm.objects.filter(
-                        pipelines__isnull=True,
-                        classifications__detection__source_image__project=project,
+                    Classification.objects.filter(
+                        algorithm_id__in=unpipelined_algorithm_ids,
+                        detection__source_image__project=project,
                     )
-                    .values_list("pk", flat=True)
+                    .order_by()
+                    .values_list("algorithm_id", flat=True)
                     .distinct()
                 )
                 # Sorted so the generated SQL is stable for a given result: cachalot keys its
                 # cache on the query string, and an unordered set would vary it.
-                #
-                # Materialising the ids is fine at this cardinality. Algorithms are created per
-                # model rather than per run, and the one path that adds them over time is class
-                # masking, which creates a single algorithm per source algorithm, taxa list and
-                # reweight mode. Should that ever reach the thousands, this wants to become a
-                # subquery rather than an IN list.
                 relevant_ids = set(configured_for_project) | set(post_processing_used_in_project)
                 qs = qs.filter(pk__in=sorted(relevant_ids))
         return qs
--- a/ami/ml/views.py
+++ b/ami/ml/views.py
@@ -15,7 +15,7 @@ from ami.base.permissions import ProjectPipelineConfigPermission
 from ami.base.views import ProjectMixin
 from ami.main.api.schemas import project_id_doc_param
 from ami.main.api.views import DefaultViewSet
-from ami.main.models import Classification, Project, SourceImage
+from ami.main.models import Project, SourceImage
 from ami.ml.schemas import PipelineRegistrationResponse
 
 from .models.algorithm import Algorithm, AlgorithmCategoryMap
@@ -61,53 +61,12 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
         if getattr(self, "action", None) == "list":
             project = self.get_active_project()
             if project:
-                # Algorithms reach the list two ways. An enabled pipeline configures most of
-                # them, and that join alone covers detectors, which never author a
-                # Classification and so cannot be found by their results at all.
-                configured_for_project = Algorithm.objects.filter(
-                    pipelines__project_pipeline_configs__project=project,
-                    pipelines__project_pipeline_configs__enabled=True,
-                ).values_list("pk", flat=True)
-
-                # Post-processing algorithms (e.g. class masking) are created standalone with
-                # no pipeline, so they are found by their classifications instead. This is done
-                # in two steps on purpose. Collecting the unpipelined algorithm ids first, as an
-                # explicit list, lets the classification lookup filter on `algorithm_id IN (...)`,
-                # which the planner answers from the algorithm_id index. Expressing the same thing
-                # as a single join (`pipelines__isnull=True, classifications__...`) hides those ids
-                # behind an anti-join at plan time, so the planner sequentially scans the whole
-                # classification table instead — a cost that grows with the platform's total
-                # classification count rather than with this project. The first query is also
-                # cheap to cache: it only changes when algorithms or pipeline links change, not on
-                # every classification write.
-                #
-                # An algorithm attached to a pipeline that is not enabled for this project is
-                # absent even when it owns determinations here. That matches the behaviour on the
-                # pipeline join alone; widening the second lookup to cover it costs several times
-                # the runtime and is left as a follow-up.
-                # Sorted for the same reason as the outer id list below: this list is rendered
-                # verbatim into the IN clause, and Algorithm's (name, version) ordering would
-                # otherwise vary the id order — and so the SQL string and its cachalot key —
-                # for identical membership.
-                unpipelined_algorithm_ids = sorted(
-                    Algorithm.objects.filter(pipelines__isnull=True).values_list("pk", flat=True)
-                )
-                # `.order_by()` clears Classification's default ordering before `.distinct()`.
-                # Without it the ordering columns join the SELECT and widen the DISTINCT, so it
-                # would return one row per classification rather than one per algorithm.
-                post_processing_used_in_project = (
-                    Classification.objects.filter(
-                        algorithm_id__in=unpipelined_algorithm_ids,
-                        detection__source_image__project=project,
-                    )
-                    .order_by()
-                    .values_list("algorithm_id", flat=True)
-                    .distinct()
-                )
-                # Sorted so the generated SQL is stable for a given result: cachalot keys its
-                # cache on the query string, and an unordered set would vary it.
-                relevant_ids = set(configured_for_project) | set(post_processing_used_in_project)
-                qs = qs.filter(pk__in=sorted(relevant_ids))
+                # Scope the list to algorithms that actually produced results in the project —
+                # any superseded pipeline version or standalone post-processing algorithm whose
+                # output still exists, so the user can filter occurrences by anything that ran.
+                # Pipeline configuration is not consulted; configured-but-never-run algorithms
+                # do not appear. See the method for cost characteristics.
+                qs = qs.used_in_project(project)  # type: ignore[union-attr] # Custom queryset method
         return qs
 
     @extend_schema(parameters=[project_id_doc_param])
--- a/ami/ml/views.py
+++ b/ami/ml/views.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.db import transaction
-from django.db.models import Prefetch
+from django.db.models import Prefetch, Q
 from django.db.models.query import QuerySet
 from django.utils.text import slugify
 from drf_spectacular.utils import extend_schema
@@ -15,7 +15,7 @@ from ami.base.permissions import ProjectPipelineConfigPermission
 from ami.base.views import ProjectMixin
 from ami.main.api.schemas import project_id_doc_param
 from ami.main.api.views import DefaultViewSet
-from ami.main.models import Project, SourceImage
+from ami.main.models import Classification, Project, SourceImage
 from ami.ml.schemas import PipelineRegistrationResponse
 
 from .models.algorithm import Algorithm, AlgorithmCategoryMap
@@ -56,14 +56,25 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
     def get_queryset(self) -> QuerySet["Algorithm"]:
         qs: QuerySet["Algorithm"] = super().get_queryset()
         qs = qs.with_category_count()  # type: ignore[union-attr] # Custom queryset method
-        # Only scope list by project. Detail stays unscoped so links from historical
+        # Only scope the list by project. Detail stays unscoped so links from historical
         # classifications whose pipeline is no longer enabled still resolve.
         if getattr(self, "action", None) == "list":
             project = self.get_active_project()
             if project:
+                # An algorithm is relevant to the project if it is configured via an
+                # enabled pipeline OR it produced classifications in the project.
+                # Post-processing algorithms (e.g. class masking) are created standalone
+                # with no pipeline, so the pipeline join alone would hide them even though
+                # they own determinations the user needs to filter occurrences by.
+                classified_in_project = Classification.objects.filter(detection__source_image__project=project).values(
+                    "algorithm"
+                )
                 qs = qs.filter(
-                    pipelines__project_pipeline_configs__project=project,
-                    pipelines__project_pipeline_configs__enabled=True,
+                    Q(
+                        pipelines__project_pipeline_configs__project=project,
+                        pipelines__project_pipeline_configs__enabled=True,
+                    )
+                    | Q(pk__in=classified_in_project)
                 ).distinct()
         return qs
 
--- a/ami/ml/views.py
+++ b/ami/ml/views.py
@@ -15,7 +15,7 @@ from ami.base.permissions import ProjectPipelineConfigPermission
 from ami.base.views import ProjectMixin
 from ami.main.api.schemas import project_id_doc_param
 from ami.main.api.views import DefaultViewSet
-from ami.main.models import Classification, Project, SourceImage
+from ami.main.models import Project, SourceImage
 from ami.ml.schemas import PipelineRegistrationResponse
 
 from .models.algorithm import Algorithm, AlgorithmCategoryMap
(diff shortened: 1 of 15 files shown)


ami/ml/views.py
```diff
@@ -56,14 +56,25 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
...
             project = self.get_active_project()
             if project:
+                # An algorithm is relevant to the project if it is configured via an
+                # enabled pipeline OR it produced classifications in the project.
+                # Post-processing algorithms (e.g. class masking) are created standalone
+                # with no pipeline, so the pipeline join alone would hide them even though
+                # they own determinations the user needs to filter occurrences by.
+                classified_in_project = Classification.objects.filter(detection__source_image__project=project).values(
+                    "algorithm"
+                )
```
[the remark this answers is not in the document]
mihow: disagree, we want to find any occurrence that used intermediate algorithms as well.

ami/ml/views.py
```diff
@@ -56,14 +56,25 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
     def get_queryset(self) -> QuerySet["Algorithm"]:
         qs: QuerySet["Algorithm"] = super().get_queryset()
         qs = qs.with_category_count()  # type: ignore[union-attr] # Custom queryset method
-        # Only scope list by project. Detail stays unscoped so links from historical
+        # Only scope the list by project. Detail stays unscoped so links from historical
         # classifications whose pipeline is no longer enabled still resolve.
         if getattr(self, "action", None) == "list":
             project = self.get_active_project()
             if project:
+                # An algorithm is relevant to the project if it is configured via an
```
mihow: I don't see a reason to list algorithms configured for the project. Only used algorithms makes sense. But it needs to be fast to load.

mihow: **Claude says:** Done in `87a68ada` — the enabled-pipeline branch is gone, the list is now purely "algorithms that produced classifications in this project".

It is also the faster of the two. Measured against a local copy of production data (roughly 834k classifications), on one of the larger projects, hitting the paginated endpoint:

| shape | pagination COUNT | page | total |
| --- | --- | --- | --- |
| pipeline OR used (previous commit) | 1722 ms | 483 ms | **2206 ms** |
| used only (now) | 275 ms | 281 ms | **556 ms** |

Most of the win is in the COUNT. The `OR` across the pipeline join forced a `SELECT DISTINCT`, and counting that distinct set was more expensive than the subquery itself.

One caveat worth naming, since it is not fixed by this change: both shapes still sequentially scan the classification and detection tables. Nothing indexes a path from a project back to a classification — it is two joins away via `detection -> source_image -> project` — so the cost tracks the total size of those tables rather than the size of the project. It stays flat as a project grows but rises as the platform accumulates classifications overall. If it becomes a problem, the fixes are a denormalised `project` on `Classification` (the way `Occurrence` already has one) or a precomputed per-project algorithm list, both bigger than this PR.

Also flagging a side effect for your call: `/projects/<id>/algorithms` is served by this same endpoint, so that page now also hides configured-but-never-run algorithms. That reads right to me for a results-oriented page, but it is a visible change beyond the occurrence filter.

ami/ml/views.py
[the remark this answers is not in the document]
mihow: **Claude says:** Applied in `80b80c2d` — the subquery is now `.values_list("algorithm", flat=True).distinct()`.

You are right about the shape: it was returning one row per classification, so a large project fed hundreds of thousands of rows into the `IN` clause to identify at most a few dozen algorithms.

Being straight about the measured effect, though: it is smaller than the reasoning suggests. Against a local copy of production data (roughly 834k classifications), over four runs of the paginated endpoint, the median went from 610 ms to 567 ms with the ranges overlapping (577-675 vs 543-595). That is close enough to run-to-run noise that I would not claim it as a speedup. Postgres still has to scan the join to produce the distinct set, so deduplicating shrinks what is handed to the outer query but not the work that dominates.

Taking it for clarity — the query now says what it means — rather than for the milliseconds. The change that actually moved the number was dropping the enabled-pipeline branch (2206 ms to 556 ms), which removed an `OR` across a multi-valued join and the `SELECT DISTINCT` it forced.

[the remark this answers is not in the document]
mihow: **Claude says:** Closing this out, agreeing with the maintainer that the subquery should stay unscoped by `terminal`.

Worth recording one more reason beyond intent, because it makes a terminal filter actively wrong here rather than merely narrow. Class masking writes its masked classification with `terminal=True` and demotes the source it was applied to with `terminal=False`, in the same transaction (`ami/ml/post_processing/class_masking.py`, the demote/create batch). So a source classifier that has since had its results re-scored by a mask holds only non-terminal rows. Scoping this subquery to terminal classifications would have hidden exactly those algorithms from the list — the ones whose output a user is most likely to want to compare a masked result against.

The endpoint has also since changed shape: the enabled-pipeline branch is gone entirely, so the list is now purely "algorithms that produced classifications in this project". The concern about intermediate rows from disabled pipelines making an algorithm appear is now the intended behaviour rather than a side effect — having run in the project is the whole membership rule.

mihow:
**Claude says:** Correcting something I got wrong earlier in this thread, and reporting a bug a structural review pass caught in the change I had pushed. Both are now fixed in `0a6dfa0f`, but the record needs setting straight.

### The performance claim I posted was measured against the wrong baseline

I said the change made the list "about four times faster". That compared against an intermediate commit on this branch, not against `main`. Main was never the pipeline-`OR`-used shape — it is the pipeline join alone, which touches only small, rarely-written tables. Measured properly, against a local copy of production data on the largest project:

| shape | median |
| --- | --- |
| `main` today (pipeline join) | **8 ms** |
| authorship-only (what I had pushed) | **595 ms** |

So that commit was roughly a 74x regression, not a speedup. Compounding it, the new query depends on `main_classification`, which cachalot invalidates on every classification insert — continuously during processing, which is exactly when people are looking at these lists. I apologise for the misleading number; it came from benchmarking two variants of my own change against each other and never re-baselining against `main`.

### It also silently dropped every detector

Scoping the list by classification authorship looked equivalent to "algorithms used in this project", but it is not. Localizers set `Detection.detection_algorithm` and never write a `Classification` row (`ami/ml/models/pipeline.py`), so the detector behind every detection in a project vanished from the list. On the largest project that removed a `FasterRCNN` localizer that had genuinely run. The `/projects/<id>/algorithms` page renders a `task_type` column specifically to show detectors alongside classifiers, so this was visible data loss, not a cosmetic change.

### What it does now

The enabled-pipeline join is the primary rule again, and post-processing algorithms are admitted by a second query restricted to algorithms with no pipeline — a handful of candidates, which is what keeps the classification lookup off the full table. The two are collected separately rather than `OR`-ed, because an `OR` across the pipeline join forces the `SELECT DISTINCT` that made the earlier variant slow.

| shape | median | detectors | post-processing algorithms |
| --- | --- | --- | --- |
| `main` today | 8 ms | yes | **no — the bug this PR fixes** |
| authorship-only (reverted) | 595 ms | **no** | yes |
| now | **46 ms** | yes | yes |

On the largest project the result is a strict superset of `main`: nothing `main` listed is lost, and the four post-processing algorithms are added.

### The one thing this gives up, which is your call

It no longer hides algorithms that are configured for the project but never ran — the thing you asked for further up this thread. I went with the pipeline join anyway because your ask was conditional ("it needs to be fast to load"), and hiding them is what costs the 8 ms to 595 ms and the detectors. Hiding unused algorithms cheaply needs a denormalised `project` on `Classification`, the way `Occurrence` already has one; that is a larger change than this PR. If you would rather have the filtering and accept the cost, it is a one-line revert of `0a6dfa0f` — but I would not recommend it in this form.

Also still absent, exactly as on `main`: an algorithm attached to no pipeline that only ever produced detections. Catching it means scanning the detection table.

mihow:
**Claude says:** A structural review pass found two more things in the algorithm-list query. One is fixed in `7d67f5be`; the other is a known gap I am documenting rather than closing here.

### Fixed: the post-processing lookup shipped one row per classification

The join from an algorithm to a project through its classifications emits one row per matching classification, not one per algorithm. So the second query was fetching every masked classification id in the project — 846 rows for 4 algorithms on one project, 1354 rows for a single algorithm on another — only to collapse them into a handful of ids in Python. Harmless at todays volumes (still tens of milliseconds), but the row count was linear in masked-classification count, which is exactly what a production masking run generates. `7d67f5be` adds `.distinct()` so the database returns one id per algorithm. A test now pins the row count of the underlying lookup, since the listed names are correct either way and the bug is invisible in the output.

For the record, the reviewer suggested `Exists()` for this rather than `.distinct()`. I measured both, and `Exists()` is the wrong call here: it ran 223 ms on one project and 1239 ms on another, versus roughly 37 ms and 49 ms for `.distinct()`. `Exists()` short-circuits when proving an algorithm *has* a classification, but the query has to run the subquery to completion for every algorithm that does *not*, and most unpipelined algorithms have none in a given project. `.distinct()` keeps the single scan and just deduplicates it.

### Known gap: an algorithm on a pipeline not enabled here, but with determinations here

The second query is restricted to algorithms with no pipeline at all, so an algorithm attached to a pipeline that exists but is not enabled for this project stays out of the list even when it owns classifications here. This is real in the data: on one project, `Turing Kenya and Uganda Species Classifier` has 207 classifications but sits on a pipeline not enabled there, so its determinations are unfilterable. That is the same class of gap this PR set out to fix, now closed for standalone post-processing algorithms but not for this shape.

I left it out on purpose. Closing it means widening the candidate set from "no pipeline" to "not configured-and-enabled here", which drops the `pipelines__isnull=True` guard and puts the classification join back over a larger candidate set — measured at roughly five times the runtime (about 175-275 ms versus 40-50 ms). It is a reasonable follow-up if operators hit it, but it is a separate scope decision with its own cost, not a fix that belongs quietly in this PR. Flagging for your call.

mihow:
**Claude says:** One more correction, and a genuine speedup that came out of it. A review pass pointed out that a mechanism I described three times (PR body, an earlier comment, the code comment) was not what the database actually did — and checking it with EXPLAIN showed they were right.

I had said that restricting the post-processing lookup to unpipelined algorithms "keeps it off the full classification table". It did not. Written as a single join —

```
Algorithm.objects.filter(pipelines__isnull=True, classifications__detection__source_image__project=project)
```

— the `pipelines__isnull=True` anti-join hides the candidate algorithm ids at plan time, so Postgres cannot use the `algorithm_id` index and sequentially scans the whole `main_classification` table (a parallel seq scan of ~278k rows per worker, ~48 ms). Its cost tracks the platform-wide classification count, and it is scanned cold on every cachalot invalidation, i.e. on every classification write.

Fixed properly in `e691c035` by doing it in two steps: collect the unpipelined algorithm ids first as an explicit list, then filter `Classification.objects.filter(algorithm_id__in=ids, detection__source_image__project=project)`. With a literal IN list the planner uses the index. EXPLAIN confirms the plan flips from a seq scan to an index scan, ~48 ms to ~6 ms, and the whole endpoint drops from ~46 ms to ~12 ms. It is now flat in table size rather than linear, which was the property I had wrongly claimed the first version already had.

One footgun handled in the same commit: `Classification` has a default `Meta.ordering`, and a `.values_list(...).distinct()` silently folds the ordering columns into the DISTINCT, which would return one row per classification again. `.order_by()` before `.distinct()` clears it; the dedup test now pins that.

So the honest mechanism is: the enabled-pipeline join covers detectors and configured classifiers; the two-step lookup adds standalone post-processing algorithms via their classifications, indexed and flat in table size. Apologies for the earlier hand-wave — I asserted the plan I wanted rather than the plan Postgres produced, which is exactly the thing an EXPLAIN would have caught, and did once someone looked.

ami/ml/views.py
```diff
@@ -56,15 +56,58 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
...
         if getattr(self, "action", None) == "list":
             project = self.get_active_project()
             if project:
-                qs = qs.filter(
+                # Algorithms reach the list two ways. An enabled pipeline configures most of
+                # them, and that join alone covers detectors, which never author a
+                # Classification and so cannot be found by their results at all.
+                configured_for_project = Algorithm.objects.filter(
                     pipelines__project_pipeline_configs__project=project,
                     pipelines__project_pipeline_configs__enabled=True,
```
mihow, line 70: I am okay with this compromise if it's the only way to make the query fast until we add a denormalized project column to the Classification table. BUT it is not accurate. The idea is to see algorithms actually used even if they are no longer enabled (e.g. old versions). Also, Detection algorithms can be retrieved from the classification__detection relationship.

ami/ml/views.py
```diff
@@ -56,15 +56,58 @@ class AlgorithmViewSet(DefaultViewSet, ProjectMixin):
...
+                # verbatim into the IN clause, and Algorithm's (name, version) ordering would
+                # otherwise vary the id order — and so the SQL string and its cachalot key —
+                # for identical membership.
+                unpipelined_algorithm_ids = sorted(
+                    Algorithm.objects.filter(pipelines__isnull=True).values_list("pk", flat=True)
+                )
+                # `.order_by()` clears Classification's default ordering before `.distinct()`.
+                # Without it the ordering columns join the SELECT and widen the DISTINCT, so it
+                # would return one row per classification rather than one per algorithm.
+                post_processing_used_in_project = (
```
mihow: Can we move this to a query method on Occurrence and or Algorithm? a general way to retrieve a unique list of algorithms used in a project and filter Occurrence by algorithm used?

mihow: **Claude says:** Done in `996277d0`. Two reusable methods now carry the definition:

- `Algorithm.objects.used_in_project(project)` returns the unique algorithms that produced results in the project — classifiers and post-processing algorithms through their classifications, detectors through `Detection.detection_algorithm`. `AlgorithmViewSet` calls it, so pipeline configuration is no longer consulted: superseded versions on disabled pipelines are included, and configured-but-never-run algorithms are dropped.
- `Occurrence.objects.detected_or_classified_by(ids)` / `not_detected_or_classified_by(ids)` filter occurrences by any of those algorithms, matching a detector's detection or a classifier's classification through one `EXISTS` subquery. `OccurrenceAlgorithmFilter` routes through them.

Rewiring the filter surfaced a counting bug: the old `detections__classifications__algorithm__in` join returned one row per matching classification with no `.distinct()`, so the paginator's COUNT over-reported (111,272 vs 55,214 for one classifier on the largest project), and it matched only classifiers — filtering by a detector returned nothing. The EXISTS form counts each occurrence once and matches both roles.

ami/ml/views.py
mihow: **Claude says:** Went with the accurate definition rather than the compromise. `used_in_project` now includes old/superseded versions whose determinations still exist (pipeline enablement is not consulted) and reaches detectors through `Detection.detection_algorithm`, exactly as you described.

The honest cost: neither `Classification` nor `Detection` has a project column, so the lookup scans — a project is reachable only through `source_image`. Measured ~0.1–0.6 s cold on the larger projects against the prod copy, cached warm by cachalot. The fast hybrid (~12 ms) is a one-method revert if that cold cost is a problem before the denormalized `project` column lands, but it can't express "used" without dropping the old versions and the pipeline-less detectors, so the accurate one is what's in. Flagged the denormalized column as the durable fix in the PR notes.

mihow: **Claude says:** Final shape, per your note about graying: one list serves both consumers instead of splitting semantics. `used_in_project` returns everything that ran (old versions + detectors + post-processing), and each row now carries `enabled_in_project` — true if it's on a pipeline the project has enabled, false if it only ran historically. The occurrence filter uses the full set; the algorithms page shows the full set but grays the `false` rows (new optional `rowClassName` on the shared Table). So old/superseded versions stay filterable and are visibly marked as no-longer-enabled rather than hidden.

Honest cost is unchanged (~0.1–0.6 s cold, cachalot-warm after) since it's the same `used_in_project` query; the `enabled_in_project` flag is a cheap per-row Exists on top. Full measurement table (algorithm-list shape ladder, EXPLAIN index-vs-seqscan, and the occurrence-filter 111,272-vs-55,214 count fix) is in the updated PR description. Pushed in `68dc34b3`.

PR #1407, opened 2026-09-03
[Draft] Retrain a classifier head from verified identifications, and score what it produces

mohamedelabbas1996:
## Summary

When someone confirms a species in Antenna, that answer should make the model better. This adds the machinery to do that without retraining BioCLIP itself.

The backbone is frozen, so only the small classifier head on top needs retraining — and because the backbone never changes, a crop's embedding never changes either. Antenna now keeps that embedding the first time a crop is classified, so retraining becomes fitting a small matrix over stored vectors: seconds, not weeks, and the images are never opened again.

A job gathers the verified crops for a project, writes them to a file in project storage, and hands a processing service the link. The service fits a new head, scores it against the head it is currently serving on the same held-out rows, and reports back. Every retrain is recorded as a new algorithm version, so any prediction stays traceable to the exact weights and dataset that produced it. Nothing is promoted automatically: a new head is registered and scored, and deciding to use it stays a human step.

Retraining raises a question it cannot answer itself — is the new head actually better? So this also adds evaluation. An occurrence set is a fixed, stored list of verified occurrences; scoring compares an algorithm's stored classifications against what people identified, in one pass over the database with no images opened and no GPU. Two heads can only be compared if they answered the same questions, which is why the membership is stored rather than re-sampled.

**Split across three PRs.** The interface for all of this is RolnickLab/antenna#1423. The processing-service half — a BioCLIP classifier and the `/train` endpoint — is RolnickLab/ami-data-companion#167. This PR is the Antenna backend.

### List of Changes

| # | What it does for a user or operator | How |
|---|---|---|
| 1 | Embeddings computed during classification are kept instead of thrown away | `features` on `ClassificationResponse`; `DetectionEmbedding` table (`ml/0029`) |
| 2 | Storing them is opt-in per project, so no project pays for it unless it will retrain | `store_classification_embeddings` feature flag |
| 3 | Verified crops missing an embedding can be filled in without re-running the detector | `generate_embeddings` job (`jobs/0025`) |
| 4 | A service can say which of its algorithms are worth retraining | `trainable` on `AlgorithmConfigResponse`, mirrored onto `Algorithm` (`ml/0030`) |
| 5 | A project's verified labels become a training set and are handed to a service | `train_classifier` job (`jobs/0024`), `training_dataset.py`, `training_dispatch.py` |
| 6 | A retrained head covers the project's species list, not only what someone verified | `Project.default_taxa_list` (`main/0096`), read when the set is built |
| 7 | Training can outlast the request that started it | `POST /jobs/{id}/training-result/`, authorised by a signed token |
| 8 | Each retrain is its own algorithm version, so past predictions stay attributable | same name, next version, new key |
| 9 | Any head can be traced to the data and scores that produced it | `Algorithm.training_info` — dataset URL, metrics, parent key, job id, warnings |
| 10 | An evaluation set can avoid anything a model trained on | `TrainingSetMembership` (`ml/0032`) |
| 11 | Two models can be compared on identical data | `OccurrenceSet` (`ml/0033`); read-only at `/ml/occurrence-sets/` |
| 12 | A model is scored without a GPU or reprocessing anything | `evaluation.py` compares stored classifications to human identifications in one pass |
| 13 | A score says both how often a model is right and how it treats rare species | `micro_accuracy` and `macro_accuracy`, plus per-species `TaxonEvaluation` rows |
| 14 | Scoring runs as a job like anything else | `evaluate_algorithm` job (`jobs/0026`) |
| 15 | The numbers reach the screens that show them | `training_crops_ready`, `algorithm_performance`, `best_model`, `evaluations` |
| 16 | A job that could never run is refused when it is made | each job type declares `required_fields` / `required_params` |

### Related Issues

Relates to the classifier-head retraining work discussed with @mihow.

### Why the dataset goes through storage rather than the request body

Antenna writes the training set as one `.npz` in project storage and sends a URL. Production holds around 273,000 human identifications; at roughly 2 KB a vector that is hundreds of megabytes — fragile to send in one request, and it starts over if the connection drops. A file is cheap to retry and needs nothing new from the service, which already downloads capture images from the same place.

**Measured** on 10,000 vectors of the real shape:

| Format | Size | Time to build |
|---|---|---|
| JSON, as the API serves it | 227.6 MB | 29.3 s |
| npz, float32 | 41.7 MB | 2.1 s |
| **npz, float16** | **21.2 MB** | **1.1 s** |

float16 loses nothing: Postgres already stores these as two bytes each (`halfvec`), so JSON was printing nine decimal places for numbers that never had them.

### Why the held-out split is grouped by occurrence

An occurrence is one insect across several frames, so its crops are near-duplicates. Splitting by detection puts near-identical crops on both sides and reports an accuracy that is far too high. The assignment is a hash of the occurrence id rather than a random draw, so re-running after new data arrives leaves the existing held-out set exactly where it was — an eval set that drifts cannot compare two heads.

### Why there is no vector index

These vectors are read in bulk to fit a head, never searched by nearest neighbour. **Measured** at 20,000 rows: the rows occupy 54 MB and an HNSW index a further 45 MB, so an index nearly doubles the cost for no benefit. An index can be built later if similarity search is ever wanted.

### Why a retrain creates a new Algorithm row

A `Classification` points at an `Algorithm`, so a version changed in place would make past predictions untraceable. Each retrain creates a row: same `name`, next `version`, new `key`, because `key` is unique on its own while `(name, version)` is unique together.

### Why scoring reads the database instead of running the model

Both halves of the comparison are already stored — what a person identified, and what the algorithm classified — so an algorithm that has processed a set is scored in a single query. An algorithm is only asked about species it can actually predict: its category map is the list of answers available to it, so an occurrence of anything else is counted as skipped rather than wrong, otherwise a regional head looks bad for not knowing a species nobody trained it on. An algorithm that never ran on the set raises rather than returning zero.

### Why the best model is ranked on the per-species average

Trap data is long-tailed. A head that handles only the common species scores well on the plain share while being useless for the species anyone wants found. `best_model` orders on `macro_accuracy` first, using `micro_accuracy` only to break ties.

### Why training-crop counts do not roll up the taxonomy

`verified_count` rolls descendants up. `training_crops_ready` deliberately does not: a head is fit on the label itself, so a crop verified as a species is not training data for its genus. It counts crops rather than occurrences, because one insect across three frames is three images to train on. Like the Example column in #1365 it is gated behind `?with_training_crop_counts=true`, and annotates `NULL` when not requested — a zero would read as a species with nothing to train on.

## How to Test the Changes

Run end to end against a BioCLIP service (ami-data-companion#167), triggered from the UI:

1. Confirm occurrences.
2. Run `generate_embeddings`. The service log shows `Skipping the localizer, use existing detections` — the detector does not run again.
3. Run `train_classifier` against a 17-species taxa list of which 10 had verified crops. The new head came out with **17 classes, 7 carried over from the current head**, and the occurrences used were recorded.
4. The new head scored **0.636** against the current head's **0.875**, so it was **not** promoted, and was still registered as a new version alongside its dataset.
5. Build an occurrence set and run `evaluate_algorithm` against two heads. Scores are stored per algorithm and per species.

That the new head lost is the expected result, not a defect: a head trained on ~96 crops should lose to one trained on the full Newfoundland set. The comparison, and the refusal, are the point.

**A caveat about the blind set used above.** It was assembled from "Confirm" clicks, and confirming means agreeing with the model, so the head in service scores 1.000 on it by construction. A fair set needs corrections in it too; building one is separate work.

**Tests:** 839 pass, 2 skipped.

## Deployment note

**This needs the `pgvector` extension in the production database.** The embedding column is `halfvec`, and `ml/0029` runs `CREATE EXTENSION vector`. Locally the Postgres image installs `postgresql-16-pgvector`; the stock image is used rather than `pgvector/pgvector:pg16`, because that image is built on an older Debian and swapping it under an existing data directory triggers a collation version mismatch.

## Known gaps

1. **Retrained heads accumulate.** Every run adds a head that is never removed, so the picker grows without limit. This wants a retirement rule or a status field on `Algorithm` — a design question rather than an oversight.
2. **No permission rules** on who may retrain.
3. **`params` is writable on the Job API.** A `train_classifier` job carries its `algorithm_key` there. Required keys are checked per job type, but the field still accepts arbitrary JSON from anyone who can create a job.
4. **`MLJob.process_images()` gained a keyword argument.** The default preserves existing behaviour, but it is a shared code path.
5. **`occurrences_safe_to_evaluate_on()` is not called in production code.** It exists and is tested; the sets used so far were assembled by hand, and wiring it in is what would make a set blind by construction.
6. **Nothing reports embedding coverage across a project**, which is the number that says whether a retrain is worth starting at all.


mohamedelabbas1996:
### Note: where retrained head weights live

Raising this here because it is a gap this PR creates rather than one it inherits.

A retrained head is written to the processing service's own disk, under
`~/.cache/torch/hub/trained_heads`. That is a **cache** directory, so the weights are lost
whenever the cache is cleared, the container is rebuilt, or the GPU VM is replaced. With
more than one replica of the service, a head trained on one is not servable by another.

Antenna records that a version exists but not where its weights are, which makes promoting
one later meaningless: the artifact being promoted may already be gone. Nothing is promoted
automatically — deliberately — but "a person decides later" only works if the file is still
there when they decide.

**What I have implemented in this PR:** the service uploads the head to Antenna after
training, Antenna stores it through `default_storage` (the same place the training set is
written), and records the location on `Algorithm.uri` — the field already documented as
"A URI to the weights or model details".

Two design points worth review:

1. **The upload goes through Antenna rather than a presigned URL.** The service holds no
   storage credentials today — it fetches everything over plain URLs — and a presigned PUT
   would only work where the storage backend is S3, not against the local filesystem. Going
   through Antenna behaves identically in both and reuses the signed callback token the
   service already has. This is the opposite direction to the training set, which is handed
   over as a URL precisely because it is large; a head is 72 KB over 17 species and about
   3 MB over 749, so passing it through is not worth avoiding.

2. **Re-download is not wired up.** The weights are now durable and their location is
   recorded, but nothing fetches them back automatically if the service's disk copy is
   gone. That wants head discovery from storage at startup, which felt like a separate
   change rather than something to fold in here.

Storage cost is small enough that keeping every version indefinitely is viable, which turns
the retention question from "which do we delete" into "which do we serve" — still open.

mohamedelabbas1996:
### Note from @mihow: UUIDs for Algorithm and Pipeline

Recording this here so it is not lost, since this PR adds new algorithm records.

> one note for the new algorithms: let's plan on switching from slugs/keys to UUIDs for
> Algorithm and Pipeline objects. Because they are shared across databases & applications
> & environments.
>
> (also Taxon will have a UUID at some point as well, since the GBIF key has turned out
> not to be a suitable global identifier)

Where this PR touches that:

- **`key` is the identifier across the Antenna / processing-service boundary.** The train
  request sends `algorithm_key`, and the service looks the head up by it. So the switch is
  not only an internal rename — it changes a contract between two applications, which is
  exactly the cross-application case the note is about.
- **Every retrain mints a new key**, `<parent-key>-v<n>-<timestamp>`. A version's identity
  is a string built from its parent's string, so the keys inherit whatever the parent's
  does.
- **`training_info.parent_algorithm_key`** records lineage by key, and would become a UUID
  reference.
- The mismatch is already visible in practice: the same model is registered under
  `bioclip-2-5-vith14-logreg-nf749` by one deployment and `bioclip_25_logreg_head_newfoundland`
  by another, and a head trained through one cannot find its baseline through the other. A
  stable identifier would remove that class of problem rather than paper over it.

I have not changed anything here — flagging it as a follow-up so the new records are part
of that migration rather than an exception to it. Happy to take it as a separate PR once
the shape is agreed.

PR #1419, opened 2026-09-15
Update docker image source for minio and minio-init from docker hub to quay.io

kra:
## Summary

Update docker image source for minio and minio-init from docker hub to quay.io.

### List of Changes

* Modified docker-compose.yml

## Detailed Description

I was unable to build after pruning without this change.

## Summary by CodeRabbit

* **Chores**
  * Updated MinIO and MinIO Client image sources to use the Quay.io registry in local and CI Compose configurations.
  * Image versions remain unchanged.
  * Existing local development and CI behavior is preserved.


kra:
Minio is now source-only since early 2026.
https://github.com/minio/minio?tab=readme-ov-file#source-only-distribution
There is allegedly a CVE in the final docker hub image, in any case, all minio/minio images have been removed from docker hub.

I used the quay.io image because hey, it's redhat, but I don't know anything about it and there doesn't seem to be documentation about where it comes from? Do sources just link to their distro sites there?
https://quay.io/repository/minio/minio

There are several minio images in docker hub, but I don't know if any are reliable.

Docker offers images with paid extended lifescycle support
https://www.docker.com/blog/minio-end-of-life-how-to-stay-patched-and-audit-ready-with-docker-els/

Minimus claims to be a drop-in replacement
https://www.minimus.io/post/minio-docker-image-changes-how-to-find-a-secure-minio-alternative

Chainguard claims to be a container image
https://www.chainguard.dev/unchained/secure-and-free-minio-chainguard-containers

There are several places distributing images which claim to be regular builds kicked off with every source release (presumably of minio and the relevant libraries etc), but I don't know if they are valid.

loppear:
This seems fine, quay.io version doesn't appear to be updated (latest tag is RELEASE.2025-09-07, while RELEASE.2025-10-15 was out before the repo went dead) but we're already tagging to old fixed releases and only using this in dev so no changes seems fine until we decide to trust another updated source/alternative.

kra:
But if I'm correct, we don't have a valid build process for a dev box until this is fixed - you and everybody else are able to build because you have the minio images cached. If, like me, you are setting up a fresh dev host, or if you "docker prune -a", you will not be able to run "docker compose up" or similar unless you incorporate this change (or another change that uses different minio images).

loppear:
Yes, my other PR builds are also broken on this on github etc, has my :+1:  but I'll let @mihow give review approval in case there are further systemic implications.

mihow:
Sorry for the delay. Thanks for finding this @kra & @loppear

Issue #1424, opened 2026-09-17
Classifier species lists as taxa lists and support for public lists


mihow:
## Where this fits

The larger goal is that a user is only shown results that are accurate and useful for the place
they monitor (#890). Most classifiers in Antenna cover far more species than occur at any one site,
so the platform needs a way to say "these are the species that matter here" and to apply that to
predictions. Taxa lists are that mechanism, and the work has been arriving in stages: taxa lists
and their editor (#1094, #1104), filtering taxa and occurrences by a list (#745, #1008, #1347),
class masking against a list (#999) on the post-processing framework (#1289), and verifying
presence species by species (#1320, #1365). Still ahead are building a list from a region (#1364,
draft #1367), a user-facing way to turn masking on (#1406), and the earlier geo-fencing idea
(#797). Further out, members add their own taxa for identification, every taxon shows whether a
model can predict it or only a person can identify it, and taxa carry regional information.

This ticket supplies the piece those all lean on: an authoritative, browsable answer to "what can
each model predict", shared across projects, that a project can copy and make its own. It also
settles how shared and project-owned objects coexist, which taxa lists, processing services and
future curator roles all need. A full list of related issues and pull requests is at the end.

## Summary

People who use Antenna regularly ask "can this model predict species X?" and today the platform
cannot answer, because the full set of classes a classifier knows (its category map) is not visible
anywhere a user can browse it. The closest thing, the "Taxa returned by <algorithm>" lists, only
holds species the model has already predicted somewhere, so it cannot show a "no".

This ticket proposes using the existing Taxa Lists feature to close that gap. Every classifier gets
one taxa list holding everything in its category map, public when the classifier itself is
available to every project. A project member can copy that list
into their own project, edit the copy (remove species that do not occur in their region, add
missing ones), and later choose it as the list for class masking. The first users are partners who
identify with a regional classifier and want a list for a neighbouring region derived from it.

Doing this well depends on a second piece of work: Antenna has several kinds of objects that can be
either shared across all projects or private to some projects (taxa lists, taxa, tags, pipelines,
processing services), and each one handles that split differently today. Public taxa lists already
exist in the database but appear to be unreachable through the API. The ticket therefore includes a
consistent rule for "public versus project-scoped" as related work.

## Terminology

Use **public** for an object available to every project and **project-scoped** (or private) for an
object tied to specific projects. Avoid "global": in this domain it already means geographic scope
(the global moth classifier, global species lists), and the two meanings collide in exactly the
[...]

PR #1429, opened 2026-09-20
Show 'Not found' instead of breaking the page when a link has an invalid id

mihow:
## Summary

A link with a bad id — mistyped, stale, or truncated — could break a whole page rather than saying the thing was not there. Opening `/projects/<id>/taxa/lists`, for example, replaced the project view with "Something went wrong! Cannot read properties of undefined (reading 'length')", and the only way out was to navigate away by hand.

This PR makes every detail link that carries an id behave the same way: if the id cannot be a real record, the app says "Not found" in the dialog or page where the record would have appeared, and the list behind it keeps working. Nothing changes for a valid link.

### List of Changes

| # | What changes for the user | How |
|---|---|---|
| 1 | A detail link with an id that is not a record number shows "Not found" instead of breaking the page | New `isDetailRouteId` helper (a positive integer, nothing else); each detail dialog or page skips the request and shows the existing error state |
| 2 | Applied to every detail route that takes an id | Taxa, occurrences, jobs, deployments, sessions, exports, pipelines, algorithms, processing services, taxa lists, and the taxon shown inside a taxa list |
| 3 | An invalid taxa-list id no longer sends a broken filter to the server | The species query for that page now receives no list filter instead of the invalid one, which was returning 500s |
| 4 | A view waiting on a request it has deliberately not made no longer shows a spinner that never resolves | `useAuthorizedQuery` reports `isLoading: false` when a caller passes `enabled: false` |

### Related Issues

Found while reviewing the taxa-list work for #1424; independent of it.

## Detailed Description

**Why a bad id could break the page rather than 404.** The detail routes accept any string, so `/taxa/lists` made the app request `/api/v2/taxa/lists/`. That is a real endpoint — the taxa lists collection — and it answers 200 with a paginated envelope. The dialog built a taxon out of `{count, next, previous, results}`, every field came back undefined, and the first component to read one threw. Other invalid ids were already fine: `/api/v2/taxa/foo/` and `/api/v2/taxa/99999999/` both return 404, which the dialogs already display. The failure needed an id that collides with a sibling collection path, so the fix belongs on the id rather than on the word "lists": any later addition such as `/occurrences/summary/` would bring the problem back.

**The spinner that never resolves.** React Query v4 removed the `idle` status, so a query created with `enabled: false` reports `isLoading: true` until it runs. Disabling the fetch for an invalid id therefore produced a blank, permanently hidden dialog instead of the not-found message. `useAuthorizedQuery` now reports `isLoading: false` when a caller explicitly passes `enabled: false`; queries that do not pass it are untouched. Of the fourteen call sites that do, thirteen pass `enabled: !!id`, and the fourteenth (`useUserInfo`, disabled when logged out) has no consumer of its loading flag.

Several `use*Details` hooks took `id: string` with no way to skip a request; they now take `string | undefined` and skip when it is missing, which is what the two hooks that already supported it did.

**Known and left alone.** On a taxa-list page with an invalid id the species query still fires once, unfiltered, before the not-found branch renders; the result is discarded. Gating it needs an `enabled` flag on a hook that has none, which is more change than this bug warrants. The breadcrumb on that page also keeps its "Loading data..." label in the not-found state.

### How to Test the Changes

Unit tests cover the helper (`12` valid; `lists`, `foo`, `1.5`, `-1`, `" 1"`, `"0"`, empty and undefined invalid) and the loading-state fix. The frontend suite passes (62 tests), along with `tsc --noEmit`, eslint and prettier.

By hand, in any project:
1. `/projects/<id>/taxa/lists` shows a "Not found" dialog over the taxa list, with no console errors. Before this change it replaced the page with an error.
2. `/projects/<id>/taxa/<a real taxon id>` still opens the taxon as usual.
3. `/projects/<id>/taxa-lists/foo` shows a not-found page rather than a run of 500s.
4. `/projects/<id>/occurrences/notanumber` shows a "Not found" dialog over the occurrence list.

## Deployment Notes

None. Frontend only, no migration, no settings.

https://claude.ai/code/session_01TngW6AshkUEhNp9D6z3U9Z

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid detail-route IDs no longer trigger unnecessary data requests.
  * Detail views now display a clear “Not found” message for invalid or missing IDs.
  * Disabled data requests correctly report that loading has finished.
  * Detail pages continue to show backend errors when valid requests fail.

* **Tests**
  * Added coverage for route-ID validation and disabled-request behavior.


ui/src/data-services/hooks/algorithm/useAlgorithmDetails.ts
```diff
@@ -6,14 +6,15 @@ import { useAuthorizedQuery } from '../auth/useAuthorizedQuery'
 const convertServerRecord = (record: ServerAlgorithm) => new Algorithm(record)
 
 export const useAlgorithmDetails = (
-  algorithmId: string
+  algorithmId: string | undefined
```
mihow, line 9: Why would we have undefined IDs? can we return a different response from the backend instead if an algorithm is not found and handle that?

